Issue #4, opened 2023-04-04
Unit testing


mondus:
Unit testing is absolutely required for this code. Ideally this could be done using Fortran by invoking the C API.

mondus:
Suggestions from @christopheredsall are

 - [funit](https://github.com/ieyasu/funit) - now achieved
 - [pfunit](https://pfunit.sourceforge.net/page_Installation.html) recommended and used at MET office 

There must be a good way to integrate with Cmake to do test discovery. Some googling required.

dorchard:
We had a discussion that, due to the nature of this code, there is a lot of correctness that would be covered by integration tests (https://github.com/Cambridge-ICCS/FTorch/issues/62), and unit testing would be a little harder. That said, I can see how `torch_module_forward` and `torch_from_blob_c` could have some unit tests.

dorchard:
fwiw, pfunit was pretty easy to build locally (in non-MPI mode at least).

PR #19, opened 2023-06-02
Restructuring

### LINKED ISSUE #12: Separate examples from library source
We should separate the examples from the key source files in the library.

Perhaps into an `/examples` directory?

jatkinson1000:
Restructure repository as discussed in #12 

Closes #12 

- [ ] Move tools such as `pt2ts.py` to a utils directory with a readme
- [ ] Move examples to examples directory with a simple readme
  - [ ] Move examples
  - [ ] Detailed documentation elsewhere
  - [ ] Remove examples from build and install process


jatkinson1000:
It occurs to me that we may want to use these simple ts_inference examples for testing, but we can address that in #4 
It may also be that we want even simpler self-contained examples (not relying on TorchVision).

dorchard (approved):

Issue #24, opened 2023-06-19
Wrap tensor generation routines from Fortran


jatkinson1000:
It would be nice if we could wrap the fortran tensor generation to transform `torch_tensor_from_blob` into something like `torch_tensor_from_array`

- [ ] Takes an array as the main, and only required, argument
- [ ] Infers dtype
- [ ] Infers size
- [ ] Infers shape
- [ ] Shape and stride are optional arguments
- [ ] Indexing `'F'` or `'C'` passed through as string -> `c_char`
  - [ ] Should raise error if this is passed in addition to stride and shape.

SimonClifford:
Documenting the discussion @jatkinson1000 and @SimonClifford  just had:
We would like two groups of functions: the ones for ordinary users and the ones for power users.  There may be some overlap.  The ordinary users want to be able to pass  an array to a function and get an appropriate tensor, so we give them `torch_tensor_from_array` that looks like the above (`torch_tensor_from_array(array, [stride])`, infers everything).  One issue with this is that due to lack of Fortran generics there may need to be a plethora of functions.

The power users continue to use `torch_tensor_from_blob` where they specify dtype, shape, and strides. This allows C-style or Fortran-style array indexing, or even some arbitrary indexing (`(i, j, k) -> (i, k, j)`). Also can implicitly reduce or increase dimensionality of array, e.g. passing a 3D `(i, j, k)` array in but creating a 2D `(ij, k)` tensor. This will, of course, need careful documenting.

Possible Fortran function signatures (can have more than one):

- `tensor_from_blob(data, ndims, shape, dtype, device, layout)` where `layout` is an array that gives the order of the indices to be used.  So to access a 3D array Fortran-style `layout` would be `(1, 2, 3)`, whereas C-style it would be `(3, 2, 1)` .
- `tensor_from_blob(data, ndims, shape, dtype, device, layout_style)` where `layout_style` is `'F'` for Fortran-style, or `'C'` for C-style. This would call the above with the appropriate `layout`.

PR #25, opened 2023-06-19
Allow users to use strides in Tensor creation

SimonClifford:
As documented in #24.  Essentially by wrapping the PyTorch `torch::from_blob` functions (https://pytorch.org/cppdocs/api/namespace_torch.html#namespace-torch) that take `strides` as an argument we can permit users to implicitly transform their data.  E.g. from Fortran indexing to C, or vice-versa.

We will provide Fortran interfaces to suit ordinary and power users.


jatkinson1000 (approved):
LGTM!

PR #29, opened 2023-07-14
Fixed missing array layouts in ResNet18 example

helgehr:
Thank you for this nice resource! I tried the ResNet18 example and noticed that the in/out layouts are missing at the call to `torch_tensor_from_blob`. This is a quick fix for that.


jatkinson1000 (approved):
This is great, thanks so much.
The striding approach with the `layout` option is a recent addition and we clearly forgot to update the examples!

This is all working for me so I'll go ahead and merge.

Apologies it took a little while (been away and teaching) but thank you so much for getting involved!

PR #31, opened 2023-07-28
Update deprecated pytorch arguments in ResNet example.

### LINKED ISSUE #30: RESNET_18 default arguments updated
In the ResNet example we use `pretrained=True`, but this is deprecated and should now be `weights='IMAGENET1K_V1'`.

jatkinson1000:
As title says.

Closes #30


TomMelt (approved):
thanks for updating this

PR #38, opened 2023-09-20
Add realistic example

### LINKED ISSUE #32: ResNet Example outputs could be improved
Currently the outputs from the python ResNet inference program are not the same as the outputs from the Fortran ResNet inference program in the example.

Really we should make sure they, at minimum, produce the same output, and ideally make it something meaningful (e.g. max arg and location, perhaps provide a real image).

ElliottKasoar:
Fixes #32(?)

Replaces the example ones tensor with a more realistic example, based on a [Pytorch Resnet example](https://pytorch.org/hub/pytorch_vision_resnet/).

To do:
[] Update resnet_infer_fortran.f90 to read in the saved text file for the image
[] Update resnet_infer_fortran.f90 to output labels (possibly save the labels in a similar manner to the image?)


examples/1_ResNet18/resnet18.py
```diff
@@ -29,20 +31,71 @@ def initialize():
...
-    print(f"Top 5 results:\n  {top5}")
+    # Transform image into the  form expected by the model
+    input_image = Image.open(filename)
+    preprocess = torchvision.transforms.Compose(
+        [
+            torchvision.transforms.Resize(256),
+            torchvision.transforms.CenterCrop(224),
+            torchvision.transforms.ToTensor(),
+            torchvision.transforms.Normalize(
+                mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225]
```
TomMelt, line 57: Are these hard-coded? are they unique to this specific example?

examples/1_ResNet18/resnet18.py
```diff
@@ -29,20 +31,71 @@ def initialize():
...
+    probabilities = torch.nn.functional.softmax(output[0], dim=0)
+
+    # Download ImageNet labels
+    url, filename = (
+        "https://raw.githubusercontent.com/pytorch/hub/e55b003/imagenet_classes.txt",
+        "imagenet_classes.txt",
+    )
+    data = urllib.request.urlopen(url)
+    for _ in data:
+        categories = [s.strip().decode("utf-8") for s in data]
```
TomMelt, line 91: is this loop necessary?

examples/1_ResNet18/resnet_infer_fortran.f90
```diff
@@ -1,62 +1,162 @@
+module precisn
+   use, intrinsic :: iso_c_binding, only: c_float
+   implicit none
+   private
+   integer, parameter:: wp = c_float
+   public wp
+end module precisn
```
TomMelt, line 7: We can put this in a separate module file

TomMelt:
Thanks for starting this. It looks great so far :+1:

examples/1_ResNet18/resnet18.py
jatkinson1000: These will be the pre-processing required for ResNet which expects a 224x224 resolution image.
You also need to normalise the channels.

The documentation tells you how to prepare these [here](https://pytorch.org/vision/stable/models/generated/torchvision.models.resnet18.html#torchvision.models.resnet18).

I think what is done here for the sake of an example is sufficient, but perhaps extend the comment to clarify @ElliottKasoar

examples/1_ResNet18/resnet_infer_fortran.f90
jatkinson1000: What's the argument for defining `c_float` in a separate module rather than with the rest of the `iso_c_binding` imports as was previously done?
Is it because we could have different float precisions - as I can't see where we are specifying anything differently?

jatkinson1000 (commented):
Thanks for working on this @ElliottKasoar 
A couple of comments:
- Whilst it is arguably 'cleaner' in terms of storage to fetch the image from url, there is no guarantee it will always be available. Since it is only a single image I would consider just adding it to the repository to make the example self-contained. This will also arguably make the python a bit cleaner.
- Saving the data to file is a good idea, but could we do it more nicely than `.flatten().to_file()`? And consider adding `*.dat` to `.gitignore`. Also a comment in the docs about writing/reading between row/column major systems for clarity could be really  helpful for users' understanding.

examples/1_ResNet18/resnet18.py
ElliottKasoar: Yeah that's right, they are the standard for all their pre-trained models, based on normalising the ImageNet dataset, I believe.

It's surprisingly hard to find anywhere were they explicitly state the origin of the numbers, but hopefully the comments I've added now are sufficient?

examples/1_ResNet18/resnet18.py
ElliottKasoar: Good spot, I had different logic at one point, but there doesn't seem to be a need anymore.

.gitignore
```diff
@@ -183,3 +183,9 @@ cython_debug/
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
+
+# Data binary
+*.dat
+
+# TorchScript model
+*.pt
```
TomMelt, line 191: these are great. Should we add *.jpg too? I think dog.jpg is downloaded via `examples/1_ResNet18/resnet18.py`

examples/1_ResNet18/README.md
```diff
@@ -34,10 +34,22 @@ pip install -r requirements.txt
...
+When using single precision, it should produce the result:
+
+```
+Top 5 results:
+
+Samoyed (id=258): probability = 0.8846225142478943
+Arctic fox (id=279): probability = 0.045805174857378006
+white wolf (id=270): probability = 0.0442761555314064
+Pomeranian (id=259): probability = 0.005621383432298899
+Great Pyrenees (id=257): probability = 0.004652013536542654
```
TomMelt, line 51: This is great. Thanks for including the IDs as well

examples/1_ResNet18/resnet_infer_fortran.f90
ElliottKasoar: Yes, having a straightforward setting for different precision was the idea. From discussion with @TomMelt earlier, I've changed the implementation from what would have been a separate module, as that would obfuscate the C-typings, which wouldn't be ideal.

I think they way it's implemented now allows for single and double precision, but it does require changing values in a few different locations.

It works, as is, for single precision, so if it's not too messy, I'll just add a comment in the readme about making sure to change the precision everywhere that's needed for double.

.gitignore
ElliottKasoar: I added dog.jpg based on @jatkinson1000's suggestion to make it a more self-contained example, but I'm happy either way.

jatkinson1000:
> > ```
> > * Whilst it is arguably 'cleaner' in terms of storage to fetch the image from url, there is no guarantee it will always be available. Since it is only a single image I would consider just adding it to the repository to make the example self-contained. This will also arguably make the python a bit cleaner.
> > ```
> 
> What about something in between (which I'm halfway along, having added the .jpg to the PR): try to download the data, but, if necessary, catch the `HTTPError` for both the image and the categories, as we can direct users to the pre-downloaded files.
> 
> I quite like having some form of the fetch present, as, for example, I wanted to test a different [dog](https://github.com/pytorch/vision/blob/v0.15.1/gallery/assets/dog1.jpg), which only requires changing one line of code at the moment, and I think it makes an example better if the barrier to trying out different data is as low as possible.

Yes, testing alternative images did cross my mind, but the purpose of this example is to serve as an (ideally minimal) example of how to save a pytorch net and couple it to Fortran. To this end I'd maybe favour having the image in the repo to read from and removing the requests code, perhaps with a note in the Readme to describe how a user might go about trying other images.

jatkinson1000:
> @jatkinson1000 What sort of thing would you consider to be nicer? My feeling was that flattening it would make it the least ambiguous, or do you mean the way it is flattened/saved? (I'll add a comment about writing/reading anyway)

For the MiMA case we wrote out with the indices with the value ([e.g. see here](https://github.com/Cambridge-ICCS/fortran-pytorch-lib-benchmark/blob/main/input_data/psfc.txt)) to remove any ambiguity when reading back in and make interpretable.

examples/1_ResNet18/resnet18.py
TomMelt: ok thanks. I can see these values are specific to the pre-trained model and not the specific example of the dog image

ElliottKasoar:
At some point @TomMelt mentioned that it would be nice to have an assert to check that the results are actually consistent e.g. `call assert_real_2d(big_array, big_result/2., test_name=msg)` (currently unused) in the [resnet benchmarking branch](https://github.com/Cambridge-ICCS/fortran-pytorch-lib-benchmark/blob/melt/add-resnet-test/benchmarker_resnet_torch.f90#L68).

Given that the probabilities are more complex than that example, and seem to only match to ~5sf, is it worth just adding an assert for the label? Or maybe we should check the first 4-5 significant figures of the probability?

examples/1_ResNet18/README.md
```diff
@@ -77,8 +98,47 @@ export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:</path/to/library/installation>/lib64
 ./resnet_infer_fortran saved_resnet18_model_cpu.pt
 ```
 
+## Trying your own data
+
+Alternative images and labels can be tested by modifying the following:
```
jatkinson1000, line 103: Modifying it where?

examples/1_ResNet18/README.md
```diff
@@ -77,8 +98,47 @@ export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:</path/to/library/installation>/lib64
...
+
+In this example code, the NumPy array is transposed before being flattened and saved to binary, allowing Fortran to `reshape` the flatted array into the correct order.
+
+An alternative would be to save the NumPy array with its original shape, but perform a transpose during or after reading the data into Fortran, e.g. using:
+
+```
+in_data = reshape(flat_data, shape(in_data), order=(4,3,2,1))
+```
+
+For more general use, it should be noted that the function used to create the input tensor from `input_batch`, `torch_tensor_from_blob`, performs a further transpose, which is required to allow the tensor to interact correctly with the model.
```
jatkinson1000, line 144: Whilst useful to explain here, I think this needs to go somewhere much more prominent in this repository and documentation (and lots of thought given to communicating it, perhaps with Simon as well as us) as it'd the bit users find the most confusing, and is the most likely source of pitfalls.

Perhaps also add a note at the top that this is one of the learning objectives of this exercise.

examples/1_ResNet18/resnet18.py
```diff
@@ -1,20 +1,30 @@
...
 # Initialize everything
-def initialize():
+def initialize(wp: type) -> torch.nn.Module:
     """
     Download pre-trained ResNet-18 model and prepare for inference.
 
+    Parameters
+    ----------
+    wp: type
+        Data type of input tensor.
```
jatkinson1000, line 17: `wp` is not super obvious to me, modify docstring to 'sets the working precision of the model' or similar?

examples/1_ResNet18/resnet18.py
```diff
@@ -1,20 +1,30 @@
...
+import numpy as np
+from PIL import Image
 import torch
-import torch.nn.functional as F
 import torchvision
 
 
 # Initialize everything
-def initialize():
+def initialize(wp: type) -> torch.nn.Module:
```
jatkinson1000, line 10: I think `wp` is of type `torch.dtype`
This is the first instance I spotted, but also occurs elsewhere.

examples/1_ResNet18/README.md
jatkinson1000, line 103: Add hyperlink to the resnet images repo?

examples/1_ResNet18/README.md
```diff
@@ -77,8 +98,47 @@ export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:</path/to/library/installation>/lib64
 ./resnet_infer_fortran saved_resnet18_model_cpu.pt
 ```
 
+## Trying your own data
```
jatkinson1000, line 101: Maybe move this under the `## Further Options`?

jatkinson1000 (changes requested):
This is looking pretty good, nice work reading the categories etc. in to Fortran to make it a complete example!

Couple of comments as below.

Also, the python code could be more black, pylint, and mypy compliant.

examples/1_ResNet18/resnet18.py
ElliottKasoar: Thanks, should be fixed now.

examples/1_ResNet18/README.md
ElliottKasoar: I couldn't find an obvious ResNet specific set of images, but I've linked to the pytorch/vision repo where the example I give is from.

ElliottKasoar: Agree that this was unclear. I meant just run that code and change the URL if you want. Hopefully the combined set of changes for this section make it clearer.

ElliottKasoar:
> Also, the python code could be more black, pylint, and mypy compliant.

Was there anything obvious that you haven't mentioned? I'd already run it my main changes to resnet18.py through flake8 and black, and I think there were only a couple of minor things in resnet_infer_python.py, which I hadn't looked at enough yet (as part of the last set of changes, it's now updated to use the new image, as it was still using the `ones`.)

I've made a few more minor changes based on a few things from pylint/mypy too in the last commit, so hopefully it's ok now?

examples/1_ResNet18/README.md
ElliottKasoar: I'll definitely add the note towards the top about the learning objectives, particularly as I should comment on the relatively complexity now that your new example is in place.

Do you think adding these sorts of comments elsewhere in the repository is probably beyond the scope of this PR, especially if it could use considerable more thought and discussion?

jatkinson1000: Yes, long-form docs on this is definitely something for another PR.

Note at top is a good idea.

TomMelt:
> At some point @TomMelt mentioned that it would be nice to have an assert to check that the results are actually consistent e.g. `call assert_real_2d(big_array, big_result/2., test_name=msg)` (currently unused) in the [resnet benchmarking branch](https://github.com/Cambridge-ICCS/fortran-pytorch-lib-benchmark/blob/melt/add-resnet-test/benchmarker_resnet_torch.f90#L68).
> 
> Given that the probabilities are more complex than that example, and seem to only match to ~5sf, is it worth just adding an assert for the label? Or maybe we should check the first 4-5 significant figures of the probability?

The [`assert_real_2d`](https://github.com/Cambridge-ICCS/fortran-pytorch-lib-benchmark/blob/8c85bbab78a5719a91beb9ed51900894f7a66f8d/benchmarker_utils.f90#L7-L14) subroutine has an optional argument which allows you to change the relative tolerance. So I think a good idea would be to use `assert_real_2d` (or something like it) but just lower the tolerance to acknowledge that the variable in question has higher inaccuracy. 

**snippet of `assert_real_2d`**
```fortran

  subroutine assert_real_2d(a, b, test_name, rtol_opt)

    implicit none

    character(len=*) :: test_name
    real, intent(in), dimension(:,:) :: a, b
    real, optional :: rtol_opt
    real :: relative_error, rtol
```

jatkinson1000 (approved):
I'm happy with this now, and all running fine under checks for me.
I'll let @TomMelt take a look as well since it's been a large PR.

TomMelt (changes requested):
Looks great, just need to tidy up some fortran variable attributes

ElliottKasoar:
Thanks @TomMelt! I think addressed all of your Fortran variable suggestions.

I also wanted to double check you were happy with the changes just before/at the same time as your review along side the assertion checks, which included the addition of `wp` separately to `c_wp` (the possibility of which was part of the intention for `c_wp` in the first place), since I'd changed `probabilities` to be a normal Fortran real.

examples/2_ResNet18/resnet_infer_fortran.f90
```diff
@@ -1,62 +1,177 @@
...
+   end subroutine load_categories
+
+   subroutine calc_probs(out_data, probabilities, out_dims, out_shape)
+
+      implicit none
+
+      integer(c_int), intent(in) :: out_dims
+      integer(c_int64_t), intent(in) :: out_shape(out_dims)
+      real(c_wp), dimension(:,:), allocatable, target, intent(in) :: out_data
+      real(c_wp), dimension(:,:), allocatable, target, intent(inout) :: probabilities
```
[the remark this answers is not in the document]
ElliottKasoar: Ah I see, thanks! Done

examples/2_ResNet18/resnet_infer_fortran.f90
```diff
@@ -1,62 +1,177 @@
...
+
+   end subroutine load_data
+
+   subroutine load_categories(filename_cats, N_cats, categories)
+
+      implicit none
+
+      character(len=*), intent(in) :: filename_cats
+      integer, intent(in) :: N_cats
+      character(len=100), dimension(:), allocatable, target, intent(inout) :: categories
```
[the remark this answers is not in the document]
ElliottKasoar: Done

examples/2_ResNet18/resnet_infer_fortran.f90
```diff
@@ -1,62 +1,177 @@
...
+
+   end subroutine main
+
+   subroutine load_data(filename, N, in_data, in_dims, in_shape)
+
+      implicit none
+
+      character(len=*), intent(in) :: filename
+      integer, intent(in) :: N
+      real(c_wp), dimension(:,:,:,:), allocatable, target, intent(inout) :: in_data
```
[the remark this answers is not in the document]
ElliottKasoar: Done

examples/2_ResNet18/resnet_infer_fortran.f90
```diff
@@ -1,62 +1,177 @@
...
+      ! Set up types of input and output data and the interface with C
+      type(torch_module) :: model
+      type(torch_tensor), dimension(1) :: in_tensor
+      type(torch_tensor) :: out_tensor
+
+      real(c_wp), dimension(:,:,:,:), allocatable, target :: in_data
+      integer(c_int), parameter :: n_inputs = 1
+      real(c_wp), dimension(:,:), allocatable, target :: out_data
+      real(c_wp), dimension(:,:), allocatable, target :: probabilities
+      character(len=100), dimension(:), allocatable, target :: categories
```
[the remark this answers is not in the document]
ElliottKasoar: Done

TomMelt (approved):

PR #40, opened 2023-09-27
Add simple example

### LINKED ISSUE #39: Super simple example
At some point it would be nice to make an example in the examples repo that is as simple as possible (i.e. a 1-D input tensor etc.) so that the user can focus on the coupling procedure rather than any of the deeper challenges that may arise in later exercises (transposes/memory layouts, precision, etc.)

Candidates might be a 'Net' that simply multiplies by two etc.

I will try and do this at some point.

jatkinson1000:
Closes #39 

I have added the most basic (whilst remaining somewhat realistinc with a 'Net') example I could think of.

This should be an entry point for users who can see how to couple without having to worry about further details such as row/column major, transposes, stride access etc. which can be introduced in later examples.

At the moment I have made it exercise '0', but we should probably bump it to 1, and ResNet to 2 if we are happy.


examples/0_SimpleNet/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+- cmake
+- fortran compiler
+- FTorch (installed as described in main package)
+- python3
+
+## Running
+
+To run this example install fortran-pytorch-lib as described in the main documentation.
+Then from this directory create a virtual environment an install the neccessary python
```
ElliottKasoar, line 32: Very minor typo: neccessary -> necessary

ElliottKasoar (changes requested):
Looks great so far, I've just added a few minor comments.

examples/0_SimpleNet/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+Alternatively we can use `make`, instead of cmake, with the included Makefile.
+However, to do this you will need to modify `Makefile` to link to and include your
+installation of FTorch as described in the main documentation. Also check that the compiler is the same as the one you built the Library with.
+```
+make
+./simplenet_infer_fortran saved_simplenet_model_cpu.pt
+```
+
+You will also likely need to add the location of the `.so` or `.dylib` files to your `LD_LIBRARY_PATH`:
```
ElliottKasoar, line 82: It might be useful to clarify what is being referred to by  `.so` or `.dylib` files?

examples/0_SimpleNet/README.md
jatkinson1000: Done.

examples/0_SimpleNet/README.md
jatkinson1000: Done.

examples/0_SimpleNet/simplenet_infer_python.py
```diff
@@ -0,0 +1,56 @@
+"""Load SimpleNet saved to TorchScript and run inference with ones."""
```
ElliottKasoar, line 1: I think this file needs updating to the new input, as when I run `python3 simplenet_infer_python.py` I get:

```
  input_tensor = torch.ones(batch_size, 5)
tensor([[2., 2., 2., 2., 2.]])
```
rather than the new input/output tensors.

ElliottKasoar (changes requested):
Thanks for the changes!

Another non-breaking but odd thing is that I get `UserWarning: Failed to initialize NumPy: No module named 'numpy' (Triggered internally at ../torch/csrc/utils/tensor_numpy.cpp:84.)`, which I didn't think I got before.

examples/0_SimpleNet/simplenet_infer_python.py
jatkinson1000: Thanks, think that's fixed now -_-

ElliottKasoar (approved):
All looks good to me!

Final comments: we could add type-hinting to simplenet_infer_python.py, and a warning about the potential for a `"No module named 'numpy'"` warning when running simplenet.py, pt2ts.py or simplenet_infer_python.py, but I'm not sure either of these are essential.

jatkinson1000:
Thanks.
Added type hints (somehow missed those).
Numpy isn't used explicitly in any scripts (will be brought in as a dep of torch) and the issue doesn't raise for me.

I'll squash and merge now.

PR #44, opened 2023-10-05
Tidy example Fortran code

ElliottKasoar:
Tidying up Fortran code from the ResNet example, primarily to remove unused variables, as well as renaming `N` -> `tensor_length` to make it more meaningful and less likely to clash and `index` -> `idx`, to avoid overwriting a keyword.


TomMelt (approved):
great work. thank you

PR #46, opened 2023-10-11
Ensure strings end with c_null_char

ElliottKasoar:
See Cambridge-ICCS/FTorch-benchmarks#25

To do:

- [x] Test fixed on CSD3

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ This minimal snippet loads a saved Torch model, creates inputs consisting of two
 
 ```fortran
 ! Import any C bindings as required for this code
-use, intrinsic :: iso_c_binding, only: c_int, c_int64_t, c_null_char, c_loc
+use, intrinsic :: iso_c_binding, only: c_int, c_int64_t, c_loc
 ! Import library for interfacing with PyTorch
 use ftorch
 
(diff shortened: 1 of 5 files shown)


jatkinson1000:
A good start, perhaps need to apply `trim()` to the filename that users supply, just in case?
And do we need to handle the cases where an astute user might have already appended the null char to their filename?

ElliottKasoar:
Thanks @jatkinson1000! I've added a trim (and adjustl, as we have in the [benchmarker_utils](https://github.com/Cambridge-ICCS/FTorch-benchmarks/blob/main/benchmarker_utils.f90#L136)).

I haven't found anything definitive to back this up, but when I discussed this with @TomMelt, my impression was that repeated null characters shouldn't be an issue, and from testing this locally, and on CSD3, this seems to be the case - running the benchmarking script as is (with a mix of `c_null_char` included and not) linked to these changes all works.

jatkinson1000:
Yes, I think that is correct, just wanted to flag it as a point we needed to check.

jatkinson1000 (changes requested):
Thanks @ElliottKasoar 

The source changes all look good, but we ought to also update the example codes to reflect these changes.

ElliottKasoar:
Good point, thanks @jatkinson1000! I think I've removed c_null_char from all examples and the README now.

jatkinson1000 (approved):
C'est super! Allons-y.
Merci @ElliottKasoar

PR #47, opened 2023-10-11
Fix incorrect variable name

ElliottKasoar:
Bug fix for #44 to correct a missed variable change.


jatkinson1000 (approved):
Thanks @ElliottKasoar 🚀

PR #48, opened 2023-10-16
Update README.md to add windows docs

mondus:
Adds details of

- [x] How to install Intel Fortran in Windows
- [x] Details on obtaining the `CMAKE_PREFIX_PATH`
- [x] Cross platform build and install commands


README.md
```diff
@@ -73,27 +85,35 @@ To build and install the library:
...
+		  
+4. Make and install the code to the chosen location with either:
+    ```
+    cmake --build .
     ```
-    make
-    make install
+	or (to make and build)
+	```
+    cmake --build . --target INSTALL
```
jatkinson1000, line 109: Does the former just make/build and the latter makes/builds and then installs?
Is `INSTALL` a keyword for CMake - I have only ever see it used lower case - https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-a-project

jatkinson1000 (changes requested):
Thanks @mondus a couple of small tweaks and a question about the install command.
I've forwarded it to the user and will see how he gets on - using `cmake --build` rather than `make` may be important!

README.md
jatkinson1000: On reading looks like `install` is for makefile generators, whilst `INSTALL` is for GUI based systems: https://cmake.org/cmake/help/book/mastering-cmake/chapter/Install.html

Will update accordingly and then merge when appropriate.

jatkinson1000:
This is WIP with #50 
Once that is resolved with the user we will update here accordingly and then merging will close #50

jatkinson1000:
I think this is good to go now (correspondent from #50 has gone quiet, but it works for us).
Since I have done a lot of this PR I will let @TomMelt do a final review for approval.

README.md
```diff
@@ -197,18 +229,18 @@ against it.
...
 When compiling the final executable add the following link flag:
-```
+```makefile
 LDFLAGS += -L<path/to/install/location>/lib64 -lftorch
 ```
 
 You may also need to add the location of the `.so` files to your `LD_LIBRARY_PATH`
 unless installing in a default location:
-```
+```makefile
```
TomMelt, line 243: I think this should technically be "```bash"

TomMelt (dismissed):
looks good to me. Just a couple of questions

examples/2_ResNet18/README.md
```diff
@@ -71,7 +71,7 @@ This can be done using the included `CMakeLists.txt` as follows:
 mkdir build
 cd build
 cmake .. -DFTorch_DIR=<path/to/your/installation/of/library/>lib/cmake/ -DCMAKE_BUILD_TYPE=Release
-make
+cmake --build .
```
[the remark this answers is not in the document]
jatkinson1000: Yes.

README.md
jatkinson1000: Done

TomMelt:
just added a note about loading intel compilers. If @mondus is happy then I think we can merge it in.

jatkinson1000:
> just added a note about loading intel compilers. If @mondus is happy then I think we can merge it in.

@TomMelt Would it be worth extending that by telling users that this can usually be done by running `C:\Program Files (x86)\Intel\oneAPI\setvars` from Cmd?

README.md
```diff
@@ -46,64 +46,89 @@ To install the library requires the following to be installed on the system:
...
+3. Create a `build` directory and execute cmake from within it using the relevant options (see below):  
+    ```bash
     mkdir build
     cd build
     cmake .. -DCMAKE_BUILD_TYPE=Release
     ```
-    It is likely that you will need to provide at least the `CMAKE_PREFIX_PATH` flag.  
-    The following CMake flags are available and can be passed as arguments through `-D<Option>=<Value>`:
+    If building on Windows you will need to add `-G "NMake Makefiles"` to the `cmake` command. Also you will need to load the
+    intel fortran compilers using `setvars.bat` which is found in the Intel compiler install directory.
```
mondus, line 85:
```suggestion
    intel fortran compilers using `setvars.bat` ([see Intel docs](https://www.intel.com/content/www/us/en/docs/oneapi/programming-guide/2023-2/use-the-setvars-script-with-windows.html)) which is found in the Intel compiler install directory.
```

mondus (commented):
Tested in Windows. All good. Happy to approve. One minor suggestion comment to add a link to Intel guide on setvars.

jatkinson1000:
Thanks @mondus, that also addresses my comment above but in an even safer/better way than suggested, thanks.
I'll approve and merge now (@mondus has confirmed offline that he is happy with the PR).

jatkinson1000 (approved):
Thanks all who contributed here, I'm now happy to go ahead so will merge in a moment.

Issue #45, opened 2023-10-10
Misleading comment


ElliottKasoar:
Comments in ftorch.f90 describing the device to specify when running on GPUs suggest [torch_kGPU](https://github.com/Cambridge-ICCS/FTorch/blob/main/src/ftorch.f90#L54), as opposed to the correct [torch_kCUDA](https://github.com/Cambridge-ICCS/FTorch/blob/main/src/ctorch.h#L29).

jatkinson1000:
Thanks Elliot - this is why testing in practice is better than 'works in theory'!!
If you can correct with a quick PR that would be great.
I'm intrigued to see how one might get the mps or xpu backends to work...

ElliottKasoar:
Resolved by #49

PR #49, opened 2023-10-16
Update for GPUs

### LINKED ISSUE #9: Test on GPUs
Run the testing suite on a GPU system. Requires saving the torch script a bit differently? @jatkinson1000

ElliottKasoar:
Resolves #9 and #45

Initial changes to allow FTorch to run on GPUs.

As discussed in #9, changes for CMake may not be required, but may be useful.

Additional documentation to be added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,11 +87,11 @@ To build and install the library:
           If Torch has been installed as PyTorch in a python [venv (virtual environment)](https://docs.python.org/3/library/venv.html),
           e.g. with `pip install torch`, then this should be `</path/to/venv/>lib/python<3.xx>/site-packages/torch/`._
 
-    <sup>2</sup> _This is often overridden by PyTorch. To ensure a CPU or GPU only version, use one of:
+    <sup>2</sup> _This is often overridden by PyTorch. When installing with pip, the `index-url` flag can be used to ensure a CPU or GPU only version is installed, e.g.
           `pip install torch --index-url https://download.pytorch.org/whl/cpu`
           or
           `pip install torch --index-url https://download.pytorch.org/whl/cu118`
-          (for CUDA 11.8) respectively._
+          (for CUDA 11.8). URLs for alternative versions can be found [here](https://pytorch.org/get-started/locally/)._
 
 
 4. Make and install the code to the chosen location with:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -79,12 +79,21 @@ To build and install the library:
...
           If Torch has been installed as PyTorch in a python [venv (virtual environment)](https://docs.python.org/3/library/venv.html),
           e.g. with `pip install torch`, then this should be `</path/to/venv/>lib/python<3.xx>/site-packages/torch/`._
+
+    <sup>2</sup> _This is often overridden by PyTorch. To ensure a CPU or GPU only version, use one of:
+          `pip install torch --index-url https://download.pytorch.org/whl/cpu`
+          or
+          `pip install torch --index-url https://download.pytorch.org/whl/cu118`
+          (for CUDA 11.8) respectively._
+
+
```
jatkinson1000, line 96: Is this related to the [Alternative versions](https://pytorch.org/get-started/locally/) linked above?

README.md
```diff
@@ -212,6 +221,26 @@ unless installing in a default location:
 export LD_LIBRARY_PATH = $LD_LIBRARY_PATH:<path/to/installation>/lib64
 ```
 
+### 4. Running on GPUs
+
+In order to run a model on a GPU, a two main changes are required:
```
jatkinson1000, line 226:
```suggestion
In order to run a model on GPU, a two main changes are required:
```

README.md
```diff
@@ -212,6 +221,26 @@ unless installing in a default location:
...
+trained_model.eval()
+trained_model_dummy_input_1 = trained_model_dummy_input_1.to(device)
+trained_model_dummy_input_2 = trained_model_dummy_input_2.to(device)
+```
+
+Note: this also moves the dummy input tensors to the GPU. This is not necessary for saving the model, but the tensors must also be on the GPU to test that the models runs.
+
+
+2. When calling `torch_tensor_from_blob` in Fortran, the device for the input tensor(s), but not the output tensor(s),
+   should be set to `torch_kCUDA`, rather than `torch_kCPU`. This ensures that the inputs are on the same device as the model.
```
jatkinson1000, line 242: This will hopefully get put into some formal API docs at some point, but good to have here for now.

jatkinson1000 (approved):
Thanks @ElliottKasoar, this is great.
I am assuming you have checked, but would be interested in going through a GPU run with you on Thursday perhaps.

Setting to approve, but consider adjusting the `CMAKE_CUDA` flag name.

README.md
ElliottKasoar: Thanks. I did a different commit rather than the suggestion to remove the second `a` at the same time.

Issue #57, opened 2023-10-31
Incorrect example in README


jatkinson1000:
Related to #56 The example code in the README is now incorrect as it has not incorporated the updates to use the `layout` argument in `torch_tensor_from_blob()`

We should either:
- Update it to use `t_t_from_array()`
- Update to add a `layout` argument

My preference is for the first (see discussion on #56).

However, this information should also be taken out of the README and placed in the longer form API docs now, perhaps as part of #53

Issue #13, opened 2023-05-26
Example for multiple inputs


jatkinson1000:
We should provide an example of using multiple inputs to the model.

Is there a pre-existing trained net we can deploy?
Davenet?

Do this after restructuring examples for #12 

Also a test?

Issue #56, opened 2023-10-31
Simplify stride interface with `F` or `C` layout


ElliottKasoar:
This is not a high priority at the moment, but as discussed in #24, it would be good to improve the user friendliness of the stride functionality, through a `F` or `C` layout style and shape/size inference.

This should also be documented in (and potentially help clarify) the discussion about when to transpose arrays, in both the ResNet example and (soon) the long-form documentation.

jatkinson1000:
If you look at the `ftorch.f90` source there is a function `t_t_from_array()` (which should probably become `torch_tensor_from_array`) that assumes `'F'` indexing and removes the need for the user to pass in the `layout` argument (used to set `strides`).

The idea of this is that is creates a tensor that can be read in with the same indexing in Torch without the user having to worry about it.

1. This should be the default option to users and we should be VERY EXPLICIT in the docs that using this function you assume the same indexing in Fortran and Torch.
2. We should look at if we can simplify the examples, making the use of `torch_tensor_from_blob` with `layout` an _'advanced'_ option
3. This should be clearly set out in the README and API docs (see #53)

jatkinson1000:
This was discussed during the work on #604 and it was decided that we would leave it as a permute argument with decision left clearly up to the user to say what they are doing, thather than obfuscate any intent with an `F` or `C` default that could be interpreted two ways.

As such I am closing this now, but we can revisit in the future if we feel there is a demand for it.

PR #53, opened 2023-10-20
Add Long-form API documentation

jatkinson1000:
Add long form API documentation using FORD

- [ ] Create GH action
- [ ] Creage GH page under ICCS
- [ ] Run FORD on code
- [ ] Update source with comments for FORD
- [ ] Move action to rebuild docs for main, not branch

PR #60, opened 2023-11-03
add mwe for fypp preprocessor

### LINKED ISSUE #58: Triage torch_tensor_from_array
We need to decide if it is best to:

- require users to specify the dtype when calling `torch_tensor_to_array()`
- Stick with the abstract interface approach currently there.

If the latter then interface needs expanding beyond the current 'float' and 'double'.

I can see an argument for the former, however, as it
- a) simplifies the code (one routine and no interface), and
- b) forces the user to be explicit about the data type they are using which, as we have seen, causes issues if incorrect.

Discussion appreciated, then we need to reach a decision and triage:
- [ ] Amend source as required
- [ ] Update docs as required
- [ ] Update examples as required
- [ ] Update benchmarks as required

TomMelt:
This PR introduces a _fortran only_ version of ftorch interface that uses interface function `torch_tensor_from_array` to create a tensor object from any fortran array.

Instructions to generate pre-processed file:
1. [install fypp](https://fypp.readthedocs.io/en/stable/fypp.html#installing)
1. run `fypp` - `fypp ftorch.fypp ftorch.f90`

This will replace the existing `ftorch.f90`. Choose a different name if you want to compare both.

For production ready version we can use the suggestions [here](https://fypp.readthedocs.io/en/stable/fypp.html#cmake) to integrate as part of our code base.

- [x] update docs
- [x] expand _fortran only_ interface to all examples in repo (maybe copy so we can show both approaches)
- [x] expand _fortran only_ interface to all examples in ftorch benchmarks?
- [x] check timings don't get worse
- [x] check compatibility with gpu version?
- [x] decide where to put `.fypp` and `.f90` (generated) files (My preference is to keep both in `src` folder)
- [ ] add example `cgdrag` showing how to use c-only features (#13)

When this is finished it will close #58


TomMelt:
I need to make a fortran-only interface with a mwe that removes the need for the end user to import `iso_c_bindings`

ElliottKasoar:
(Following earlier discussion with @jatkinson1000 and @TomMelt where I think we concluded that we want to store the processed .f90 files, as well as the .fypp files)

I'm not sure if you had something like this in mind already, but we could potentially do something similar to [cp2k](https://www.cp2k.org/dev:fypp), which seems to place the compiled .f90 files in an /obj directory, potentially making the /src directory less cluttered.

(I haven't seen anywhere actually store the pre-processed files with git, but I think the idea still works.)

jatkinson1000:
Yes, #61 opened to log this proposal.

TomMelt:
@jatkinson1000 @ElliottKasoar , ok I think I have done it. I have a _fortran only_ version of ftorch interface that uses interface function `torch_tensor_from_array` to create a tensor object from any fortran array.

Can you please both take a look? And feel free to try it out on your machines too.

**Caveats**
* `iso_fortran_env` only supports the following types: `int8, int16, int32, int64, real32, real64` not the full list of torch enums e.g, `torch_kUInt8` and `torch_kFloat16`. I have therefore not generated overloaded functions for these data types.
* I have only converted and tested the `resnet_infer_fortran.f90` example. We should look at converting the others and testing before I merge this.
* Also I would like to make sure this doesn't make timings worse :crossed_fingers:

jatkinson1000 (commented):
Thanks for this @TomMelt.

This is only an interim review as there's a lot to unpack here.
I thought I'd voice things in stages as I work through them.

So far I have:
- Compiled and built this version of FTorch on M2 Mac (gfortran, Apple clang)
- Build and run exercise 2 (Resnet) using your updated code and this version of the library
- Generated FORD docs for this updated version of the libaray

Observations:
- Documentation is now very messy - I'm looking if FORD can generate interface docs.
- I would suggest we rename the functions from e.g. `torch_tensor_from_array_int16_1` to `torch_tensor_from_array_int16_1d` as I think it makes it clearer what the suffix is.
- [ ] Documentation needs updating with fypp install/run instructions
- [ ] We discussed a git hook to generate the ftorch

Other comments:
- [ ] I agree with not supporting non-Fortran types, but we should note this in longer docs
- I agree with fypp and f90 being alongside one another in src/
- Perhaps not for this PR, but we may consider a short explanation of the `sp` precision syntax.

I'll start a closer look at the code now and make a few additions as suitable.

jatkinson1000:
Having been through and generated a pre-commit hook to check fypp is generated accordingly I realised I forgot that pre-commit hooks cannot be shared between users or stored as part of the git repo. Therefore I will add an action to check instead, and instructions for developers to add the pre-commit hook manually.

jatkinson1000:
Now happy with the workflow.
Slightly watered down from the original proposal, but the best we can do with GitHub I think.
It compares the result of fypp generated code to the existing `ftorch.f90`, and fails if they do not match.

This does not stop people from pushing mismatching code or modifying `ftorch.f90` directly, but such a situation should become apparent when conducting a code review and this check will ensure the pair of files match as a first line of defense.

I have included a more powerful pre-commit hook in the main repo (could move elsewhere?) that will enforce these things for developers. I am happy (with someone else checking) to say this now closes #61

jatkinson1000:
Updated the README such that merging this will also close #57 
Would appreciate a second pair of eyes on this.

This will need amending further if we make a decision about simplifying the indexing as well as per #56

jatkinson1000:
Since fypp is not a required dependency to run I suggest we move description of how to install this (and use the git hook) to a 'developers'' section of the documentation in #53

jatkinson1000:
I'm going to pause here for now, but it would be great if we can get this merged in time for the talk!
Doing so I think requires:
- [ ] Updating examples (I agree with preserving one to demonstrate using blob)
- [ ] Updating benchmarks and timing checks (@ElliottKasoar)
- [ ] Checking GPU functionality (@ElliottKasoar)

We should also consider if we want to make the strides an optional argument (#56), but given the size of this PR this should perhaps be done as another feature. Changing them to be optional should be possible in a backwards compatible way I'd hope.

Once this is clear I'd like to complete #53 in some form, maybe #48, and try to get some MiMA numbers.
And sort some slides I guess...

ElliottKasoar (commented):
This isn't entirely newly introduced by these changes (e.g. [torch_tensor_from_blob](https://github.com/Cambridge-ICCS/FTorch/blob/main/src/ftorch.f90#L48) vs [torch_from_blob_C](https://github.com/Cambridge-ICCS/FTorch/blob/main/src/ftorch.f90#L62)) but something to consider is adding a bit more consistency between the order of parameters.

For example, `layout` and `device` are switched changing from `torch_tensor_from_blob(data, ndims, tensor_shape, dtype, device, layout)` to `torch_tensor_from_array(data_in, layout, c_device)`.

Particularly for integer arguments, this could be easy to miss.

jatkinson1000:
Great point, and I agree @ElliottKasoar

ElliottKasoar:
> I'm going to pause here for now, but it would be great if we can get this merged in time for the talk! Doing so I think requires:
> 
> 
>     * [ ]  Updating benchmarks and timing checks (@ElliottKasoar)
> 
>     * [ ]  Checking GPU functionality (@ElliottKasoar)
> 
> 
> We should also consider if we want to make the strides an optional argument (#56), but given the size of this PR this should perhaps be done as another feature. Changing them to be optional should be possible in a backwards compatible way I'd hope.
> 
> Once this is clear I'd like to complete #53 in some form, maybe #48, and try to get some MiMA numbers. And sort some slides I guess...

Changes tested on CSD3 GPUs, and no apparent change in timings seen on CPU or GPU.

My PR will also hopefully have an example of [cgdrag](https://github.com/Cambridge-ICCS/FTorch-benchmarks/blob/time-full-program/benchmark_cgdrag/benchmarker_cgdrag_torch.f90) with both `from_array` (explicit reshaping) and `from_blob` (implicit reshaping), so that can potentially be adapted/used to inform new examples.

jatkinson1000 (changes requested):
I'm now happy with this, but agree we should standardise inputs to functions.
@ElliottKasoar @TomMelt How do we feel about the following precedence order:
- input data
- device
- layout
- ndims
- tensor_shape
- dtype

This keeps required first, and anticipates future development when `layout` may become optional.

I'd also suggest the 'c-only-example' becomes its own pull request and we merge this once the above is done.

ElliottKasoar:
> I'm now happy with this, but agree we should standardise inputs to functions. @ElliottKasoar @TomMelt How do we feel about the following precedence order:
> 
> * input data
> * device
> * layout
> * ndims
> * tensor_shape
> * dtype
> 
> This keeps required first, and anticipates future development when `layout` may become optional.
> 
> I'd also suggest the 'c-only-example' becomes its own pull request and we merge this once the above is done.

I think @TomMelt mentioned to me earlier deliberately putting `device` at the end, as it's a bit more detached from specification of the tensor (please correct me if I'm misquoting), but I think I agree with @jatkinson1000 that the more optional the argument, the lower the precedence, probably makes more sense.

jatkinson1000:
OK, that makes sense.
I have standardised, so how does that look now?
Kept device at the end which I think makes sense (this is also what Torch does).

ElliottKasoar (changes requested):
I think we discussed this, but is there a reason to keep test_tensor.f90 still?

I would maybe either update the instances of [torch_tensor_from_blob](https://github.com/Cambridge-ICCS/FTorch/blob/mwe-fypp-preprocessor/examples/tensor_tests/test_tensor.f90#L40) or delete it, but otherwise looks great!

pre-commit
ElliottKasoar, line 1: I'm wondering if this should be in utils, especially since most users don't need to worry about it?

pre-commit
```diff
@@ -0,0 +1,69 @@
+#!/bin/sh
+#
+# A hook script to verify what is about to be committed.
+# Called by "git commit" with no arguments.  The hook should
+# exit with non-zero status after issuing an appropriate message if
+# it wants to stop the commit.
+
+# Fail immediately at first issue with the relevant exit status.
+set -eo pipefail
```
ElliottKasoar, line 9: I seem to get an error using this (after moving it into /.git/hooks): `.git/hooks/pre-commit: 9: set: Illegal option -o pipefail`. Any idea why?

ElliottKasoar: Should it be `#!/bin/bash` rather than `#!/bin/sh`?

pre-commit
jatkinson1000: I also thought that, but then decided `utils/` was more for toold and utilities that most users _do_ want.
I'm not sure top is the best place for it, but also not sure it's `utils/`.

ElliottKasoar: Hmm true, maybe something like a `.githooks` directory?

I haven't really come across hooks before, but I take it you haven't come across any particular convention?

pre-commit
ElliottKasoar: Removing the line doesn't seem to change the error, unfortunately.

(Changing it to `#!/bin/bash` did work)

It definitely needs execution rights, as removing them (`chmod -x`) meant that `git commit` ignored the hook, so we definitely need to ensure it's present, but it doesn't fix the option error.

jatkinson1000: I think let's make it `#!/bin/bash` in that case.

pre-commit
jatkinson1000: That seems a sensible suggestion to me. Done.

ElliottKasoar (changes requested):
I'm not sure the .githooks directory was added in your last commit @jatkinson1000.

jatkinson1000:
> I'm not sure the .githooks directory was added in your last commit @jatkinson1000.

Uuurrgghhhhhh I just used `mv` instead of `git mv`.
Thank you.
Should be sorted now.

ElliottKasoar (approved):

TomMelt:
> OK, that makes sense. I have standardised, so how does that look now? Kept device at the end which I think makes sense (this is also what Torch does).

I think we need the following order for `torch_tensor_from_blob`:
- input data
- ndims
- tensor_shape
- layout/strides
- dtype
- device

This will then match the order of args in `torch_from_blob_c`
```fortran
function torch_from_blob_c(data, ndims, tensor_shape, strides, dtype, device) result(tensor_p)
```
and those in `torch_tensor_from_array`
```fortran
function torch_tensor_from_array(data_in, layout, c_device) result(tensor)
```

What do you think @jatkinson1000 ? If you are happy I have a commit waiting to be pushed that fixes it

jatkinson1000:
That sounds sensible to me!

Issue #65, opened 2023-11-10
Simple linting and QC


jatkinson1000:
We should add some basic linting and QC on the repository to help guide PRs to meet some basic standards in advance.

For now we should do this for python.
Apply:
- [x] black
- [x] pydocstyle --numpy

Consider:
- [ ] mypy - may not want this as not great on standalone scripts and want to keep exercises simple
- [ ] pylint - we may not be compliant... 😅

PR #66, opened 2023-11-10
Python QC workflow

jatkinson1000:
Addresses #65 

Checks that all python code in the repo is black and numpy pydocstyle compliant.
Does NOT change code, instead presents a ❌ to the user to fix and recommit.


.github/workflows/python_qc.yaml
```diff
@@ -0,0 +1,32 @@
...
+    name: Python Code QC (Black, pydocstyle)
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
+      with:
+        python-version: "3.11"
+    - run: pip install black pydocstyle
+
+    # annotate each step with `if: always` to run all regardless
```
ElliottKasoar, line 26: Just to check, does this mean pydocstyle will still run if black fails (a bit like `fail-fast: false` for a matrix?)

ElliottKasoar (approved):
Looks great, thanks for this!

.github/workflows/python_qc.yaml
jatkinson1000: Indeed.

PR #67, opened 2023-11-19
CMake fix

### LINKED ISSUE #14: CMake Improvements
We should add some CMake flags to improve the build experience this should include

 - Examples: A flag for each to enable building
 - Documentation: We should remove the dependencies for documentation building if only building the library
 - Testing: We should remove the dependencies for testing if only building the library

jatkinson1000:
I believe this closes #14 

Namely the comments below the main one regarding install location.
cmake files are now placed at `$CMAKE_INSTALL_PREFIX/lib/cmake/FTorch/` allowing CMake to locate things automatically if passed `CMAKE_PREFIX_PATH = $CMAKE_INSTALL_PREFIX`.

This means we no longer specify the `FTorch_DIR` argument, instead using `CMAKE_PREFIX_PATH` as standard.

Regarding the other items in #14:
- Documentation has no dependencies as it is deployed by a GitHub action
- Examples exist as standalone items with their own CMake - I would argue that they should not in any way be combined with the main CMakeLists
- Currently no testing dependencies (or testing 😬) but we can make this a flag option when we write them.

Tested on apple M1 mac to build library, install, and build/run examples 1 and 2.
Would appreciate someone else replicating on a different system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ To build and install the library:
     ```
     This will place the following directories at the install location:  
     * `CMAKE_INSTALL_PREFIX/include/` - contains header and mod files
-    * `CMAKE_INSTALL_PREFIX/lib64/` - contains `cmake` directory and `.so` files
+    * `CMAKE_INSTALL_PREFIX/lib/` - contains `cmake` directory and `.so` files
 
 
 ## Usage
@@ -206,13 +206,13 @@ FCFLAGS += -I<path/to/install/location>/include/ftorch
 
 When compiling the final executable add the following link flag:
 ```
-LDFLAGS += -L<path/to/install/location>/lib* -lftorch
+LDFLAGS += -L<path/to/install/location>/lib -lftorch
 ```
 
 You may also need to add the location of the `.so` files to your `LD_LIBRARY_PATH`
 unless installing in a default location:
 ```
-export LD_LIBRARY_PATH = $LD_LIBRARY_PATH:<path/to/install/location>/lib*
+export LD_LIBRARY_PATH = $LD_LIBRARY_PATH:<path/to/install/location>/lib
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -189,7 +189,10 @@ find_package(FTorch)
 target_link_libraries( <executable> PRIVATE FTorch::ftorch )
 message(STATUS "Building with Fortran PyTorch coupling")
 ```
-and using the `-DFTorch_DIR=</path/to/install/location>` flag when running cmake.
+and using the `-DCMAKE_PREFIX_PATH=</path/to/install/location>` flag when running cmake.  
+Note: If you used the `CMAKE_INSTALL_PREFIX` argument when
+[building and installing the library](#libaray-installation) above then you should use
+the same path for `</path/to/install/location>`.
 
 #### Make
 To build with make we need to include the library when compiling and link the executable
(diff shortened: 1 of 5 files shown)


ElliottKasoar:
Both examples seem to work for me on WSL (Ubuntu 20.04)

I assume we're not updating the work-in-progress example for now?

Worth noting we'll need to update the benchmarking repo too.

README.md
```diff
@@ -203,13 +206,13 @@ FCFLAGS += -I<path/to/install/location>/include/ftorch
 
 When compiling the final executable add the following link flag:
 ```
-LDFLAGS += -L<path/to/install/location>/lib64 -lftorch
+LDFLAGS += -L<path/to/install/location>/lib* -lftorch
```
SimonClifford, line 209: Can you do this?  I know it's a suggestion in documentation but if a user puts a `*` there expecting it to work, I don't think it will, will it?

README.md
```diff
@@ -203,13 +206,13 @@ FCFLAGS += -I<path/to/install/location>/include/ftorch
...
 ```
 
 You may also need to add the location of the `.so` files to your `LD_LIBRARY_PATH`
 unless installing in a default location:
 ```
-export LD_LIBRARY_PATH = $LD_LIBRARY_PATH:<path/to/installation>/lib64
+export LD_LIBRARY_PATH = $LD_LIBRARY_PATH:<path/to/install/location>/lib*
```
SimonClifford, line 215: As above.

SimonClifford (commented):
Not sure about glob patterns in `LDFLAGS` suggestions in README.md

README.md
jatkinson1000: Thanks.
I wasn't too sure about this as it seems system-dependent.
Sometimes it's `lib`, sometimes `lib64` etc.
I meant to use `*` as 'there may be something or nothing here' like the CMake documentation does, but I agree that isn't clear.
Do you have any suggestions?

TomMelt (changes requested):
looks good but not happy with *

README.md
SimonClifford: But these are the directions for Make.  I _think_ if you say `/foo/bar/lib*' you'll add a directory called `/foo/bar/lib*` (because `*` is a legitimate character in pathnames!).

I think you can specify `lib` and maybe add a comment that says:
`# or lib64` or something like that.

README.md
```diff
@@ -189,7 +191,10 @@ find_package(FTorch)
 target_link_libraries( <executable> PRIVATE FTorch::ftorch )
 message(STATUS "Building with Fortran PyTorch coupling")
 ```
-and using the `-DFTorch_DIR=</path/to/install/location>` flag when running cmake.
+and using the `-DCMAKE_PREFIX_PATH=</path/to/install/location>` flag when running cmake.  
+Note: If you used the `CMAKE_INSTALL_PREFIX` argument when
```
ElliottKasoar, line 195: Minor/stylistic but maybe add an indent/italic as we do elsewhere for notes?

README.md
```diff
@@ -189,7 +191,10 @@ find_package(FTorch)
...
+[building and installing the library](#libaray-installation) above then you should use
```
ElliottKasoar, line 196:
```suggestion
[building and installing the library](#library-installation) above then you should use
```

ElliottKasoar (changes requested):

ElliottKasoar (approved):

TomMelt (approved):

PR #69, opened 2023-11-22
Patch on example utils files.

jatkinson1000:
Remove redundant lines that caused errors when uncommented running on GPU.


ElliottKasoar (changes requested):
Ah yes, thanks for spotting this!

I got rid of  `_1` etc. and moved the filename into the same block as the rest of the GPU changes for the [benchmarking changes](https://github.com/Cambridge-ICCS/FTorch-benchmarks/blob/main/resnet_model/pt2ts.py), but I think your change here is sufficient.

Just noticed that both of the changed files here (and in the resnet benchmark) actually still have ` # This example assumes two inputs of size (512x40) and (512x1)` which we probably ought to remove.

ElliottKasoar (approved):
Looks good to me, thanks!

PR #70, opened 2023-11-22
Increase assertion tolerance for Python inference

ElliottKasoar:
Increase the tolerance slightly to match the example Fortran tolerance, as the results may only match to around five significant figures.

(Without this change, it threw an assertion error for me when running `python resnet_infer_python.py`)


jatkinson1000 (commented):
I think this is probably OK.
Does it need changing on exercise 1 as well?

jatkinson1000 (approved):
Just checked, and the answer is no, as we only print for the user to check there.
Will merge now.

PR #75, opened 2023-11-28
Update README.md to give some extra guidance to Mac OS X users 

dorchard:
This is something I noticed when building on my M1 Mac. I've added a little guidance (a note) to suggest the most likely succesfull path to building for other such users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,8 @@ To build and install the library:
           If Torch has been installed as PyTorch in a python [venv (virtual environment)](https://docs.python.org/3/library/venv.html),
           e.g. with `pip install torch`, then this should be `</path/to/venv/>lib/python<3.xx>/site-packages/torch/`.  
 		  You can find the location of your torch install by importing torch from your python environment (`import torch`) and running `print(torch.__file__)`_
+
+	  Note __Mac OS X users__ are more likely to have success using the approach of pointing cmake at python install of PyTorch.
 		  
 4. Make and install the library to the desired location with either:
 	```bash
--- a/README.md
+++ b/README.md
@@ -57,6 +57,9 @@ If building in a windows environment then you can either:
 
 Note that libTorch is not supported for the GNU Fortran compiler with MinGW.
 
+#### Apple Silicon Support
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -104,6 +104,8 @@ To build and install the library:
           If Torch has been installed as PyTorch in a python [venv (virtual environment)](https://docs.python.org/3/library/venv.html),
           e.g. with `pip install torch`, then this should be `</path/to/venv/>lib/python<3.xx>/site-packages/torch/`.  
 		  You can find the location of your torch install by importing torch from your python environment (`import torch`) and running `print(torch.__file__)`_
+
+	  Note __Mac OS X users__ are more likely to have success using the approach of pointing cmake at python install of PyTorch.
```
ElliottKasoar, line 108: Possibly unnecessarily specific, but it might be worth clarifying that this applies to ARM-based Macs specifically (I assume).

I would maybe also consider moving this to a (very short!) macOS support section, similar to for [Windows](https://github.com/Cambridge-ICCS/FTorch/tree/dorchard-patch-1#windows-support), as otherwise this is very easy to miss, with pip-installed PyTorch as the recommended/tested approach (pending @jatkinson1000's testing)?

jatkinson1000 (changes requested):
It looks like libtorch is currently only available in binary form for x86 architectures according to https://pytorch.org/ (though there are some unofficial binaries going around) which I think is the root of the issues here. It's strange, as there must be an arm build that comes as part of the pip install.

As such I don't think this should be communicated in its current form at the moment as there could well be an arm binary in future.
I would agree with @ElliottKasoar that it is worth perhaps noting in an "Apple Silicon Support" section that at time of writing there is no binary and a pip install of torch is required for apple silicon macs.

Related - I am interested to see if it is possible to use MPS from libtorch.

dorchard:
Thanks. I've made an "Apple Silicon support" section similar to the Windows one that explains the situation in more detail.

jatkinson1000 (approved):
Yep, I am happy with this now.
Since it matches @ElliottKasoar's suggestion I will approve and merge now.
Worth keeping in mind to amend if an arm binary does appear in future.

PR #78, opened 2023-12-05
Add options for eval and gradient required

### LINKED ISSUE #73: Loaded TorchScript missing no_grad context
When using `jit.script`, it appears that `.eval` and `no_grad` context is [not saved](https://discuss.pytorch.org/t/does-torch-jit-script-requires-no-grad-and-eval-modes/132181/6).

This is likely to be partially responsible for worse performance than expected during inference.

My current solution has been to add:
```
model->eval();
torch::NoGradGuard no_grad;
```
in `torch_jit_module_forward`, as adding these in `torch_jit_load` did not appear to change the behaviour.

As it is also possible that these functions may be used in training, adding these settings conditionally would also be preferable.

ElliottKasoar:
Resolves #73

Adds flags in all(?) functions that operate on tensors (tensor creation, model loading, forward) to optionally disable autograd, which should improve performance for inference.

Also adds a similar flag to set evaluation mode for the loaded model.

- Evaluation mode
  - Contrary to my initial comments in #73, from testing evaluation mode does appear to be preserved, both between saving and loading TorchScript, and when applied to the loaded model.
  - In most cases evaluation mode is therefore likely to already be set, but I think it's useful to have the option to change it, particularly if FTorch may be extended to facilitate training (#22).

- NoGradMode
  - Enabling or disabling gradients is more complicated, as it defined via a context manager, which only appears to define the behaviour within its own scope, and so it seems necessary to enable/disable gradients before every code block that operates on tensors (similar to the Python equivalent `with torch.no_grad():`).

- InferenceMode
  - No changes are currently included, but it would be good to support [InferenceMode](https://pytorch.org/cppdocs/notes/inference_mode.html) too eventually, as it should provide further performance benefits over NoGradMode.
  - However, it has stricter requirements, and the mode was only added (as a beta) in PyTorch 1.9, so we would need to be careful if we want to support older versions.

- Model freezing
  - No changes are currently included, and less directly applicable to the main FTorch library, although there are still interactions e.g. freezing the model can allow InferenceMode to be enabled when loading the model.
  - Freezing is currently the "default" when tracing in [pt2ts.py](https://github.com/Cambridge-ICCS/FTorch/blob/main/utils/pt2ts.py), but not for scripting, despite potentially [improving performance](https://pytorch.org/tutorials/prototype/torchscript_freezing.html).
  - Freezing appears to (sometimes) introduce numerical errors when saving the reloading (differences ~10^-6), and can seem to lead to issues loading with Forpy too.

(For more general explanation of autograd/evaluation mode, see [autograd mechanics](https://pytorch.org/docs/1.9.0/notes/autograd.html)).

Note: I've also removed the old, commented out `torch_from_blob` function.

### DIFF
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -43,6 +43,16 @@ constexpr auto get_device(torch_device_t device)
   }
 }
 
+void set_is_training(torch_jit_script_module_t module, const bool is_training)
+{
+  auto model = static_cast<torch::jit::script::Module*>(module);
+  if (is_training) {
+    model->train();
+  } else {
+    model->eval();
+  }
+}
+
 torch_tensor_t torch_zeros(int ndim, const int64_t* shape, torch_data_t dtype,
                            torch_device_t device, const bool requires_grad)
 {
@@ -171,11 +181,7 @@ torch_jit_script_module_t torch_jit_load(const char* filename,
     delete module;
     exit(EXIT_FAILURE);
   }
-  if (is_training) {
-    module->train();
-  } else {
-    module->eval();
-  }
+  set_is_training(module, is_training);
 
   return module;
 }
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -44,8 +44,9 @@ constexpr auto get_device(torch_device_t device)
 }
 
 torch_tensor_t torch_zeros(int ndim, const int64_t* shape, torch_data_t dtype,
-                           torch_device_t device)
+                           torch_device_t device, const bool requires_grad)
 {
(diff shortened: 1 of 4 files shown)


TomMelt (approved):
thanks @ElliottKasoar , these changes look great to me

jatkinson1000:
I'll review this in the new year.
I see @TomMelt is happy, but there are a lot of changes so I also want to be sure I've looked through them.

jatkinson1000:
A new PR rebasing this work on top of the latest changes to resolve conflicts is at #103

jatkinson1000:
Since #103 is progressing I am closing this.

PR #87, opened 2024-01-02
Make fypp CI run on PRs from forks.

jatkinson1000:
FYPP check isn't running in forks as set to:
```
on:
    push:
```
which only works for pushes to the main repo.

This updates it to also run on pull requests from forks.

Means it will run twice on internal PRs but we can live with that.


TomMelt (approved):
LGTM. Thanks

PR #91, opened 2024-03-18
Remove unused code

### LINKED ISSUE #89: Remove unused code
The following code is commented out and should be removed. Similarly, for the `torch_from_blob_f` header, which is unused.

https://github.com/Cambridge-ICCS/FTorch/blob/2653bf80d0b807786df72e9ca701208e3ba8f510/src/ctorch.cpp#L112-L138

jwallwork23:
Closes #89.

Drops commented out `torch_from_blob` C++ code and unused `torch_from_blob_f` header.

### DIFF
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -109,33 +109,6 @@ torch_tensor_t torch_empty(int ndim, const int64_t* shape, torch_data_t dtype,
   return tensor;
 }
 
-/*
-// Exposes the given data as a Tensor without taking ownership of the original
-// data
-torch_tensor_t torch_from_blob(void* data, int ndim, const int64_t* shape,
-                               torch_data_t dtype, torch_device_t device)
-{
-  torch::Tensor* tensor = nullptr;
-  try {
-    // This doesn't throw if shape and dimensions are incompatible
-    c10::IntArrayRef vshape(shape, ndim);
-    tensor = new torch::Tensor;
-    *tensor = torch::from_blob(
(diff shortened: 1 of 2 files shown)


src/ctorch.cpp
```diff
@@ -109,33 +109,6 @@ torch_tensor_t torch_empty(int ndim, const int64_t* shape, torch_data_t dtype,
...
-  } catch (const std::exception& e) {
-    std::cerr << "[ERROR]: " << e.what() << std::endl;
-    delete tensor;
-    exit(EXIT_FAILURE);
-  }
-  return tensor;
-}
-
-*/
 // New version of torch_from_blob that uses strides
```
jatkinson1000, line 112:
```suggestion
// Exposes the given data as a Tensor without taking ownership of the original
// data
```

I would preserve the original docstring or something similar as "new" will no longer make sense in context without the old version.

jatkinson1000 (changes requested):
One small suggestion on code comments/function names.

Otherwise all good as discussed.
Suggest you rebase on main after #93 is merged in a moment to satisfy CI.

src/ctorch.cpp
jwallwork23: Ah yes good point, thanks. Fixed in 81795c1.

jatkinson1000 (approved):
Nice and simple, all looks good to me.
I'll squash and merge.
Congratulations on your first contribution @jwallwork23!

PR #103, opened 2024-03-28
 Add options for eval and gradient required #78 - rebase

### LINKED ISSUE #73: Loaded TorchScript missing no_grad context

jatkinson1000:
This is an updated version of #78 rebased onto main after the GPU changes by @jwallwork23

@ElliottKasoar's comment on the original PR:

Resolves #73

Adds flags in all(?) functions that operate on tensors (tensor creation, model loading, forward) to optionally disable autograd, which should improve performance for inference.

Also adds a similar flag to set evaluation mode for the loaded model.

- Evaluation mode
  - Contrary to my initial comments in #73, from testing evaluation mode does appear to be preserved, both between saving and loading TorchScript, and when applied to the loaded model.
  - In most cases evaluation mode is therefore likely to already be set, but I think it's useful to have the option to change it, particularly if FTorch may be extended to facilitate training (#22).

- NoGradMode
  - Enabling or disabling gradients is more complicated, as it defined via a context manager, which only appears to define the behaviour within its own scope, and so it seems necessary to enable/disable gradients before every code block that operates on tensors (similar to the Python equivalent `with torch.no_grad():`).

- InferenceMode
  - No changes are currently included, but it would be good to support [InferenceMode](https://pytorch.org/cppdocs/notes/inference_mode.html) too eventually, as it should provide further performance benefits over NoGradMode.
  - However, it has stricter requirements, and the mode was only added (as a beta) in PyTorch 1.9, so we would need to be careful if we want to support older versions.

- Model freezing
  - No changes are currently included, and less directly applicable to the main FTorch library, although there are still interactions e.g. freezing the model can allow InferenceMode to be enabled when loading the model.
  - Freezing is currently the "default" when tracing in [pt2ts.py](https://github.com/Cambridge-ICCS/FTorch/blob/main/utils/pt2ts.py), but not for scripting, despite potentially [improving performance](https://pytorch.org/tutorials/prototype/torchscript_freezing.html).
  - Freezing appears to (sometimes) introduce numerical errors when saving the reloading (differences ~10^-6), and can seem to lead to issues loading with Forpy too.

(For more general explanation of autograd/evaluation mode, see [autograd mechanics](https://pytorch.org/docs/1.9.0/notes/autograd.html)).

Note: I've also removed the old, commented out `torch_from_blob` function.

### DIFF
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -59,6 +59,16 @@ const auto get_device(torch_device_t device_type, int device_index)
   }
 }
 
+void set_is_training(torch_jit_script_module_t module, const bool is_training)
+{
+  auto model = static_cast<torch::jit::script::Module*>(module);
+  if (is_training) {
+    model->train();
+  } else {
+    model->eval();
+  }
+}
+
 torch_tensor_t torch_zeros(int ndim, const int64_t* shape, torch_data_t dtype,
                            torch_device_t device_type, int device_index = -1, const bool requires_grad)
 {
@@ -197,11 +207,7 @@ torch_jit_script_module_t torch_jit_load(const char* filename,
     delete module;
     exit(EXIT_FAILURE);
   }
-  if (is_training) {
-    module->train();
-  } else {
-    module->eval();
-  }
+  set_is_training(module, is_training);
 
   return module;
 }
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -60,8 +60,9 @@ const auto get_device(torch_device_t device_type, int device_index)
 }
 
 torch_tensor_t torch_zeros(int ndim, const int64_t* shape, torch_data_t dtype,
-                           torch_device_t device_type, int device_index = -1)
+                           torch_device_t device_type, int device_index = -1, const bool requires_grad)
 {
+  torch::AutoGradMode enable_grad(requires_grad);
   torch::Tensor* tensor = nullptr;
   try {
     // This doesn't throw if shape and dimensions are incompatible
@@ -82,8 +83,9 @@ torch_tensor_t torch_zeros(int ndim, const int64_t* shape, torch_data_t dtype,
 }
 
 torch_tensor_t torch_ones(int ndim, const int64_t* shape, torch_data_t dtype,
-                          torch_device_t device_type, int device_index = -1)
+                          torch_device_t device_type, int device_index = -1, const bool requires_grad)
 {
+  torch::AutoGradMode enable_grad(requires_grad);
   torch::Tensor* tensor = nullptr;
   try {
     // This doesn't throw if shape and dimensions are incompatible
@@ -104,8 +106,9 @@ torch_tensor_t torch_ones(int ndim, const int64_t* shape, torch_data_t dtype,
 }
 
 torch_tensor_t torch_empty(int ndim, const int64_t* shape, torch_data_t dtype,
-                           torch_device_t device_type, int device_index = -1)
+                           torch_device_t device_type, int device_index = -1, const bool requires_grad)
 {
+  torch::AutoGradMode enable_grad(requires_grad);
   torch::Tensor* tensor = nullptr;
   try {
     // This doesn't throw if shape and dimensions are incompatible
@@ -129,8 +132,10 @@ torch_tensor_t torch_empty(int ndim, const int64_t* shape, torch_data_t dtype,
 // data
 torch_tensor_t torch_from_blob(void* data, int ndim, const int64_t* shape,
                                const int64_t* strides, torch_data_t dtype,
-                               torch_device_t device_type, int device_index = -1)
+                               torch_device_t device_type, int device_index = -1,
+                               const bool requires_grad)
 {
+  torch::AutoGradMode enable_grad(requires_grad);
   torch::Tensor* tensor = nullptr;
 
   try {
@@ -174,8 +179,11 @@ void torch_tensor_delete(torch_tensor_t tensor)
 
 torch_jit_script_module_t torch_jit_load(const char* filename,
                                          const torch_device_t device_type = torch_kCPU,
-                                         const int device_index = -1)
+                                         const int device_index = -1,
+                                         const bool requires_grad,
+                                         const bool is_training)
 {
+  torch::AutoGradMode enable_grad(requires_grad);
   torch::jit::script::Module* module = nullptr;
   try {
     module = new torch::jit::script::Module;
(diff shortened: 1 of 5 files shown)


jatkinson1000:
Hi @ElliottKasoar after the work @jwallwork23 did there were some conflicts with your PR in #78.

I have done my best to rebase your work, but please could you take a look and see if everything seems in order to you?

@TomMelt You originally reviewed this PR and approved, but if you could take a quick glance and re-review it would be appreciated - since @jwallwork23 restructured the order of functions in the files and added additional arguments in the same place as @ElliottKasoar some of the merge conflicts got a little hairy so I may have missed the odd thing!

jatkinson1000:
Before we merge we need to:

- [ ] Add some docs for these new options
- [ ] Add an example showing the use of these new options? 
        Or perhaps not given we are not training yet and just use the defaults...?

src/ctorch.cpp
```diff
@@ -59,9 +59,20 @@ const auto get_device(torch_device_t device_type, int device_index)
...
+  if (is_training) {
+    model->train();
+  } else {
+    model->eval();
+  }
+}
+
 torch_tensor_t torch_zeros(int ndim, const int64_t* shape, torch_data_t dtype,
-                           torch_device_t device_type, int device_index = -1)
+                           torch_device_t device_type, int device_index = -1, const bool requires_grad)
```
jwallwork23, line 73: Should we provide a default value for `requires_grad` on the C++ side in the same way as for `device_index`? If not, should we reorder the arguments so that optional ones come last? (Similarly for other functions with this argument.)

I know this is already done on the Fortran side in this PR, but there may be users that want to use the C++ interface, too. Note that if `requires_grad` isn't made optional in this file then the draft example `n_c_and_cpp` will break because the API will change.

src/ctorch.cpp
```diff
@@ -174,8 +189,11 @@ void torch_tensor_delete(torch_tensor_t tensor)
 
 torch_jit_script_module_t torch_jit_load(const char* filename,
                                          const torch_device_t device_type = torch_kCPU,
-                                         const int device_index = -1)
+                                         const int device_index = -1,
+                                         const bool requires_grad,
+                                         const bool is_training)
```
jwallwork23, line 194: Similarly, would it make sense to use a default value for `is_training`?

jwallwork23 (changes requested):
Nice, seems like useful functionality.

Just a couple of comments that we need to address either by following my recommendations or by updating the `n_c_and_cpp` example to use the new API.

src/ftorch.fypp
```diff
@@ -160,26 +175,34 @@ contains
...
     integer(c_int), intent(in)     :: device_type  !! Device type the tensor will live on (`torch_kCPU` or `torch_kCUDA`)
     integer(c_int), optional, intent(in) :: device_index     !! device index to use for `torch_kCUDA` case
-    integer(c_int), intent(in)     :: layout(*)  !! Layout for strides for accessing data
+    logical(c_bool), optional, intent(in) :: requires_grad_opt  !! Whether gradients need to be computed for the created tensor
     type(torch_tensor)             :: tensor     !! Returned tensor
 
     integer(c_int)                 :: i          !! loop index
     integer(c_int64_t)             :: strides(ndims) !! Strides for accessing data
     integer(c_int)                 :: device_index_value  !! device index used
+    logical(c_bool)                :: requires_grad  !! Whether gradients need to be computed for the created tensor
```
ElliottKasoar, line 205:
```suggestion
    logical(c_bool)                :: requires_grad  !! Whether gradients need to be computed for the created tensor

    if (.not. present(requires_grad_opt)) then
      requires_grad = logical(.false., c_bool)
    else
      requires_grad = requires_grad_opt
    end if
```

src/ftorch.fypp
```diff
@@ -336,14 +386,15 @@ contains
   #:for RANK in RANKS
   !> Return a Torch tensor pointing to data_in array of rank ${RANK}$ containing data of type `${PREC}$`
   function torch_tensor_from_array_${PREC}$_${RANK}$d(data_in, layout, c_device_type, device_index) result(tensor)
```
ElliottKasoar, line 388:
```suggestion
  function torch_tensor_from_array_${PREC}$_${RANK}$d(data_in, layout, c_device_type, device_index, requires_grad_opt) result(tensor)
```

ElliottKasoar (changes requested):
Thanks! Looks great other than a couple of places where I think the code may have got lost in the rebase (unless intentional?)

src/ctorch.cpp
ElliottKasoar: I think at the time I may have been hesitant to introduce too many sources of defaults, but on reflection, yes I agree that adding the same defaults to the C++ side makes sense, especially to keep the pure C++ interface as an option.

src/ftorch.fypp
jatkinson1000: Nope - that's exactly why I wanted you to take a look!
Have committed the suggestions and will look at setting defaults later on.

src/ctorch.cpp
jatkinson1000: This should now be done in https://github.com/Cambridge-ICCS/FTorch/pull/103/commits/75b7aa0a246ede7e6c3cc0629939923941749543 if you are both happy with that?

src/ctorch.cpp
jatkinson1000: This should now be done in https://github.com/Cambridge-ICCS/FTorch/pull/103/commits/75b7aa0a246ede7e6c3cc0629939923941749543 if you are both happy with that?

src/ctorch.cpp
```diff
@@ -189,14 +211,16 @@ torch_jit_script_module_t torch_jit_load(const char* filename,
...
   }
+  set_is_training(module, is_training);
 
   return module;
 }
 
 void torch_jit_module_forward(const torch_jit_script_module_t module,
                               const torch_tensor_t *inputs, const int nin,
-			      torch_tensor_t output)
+			                        torch_tensor_t output, const bool requires_grad=false)
```
jwallwork23, line 221: This indentation looks like a typo?

jwallwork23 (approved):
One minor point on formatting but otherwise looks good to me!

src/ctorch.cpp
jatkinson1000: No idea what is going on as it all looks fine locally in my vim!?!
Even checked they were spaces, not tabs.
Oh well, removed and reinserted and now it looks OK on here I think.

jwallwork23: How odd! Thanks.

jatkinson1000:
Added a note to the FAQ about eval and no_grad settings.
A detailed example will perhaps wait until these are used as part of #111 since for now they are the sensible defaults for running inference.

Squashing and merging shortly.

Issue #115, opened 2024-04-15
Integration Tests


jatkinson1000:
For JOSS submission we require some form of testing.

A simple start would be integration tests.

I propose we adapt the SimpleNet example as follows:

- Use CMakeTest framework to build SimpleNet example
- Run for a variety of inputs and check that: 
  - The code runs without error 
  - outputs are correct
- This could use PyTest with system calls to the python to wrap around for numerical comparisons?

Longer term we can then work on PFUnit testing as per #76

PR #117, opened 2024-04-15
Fix build failures

### LINKED ISSUE #116: Fix build failures
I just pulled the most recent version of FTorch and tried to build it with the GNU compilers. However, I'm getting build errors due to places where `c_bool` is not available and where lines have become too long.

Will work on a quick fix.

jwallwork23:
Closes #116.

These fixes worked for me locally, we should be able to track such issues once #115 has been addressed.

I gather we haven't set a line length, but it seems there are a few that have got too long for the GNU compiler to be happy with. I restrict the offending lines to 80 characters here but we should probably choose a number of characters and stick with it. (There are many other lines longer than 80.)

### DIFF
--- a/src/ftorch.f90
+++ b/src/ftorch.f90
@@ -79,7 +79,7 @@ module ftorch
                                device_type, device_index, &
                                requires_grad) result(tensor_p) &
                                bind(c, name = 'torch_from_blob')
-      use, intrinsic :: iso_c_binding, only : c_int, c_int64_t, c_ptr, c_bool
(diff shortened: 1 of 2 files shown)


jatkinson1000 (approved):
Oh damn, not sure how that happened, but it's almost certainly on me!
Thanks for spotting and I am happy with this so will merge now.

jwallwork23:
> Oh damn, not sure how that happened, but it's almost certainly on me! Thanks for spotting and I am happy with this so will merge now.

I also should've tested your changes ran for me when I reviewed your PR.

Thanks!

Issue #37, opened 2023-09-20
Potential pytorch incompatibility


ElliottKasoar:
This is not an issue I've encountered, but having followed the FTorch build instructions, the version of libtorch/pytorch installed may mean that FTorch is incompatible with the model saved in the [examples](https://github.com/Cambridge-ICCS/fortran-pytorch-lib/blob/main/examples/1_ResNet18/README.md#running), as this pip installs torch in a (new) virtual environment.

This would only lead to errors if breaking changes were made to the TorchScript format between the versions, and in many cases the same pip-installed torch would be used anyway.

jatkinson1000:
Mmmmm this is a good point - I can see it being an issue if someone uses newer features of pytorch in their model, but then implements FTorch linked to an older version of LibTorch without those features.

I can't immediately think of an easy way around this other than recommending that users ensure that their LibTorch version is at least as new as the one their model was built with.

I don't think we need to change anything code-wise, but it would be interesting to know what the error raised would be so that we can recognise this in future should users come across it.

ElliottKasoar:
In terms of LibTorch, I don't think users can go too far back, as I get errors when trying to build FTorch of the form:

```
/home/ek/ICCS/fortran-pytorch-lib/fortran-pytorch-lib/ctorch.cpp:235:18: error: ‘synchronize’ is not a member of ‘torch::cuda’
  235 |     torch::cuda::synchronize();
```

when running `make` for versions <= 1.7.

For versions between 1.8 and 1.10, I can build FTorch successfully, but encounter errors when going through the example of the form:

```
 ./resnet_infer_fortran ../saved_resnet18_model_cpu.pt
[ERROR]: terminate called after throwing an instance of 'c10::Error'
  what():  isTuple()INTERNAL ASSERT FAILED at "../aten/src/ATen/core/ivalue_inl.h":1306, please report a bug to PyTorch. Expected Tuple but got String
Exception raised from toTuple at ../aten/src/ATen/core/ivalue_inl.h:1306 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::string) + 0x42 (0x7fd324cac302 in /home/ek/ICCS/libtorch-1.9/libtorch/lib/libc10.so)
[...]

PR #118, opened 2024-04-15
Introduce integration tests based on the examples

### LINKED ISSUE #115: Integration Tests

jwallwork23:
Closes #115.

Turns existing examples into integration tests using CTest.


jwallwork23:
That didn't turn out to be too difficult. A few notes:
1. Couldn't figure out how to get the Python scripts under CTest without making them executable - hopefully not an issue.
2.  The 2nd example depends on `numpy` but this wasn't in the requirements so I added it.
3.  I used a Python venv approach because that's what I'm familiar with and know how to easily determine the `${Torch_DIR}` env variable.
4.  Added a "proof of existence" assert at the end of the `pt2ts.py` scripts in the first two examples. Maybe this is something useful to have in the version in `src`, too?
5. Note that I had to truncate the outputs to 6 d.p. in the 2nd example to get the tests to pass.
6. Skipped the 3rd demo (multi-GPU) because we don't currently have access to GPUs in the CI.

dorchard:
Can you add some documentation about how to run locally?

jatkinson1000 (commented):
Hi @jwallwork23
Firstly, this is excellent, thank you!

I'll leave specifics for now as I have some more general structure questions/suggestions.

Currently we have to build and install the library, and then go and build each example to run a series of individual `ctest`s if I understand correctly?

I think it would be better if users could provide a flag when building the library, e.g. `-DCMAKE_BUILD_TESTS=ON` and then as part of the build process we make a local copy of the examples in the build directory and then run and execute the test process from there. This would add python as a dependency, but only if building with tests.
This would seem more natural to me - I expect for tests to be built and run as I build, before I install or do anything else.

At the moment it feels the tests are just a slightly more rigorous version of asking users to build and run the examples.

I feel I have seen similar in other cmake projects before where you provide a flag if you want to build with tests, and then run `ctest` before `cmake install`.

@TomMelt you have more experience here -- does this sound like a reasonable suggestion or am I barking up the wrong tree?

And as @dorchard says, definitely some docs/instructions.
I understand leaving full docs until things are settled, but some run instructions on the PR would have been useful - it took me a while (and a read of the workflow) to figure out how to make it work as someone who isn't super familiar with `ctest`!

jatkinson1000:
FTorch development meeting discussion:

- As part of CMake under a `test` flag create an empty directory in `src/` to hold tests.
- Copy code across from examples
- build and run tests in the new test directory if `test` flag is enabled
- Add the name of this directory to `.gitignore` in case anyone is using git as a developer etc.

jwallwork23:
@jatkinson1000 @dorchard Okay, I think I addressed the requests above. If FTorch is built with `-DCMAKE_BUILD_TESTS=TRUE` then the examples get build as integration tests in `src/test/examples`. The tests can then be run by going to the subdirectory of interest and running `ctest`, as described in [`src/test/README.md`](https://github.com/Cambridge-ICCS/FTorch/blob/115_integration_tests/src/test/README.md).

Note that I had a different attempt [`115_integration_tests_take2`](https://github.com/Cambridge-ICCS/FTorch/tree/115_integration_tests_take2) that moves the `CMakeLists.txt` to the root level (as well as moving the recommended build directory down one level), rather than copying over example files. I think either could work, just a matter of preference.

(Possibly) still to do:
* Add information to the user doc pages, as well as the READMEs?
* Add `ctest` setup for the MultiGPU example, even if it doesn't get run in the CI?

examples/1_SimpleNet/pt2ts.py
```diff
@@ -155,3 +158,7 @@ def load_torchscript(filename: Optional[str] = "saved_model.pt") -> torch.nn.Mod
             "Saved Torchscript model is not performing as expected.\n"
             "Consider using scripting if you used tracing, or investigate further."
         )
+
+    # Check that the model file is created
+    filepath = os.path.dirname(__file__) if len(sys.argv) == 1 else sys.argv[1]
+    assert os.path.exists(os.path.join(filepath, saved_ts_filename))
```
jatkinson1000, line 164: Would probably be good to incorporate this into the regular `pt2ts.py`, but as a separate issue?

src/README.md
```diff
@@ -17,6 +17,10 @@ We have created a minimum C wrapper interface of LibTorch (c_torch.h) that allow
 
 To build, follow the instructions in the [main repository README](/README.md).
 
+## Testing
+
+For information on testing, see the [test subdirectory](test/README.md).
+
```
jatkinson1000, line 23: I'm not sure I fully understand what `src/test/` exists, as whilst it gets populated I can't run `ctest` in there, but have to use the one in the build directory.

Is it because you have to copy the examples somewhere inside `src/` in order for the single `cmake` command to pick them up?

jatkinson1000 (changes requested):
Firstly, my sincere apologies it took me so long to get to this.
Secondly, this is fantastic and was really easy to read and follow, thank you.

I have a couple of user experience comments based on running this as well as any minor ones in the review.

At the first invocation of `cmake` build I get a warning that it cannot find FTorch (obviously) for the exercises. Is there any way to suppress this at all?

During `cmake --build .` I get a warning about duplicate linking:
```
ld: warning: ignoring duplicate libraries: '-lemutls_w', '-lgcc'
```
which would be nice to stop.

It was not immediately clear that I need to go into `./test/examples/examplename/` to invoke ctest. This also feels a bit clunky. I wonder if we could use a shell-script or something to allow users to run the tests from the build or test directory in a one-shot?
Feel free to push back if here if you would prefer to get this done and just write clear docs.

Finally, I think there is an element of 'works on my machine' as the regex didn't match on mine causing the tests to fail. Hopefully this can be easily fixed with formatted output...:
```
[jwa34@UNKNOWN 1_SimpleNet]$ ctest .
Test project /Users/jwa34/FTorch/src/build_test/test/examples/1_SimpleNet
    Start 1: simplenet
1/4 Test #1: simplenet ........................***Failed  Required regular expression not found. Regex=[0., 2., 4., 6., 8.
]  0.27 sec
    Start 2: pt2ts
2/4 Test #2: pt2ts ............................***Failed    0.21 sec
    Start 3: simplenet_infer_python
3/4 Test #3: simplenet_infer_python ...........***Failed  Required regular expression not found. Regex=[0., 2., 4., 6., 8.
]  0.17 sec
    Start 4: simplenet_infer_fortran
4/4 Test #4: simplenet_infer_fortran ..........***Failed  Required regular expression not found. Regex=[   0.00000000       2.00000000       4.00000000       6.00000000       8.00000000
]  0.73 sec

0% tests passed, 4 tests failed out of 4

Total Test time (real) =   1.38 sec

The following tests FAILED:
          1 - simplenet (Failed)
          2 - pt2ts (Failed)
          3 - simplenet_infer_python (Failed)
          4 - simplenet_infer_fortran (Failed)
Errors while running CTest
Output from these tests are in: /Users/jwa34/FTorch/src/build_test/test/examples/1_SimpleNet/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
```
and
```
[jwa34@UNKNOWN 2_ResNet18]$ ctest .
Test project /Users/jwa34/FTorch/src/build_test/test/examples/2_ResNet18
    Start 1: resnet18
1/3 Test #1: resnet18 .........................***Failed  Required regular expression not found. Regex=[0.004652
]  1.74 sec
    Start 2: pt2ts
2/3 Test #2: pt2ts ............................***Failed    0.15 sec
    Start 3: resnet_infer_fortran
3/3 Test #3: resnet_infer_fortran .............***Failed  Required regular expression not found. Regex=[0.884623
]  0.53 sec

0% tests passed, 3 tests failed out of 3

Total Test time (real) =   2.43 sec

The following tests FAILED:
          1 - resnet18 (Failed)
          2 - pt2ts (Failed)
          3 - resnet_infer_fortran (Failed)
Errors while running CTest
Output from these tests are in: /Users/jwa34/FTorch/src/build_test/test/examples/2_ResNet18/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
```

src/README.md
jwallwork23: > I'm not sure I fully understand what `src/test/` exists, as whilst it gets populated I can't run `ctest` in there, but have to use the one in the build directory.

Yeah `src/test/` will just contain the copied files. I guess we could make these copies intermediate temporaries and have CMake delete them once the ones in `src/build/test` are built? Would that reduce or increase user confusion?

> Is it because you have to copy the examples somewhere inside `src/` in order for the single `cmake` command to pick them up?

That's right.

jwallwork23:
> At the first invocation of `cmake` build I get a warning that it cannot find FTorch (obviously) for the exercises. Is there any way to suppress this at all?

Please could you share the warning message? I don't think I see such a warning.

> During `cmake --build .` I get a warning about duplicate linking:
> 
> ```
> ld: warning: ignoring duplicate libraries: '-lemutls_w', '-lgcc'
> ```
> 
> which would be nice to stop.

Hm, I haven't seen that one. Will look into it.

examples/1_SimpleNet/pt2ts.py
jwallwork23: Addressed in bd2bc14 as it was easy enough to do.

jwallwork23:
> It was not immediately clear that I need to go into `./test/examples/examplename/` to invoke ctest. This also feels a bit clunky. I wonder if we could use a shell-script or something to allow users to run the tests from the build or test directory in a one-shot? Feel free to push back if here if you would prefer to get this done and just write clear docs.

Provided a shell script in fe6bbd2, added some text, and hooked it up in the CI test suite. Let me know if it helps.

jwallwork23:
> Finally, I think there is an element of 'works on my machine' as the regex didn't match on mine causing the tests to fail. Hopefully this can be easily fixed with formatted output...:

@jatkinson1000 please could you rerun these with verbose output and send them over (e.g., in Slack)? So that I can better interpret why the regexes are failing. (For verbose mode, use `ctest -V`.)

src/README.md
jatkinson1000: Hhhmmm, my feeling is that making them temporary and having CMake remove might be nice, as keeps things clean and avoids this confusion.

If people build with tests enabled one assumes they will only ever go to 'wherever the docs instruct them to go to' and run the tests. If they wanted to explore more they should build the examples themselves.

So I think make CMake clean up after the build is a great idea.

run_integration_tests.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/bash
```
jatkinson1000, line 1:
```suggestion
#!/bin/bash
```

I think you want this for compatibility across systems (I had to change it to this go get it to work for me and have seen similar issues in the past).

run_integration_tests.sh
```diff
@@ -0,0 +1,24 @@
...
+# your configuration.
+#
+# See `src/test/README.md` for more details on integration testing.
+# ---
+
+set -eux
+
+CTEST_ARGS=$@
+EXAMPLES="1_SimpleNet 2_ResNet18"
+BUILD_DIR=src/build
```
jatkinson1000, line 16: As a note:
This initially failed for me because my build directory was `build_joe_tests/` rather than `build/`.

But I suppose if users are following our instructions verbatim then their build directory will be `./src/build/` and use of this script is optional, so it's probably OK.

src/test/README.md
jatkinson1000, line 1: As discussed, guidance that you need a venv/environment with all the necessary modules for the python side would be useful.

Though also as discussed, this folder/guidance may now get moved elsewhere to keep things clean.

README.md
```diff
@@ -129,6 +129,7 @@ To build and install the library:
     | [`CMAKE_PREFIX_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html)        | `</path/to/libTorch/>`       | Location of Torch installation<sup>2</sup>                    |
     | [`CMAKE_INSTALL_PREFIX`](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html)  | `</path/to/install/lib/at/>` | Location at which the library files should be installed. By default this is `/usr/local` |
     | [`CMAKE_BUILD_TYPE`](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html)          | `Release` / `Debug`          | Specifies build type. The default is `Debug`, use `Release` for production code|
+    | `CMAKE_BUILD_TESTS                                                                                | `TRUE` / `FALSE`             | Specifies whether to compile FTorch's test suite as part of the build. |
```
jatkinson1000, line 132:
```suggestion
    | `CMAKE_BUILD_TESTS`                                                                                | `TRUE` / `FALSE`             | Specifies whether to compile FTorch's test suite as part of the build. |
```
missing backtick.

jatkinson1000 (commented):
Couple more points as discussed offline.
Plus couple of suggestions I spotted whilst running/reading.

run_integration_tests.sh
jwallwork23: Yeah I also added a comment about editing `BUILD_DIR` above.

src/test/README.md
jwallwork23: See e34506b1d40e7e877979a99f0d2489c48dedea76.

run_integration_tests.sh
jwallwork23: Addressed in 143b1342a11bff1ffd6568b671e54c8a1b7b646f.

run_integration_tests.sh
jatkinson1000: Yeah, apologies -_-

src/README.md
jwallwork23: Hm, on further investigation, this isn't going to work.

The Python executables are run from `${PROJECT_SOURCE_DIR}`, whereas the Fortran executables are run from `${PROJECT_BUILD_DIR}`. We could copy over the Python executables, but then (in some of the cases) we'd hit the issue that the data they rely on are no longer accessible because the data directory is in the (deleted) source directory.

We could tweak the data directory paths so they are read from the original example directories (outside of `src`) but (a) CMake doesn't know about those locations and (b) it's becoming sufficiently hacky that we might be better going with the the alternative approach in [`115_integration_tests_take2`](https://github.com/Cambridge-ICCS/FTorch/tree/115_integration_tests_take2) (i.e., move the main `CMakeLists.txt` to the root FTorch directory), which doesn't involve copying.

run_integration_tests.sh
jwallwork23: No worries

src/README.md
jwallwork23: There are downsides to that approach, as already discussed. (Namely, users might want to run the examples themselves and find that the test suite has already created various files when running them itself.)

A third way might be to have a permanent `src/test/examples` (or similar) location, containing symlinks to the original examples. However, this doesn't avoid the potential user confusion issue and I'm not sure whether having symlinks in your repo is good practice.

README.md
```diff
@@ -200,7 +200,7 @@ instructions to modify, build, and run as necessary.
 
 ## Tests
 
-For information on testing, see the [test subdirectory](src/examples/README.md).
+For information on testing, see the [test subdirectory](src/test/README.md).
```
jatkinson1000, line 203: Could we also add a link to the online page that gets built?

src/README.md
jatkinson1000: Those are all sensible points, thank you for thinking about this!

Yes, I'm inclined to agree that it is looking hacky for other approaches.
I think with tests being optional, and providing very clear instructions on how to build and run them, we should be OK.

This is at least sufficient for a starting point with integration tests and perhaps the JOSS reviewers might come up with some wonderful solution that we hadn't yet considered 🤞
Or one of us will have a brainwave in the shower and we can refactor at a later point.

jatkinson1000 (changes requested):
OK, This is amazing stuff and I think 99% there so I'm keen to get it merged soon.

My last request would be for a detailed page for the online documentation on how to build and run the tests, and details of "what/how" gets tested.
We can then point to this from the main README.

To do this you want to create `pages/testing.md` and copy the style from an adjacent file.
This should then create a webpage "https://cambridge-iccs.github.io/FTorch/page/tests.html" which would be available once pushed to `main` There is a workflow to check it will build from the development branch, and you can see what it looks like locally by running FORD and viewing in a browser.

src/README.md
jwallwork23: Okay cool. We can discuss at a future FTorch dev meeting - maybe someone will have an idea.

README.md
jwallwork23: Sure. Addressed in f476b17.

jwallwork23:
Okay @jatkinson1000, I added a `pages/testing.md` page as requested and linked it in elsewhere.

There's a comment above on two new warnings that haven't been addressed. Perhaps we could raise these as issues? Otherwise, ready for re-review.

pages/testing.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Building the integration tests
+
+To enable FTorch's integration tests, ensure that the `CMAKE_BUILD_TESTS` option
+is set to `TRUE` for the build.
+
+### Running the integration tests
+
+Once the build is complete, activate the Python virtual environment you created
+for FTorch and run an individual test by going to the desired subdirectory of
```
jatkinson1000, line 19: It is possible that the user built against a libtorch binary rather than a pip installed version, so if they did this they just need to make sure they are in a virtual environment I guess (and torch will get installed).

This does raise again the points in #37, but I don't think we should worry about that here.

pages/testing.md
```diff
@@ -0,0 +1,31 @@
+title: FTorch test suite
+
+[TOC]
+
+## Testing
+
+FTorch's test suite is currently comprised of integration tests based on a
+subset of the [examples](examples.html). These tests are built and run and their
+outputs are analysed to check they contain expected regular expressions.
```
jatkinson1000, line 9: Brilliant, thanks!

pages/testing.md
```diff
@@ -0,0 +1,31 @@
...
+and which failed for your build. Note that some of the examples have additional
+dependencies, which may need installing into your virtual environment.
+
+As an alternative to the above, simply run the
+[helper script](https://github.com/Cambridge-ICCS/FTorch/blob/main/run_integration_tests.sh)
+in the root FTorch directory:
+```
+./run_integration_tests.sh
+```
+This will automatically install any additional dependencies for the examples.
```
jatkinson1000, line 31: Consider moving this above as the 'recommended' way to run tests, with running each one individually being the 'alternative'.

jatkinson1000 (approved):
Yes, I think this is good now.

Small comments that are up to you to consider for the online docs, but I am approving.

Please squash and merge once once you have read and done/not done any further changes. 😄

pages/testing.md
jwallwork23: Good point. See 3d03e8866b75aa08ad6a4e26be540c697c06e3c6.

pages/testing.md
jwallwork23: Addressed in 73095d7.

jwallwork23:
Thanks very much for your review @jatkinson1000! Will merge now.

PR #123, opened 2024-04-30
Update Windows help

jatkinson1000:
Whilst working through helping a Windows user in #124 I found that the advice we provide could be clearer:

- Clarify that you need the C++ tools as part of the VS install 
- Link to the Base and HPC toolkits separately - currently we name both, but only link to the base
- Remove hardcoded filepath to be more generic


TomMelt (approved):
Looks good. Thank you

Issue #124, opened 2024-04-30
Windows Installation Issue


vbalmer:
I am rather new to the Fortran environment in Windows and have encountered some trouble during installation. Even though I have followed the instructions provided on the Website for Windows (https://cambridge-iccs.github.io/FTorch/page/troubleshooting.html), the following error consistently shows up when running the "cmake --build ." command: 
 
1.	“fatal error LNK1181: cannot open input file ‘mkl_blacs_ilp64.lib’” (see attached screenshot) . Apparently I am not the only person with this issue (https://community.intel.com/t5/Intel-oneAPI-Math-Kernel-Library/Intel-oneAPI-Base-Toolkit-v-2024-1-Error-mkl-blacs-ilp64-lib/m-p/1587668#M3636), but also changing the environment variables didn’t do the job for me. For your understanding I have attached the full command lines that I have used for installation as txt to this E-Mail. I have removed this error with a very messy cleanup by copying and renaming that specific dll file in the intel compiler to test the further workflow, where I received more different errors (see 2.).
2.	When I try to run the Simplenet example the following errors show up: unresolved external symbol FTORCH_mp_TORCH_TENSOR_FROM_ARRAY_REAL32_1D referenced in function MAIN__ 
(and additional functions see second image attached). Is this error related to the first? Where is this variable defined?

The following changes to the installation have also been tried out but were not successful: 
1. running
cmake .. -G "NMake Makefiles" -DCMAKE_BUILD_TYPE=Release-DCMAKE_PREFIX_PATH="C:\Users\vbalmer\Anaconda3\envs\Py-F-02\Lib\site-packages\torch" -DCMAKE_INSTALL_PREFIX="C:\Users\vbalmer\AppData\Local\FTorch"
--> doesn't successfully run cmake command
2. including the old Fortran compiler ifort instead of ifx --> leads to same error as outlined above

Thank you very much for your help in advance!

![Error_Screenshot (005)](https://github.com/Cambridge-ICCS/FTorch/assets/116555051/6175eac3-ea5a-47e5-a4e6-33ae1cf9cfe8)
[BuildInstructions_Clean.txt](https://github.com/Cambridge-ICCS/FTorch/files/15168045/BuildInstructions_Clean.txt)
![Error_SimpleNet_Screenshot](https://github.com/Cambridge-ICCS/FTorch/assets/116555051/61f06c9e-2cdf-43b2-bac7-28aa86498dc6)

[...]

PR #134, opened 2024-06-05
Require ftorch when coupling

### LINKED ISSUE #133: Make FTorch REQUIRED in the examples
It came up in #124 that some linking could pass undetected for the examples as CMake does not currently '`REQUIRE`' FTorch.

This could lead to linking issues later on at link or runtime that should be detected earlier when building.

We should update the examples (which are then pulled into the tests) and the online coupling guidance.

jatkinson1000:
Closes #133

It came up in #124 that some linking could pass undetected for the examples as CMake does not currently '`REQUIRE`' FTorch.

This PR changes that for the examples (which are then pulled into the tests) and in the online coupling guidance.

### DIFF
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -41,11 +41,13 @@ jobs:
         run: |
           . ftorch/bin/activate
           VN=$(python -c "import sys; print('.'.join(sys.version.split('.')[:2]))")
-          export Torch_DIR=${VIRTUAL_ENV}/lib/python${VN}/site-packages
+          export Torch_DIR=${VIRTUAL_ENV}/lib/python${VN}/site-packages/torch
           export BUILD_DIR=$(pwd)/src/build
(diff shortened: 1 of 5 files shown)


jwallwork23:
Hm so this implies we need to build the tests separately from FTorch? Doable but means we have two build steps.

Either that or perhaps we can use `POST_BUILD`. I wonder if this is relevant? https://stackoverflow.com/questions/50913377/cmake-add-custom-command-post-build-depends-option-is-ignored

jatkinson1000:
My feelings are that this is not a priority, so can perhaps be put on a backburner for now - I had hoped it would be a 1-line change, but that does not seem the case. 😅

That said, don't let it stop you looking if you are interested!

I'm softening on it being a necessity to build tests as a one-shot with CMake - maybe having a separate "build tests" step wouldn't be an issue as one still has to follow a separate "run tests" step? I still think separating tests from examples (even if they use the same code) is sensible, however.

PR #136, opened 2024-06-12
Handle multiple outputs from models.

### LINKED ISSUE #3: Handling multiple outputs from a model
At present the wrapper codes can only accept a single output tensor from a TorchScript model.

However, it is perfectly possible to return multiple output tensors from a PyTorch Model.
In this instance the python code returns them as a Tuple.

Is it possible to return multiple output tensors when using the C++ API (and by extension our wrapper scripts)?

The 'simple' answer is to add a concatenation layer onto the end of the pytorch model to return a single tensor and then unpack later. However, this is hacky, requires model alteration to use our code, and restricts us to returning only a single type(?).

The forward method in the API returns an `IValue` which is a union over various types: https://pytorch.org/cppdocs/api/structc10_1_1_i_value.html

There has been some discussion on this topic:
- There is a [Stack Overflow](https://stackoverflow.com/questions/61988134/pytorch-c-frontend-returning-multiple-tensors-in-forward)
- This has previously been raised as an [official bug](https://github.com/pytorch/pytorch/issues/36429), but in a questionable manner
- There is a [PyTorch discussion](https://discuss.pytorch.org/t/strange-issue-returning-torch-tensors-in-libtorch-c/40242/2) with a suggested, but unverified, solution.

It looks like this is probably possible, but may need a bit of rummaging around with C/C++ types and moving pointers.

jatkinson1000:
Currently work in progress.

Requested by user in #135 
Closes #3 

Definitely possible, but my C++-fu is not as strong as Fortran so currently stuck in pointer hell.

- [x] Update C++ to process the option of multiple outputs
    - [x] Inspect result to see if it is a tensor or tuple and handle accordingly
- [x] Update Fortran to handle this
    - [x] Make outputs a vector like inputs
    - [x] Require n_out as argument to forward

Other thoughts:

- This will be a breaking API change for current users 😞 
- [x] Should we make this an interface so users can pass a single tensor or an array of tensors?
- [x] Can we somehow detect n_in? I have a feeling we looked before and the answer was no...  
    - The answer is, in fact, yes - see https://github.com/Cambridge-ICCS/FTorch/pull/136/commits/4ab920a5d9e0260a30b3c844aeeac06223910fbf


jwallwork23:
Okay @jatkinson1000 @TomMelt @ma595 this is now working! Ready for review.

examples/3_MultiGPU/CMakeLists.txt
```diff
@@ -16,6 +16,6 @@ find_package(MPI REQUIRED)
 message(STATUS "Building with Fortran PyTorch coupling")
 
 # Fortran example
-add_executable(simplenet_infer_fortran simplenet_infer_fortran.f90)
-target_link_libraries(simplenet_infer_fortran PRIVATE FTorch::ftorch)
-target_link_libraries(simplenet_infer_fortran PRIVATE MPI::MPI_Fortran)
+add_executable(simplenet_infer_fortran_gpu simplenet_infer_fortran.f90)
+target_link_libraries(simplenet_infer_fortran_gpu PRIVATE FTorch::ftorch)
+target_link_libraries(simplenet_infer_fortran_gpu PRIVATE MPI::MPI_Fortran)
```
jatkinson1000, line 21: I feel like ideally this change would be made in a different PR as it is not related to the subject of this PR (as far as I can tell, is it to do with a name clash when building tests?).
Easiest to leave it in here now.

examples/4_MultiIO/README.md
jatkinson1000, line 1: I'm sure you are aware, but just flagging that this will need completing.

src/ctorch.cpp
```diff
@@ -242,9 +243,24 @@ void torch_jit_module_forward(const torch_jit_script_module_t module,
...
+      // Single output models will return a tensor directly.
+      std::move(*out[0]) = model_out.toTensor();
+    }
+    else if (model_out.isTuple()) {
+      // Multiple output models will return a tuple => cast to tensors.
+      // See https://github.com/pytorch/pytorch/issues/15523
+      for (int i=0; i<nout; ++i) {
+        std::move(*out[i]) = model_out.toTuple()->elements()[i].toTensor();
+      }
+    }
```
jatkinson1000, line 258: On data transfers, I am not a C++ expert, but I think the only dangerous step here is whether `.toTensor()` or `.toTuple()` create data copies.
That said, I think this mirrors the previous approach.
Let me know your thoughts.

utils/pt2ts.py
```diff
@@ -144,24 +144,29 @@ def load_torchscript(filename: Optional[str] = "saved_model.pt") -> torch.nn.Mod
...
+        trained_model_testing_outputs = (trained_model_testing_outputs,)
+    for ts_output, output in zip(ts_model_outputs, trained_model_testing_outputs):
+        if torch.all(ts_output.eq(output)):
+            print("Saved TorchScript model working as expected in a basic test.")
+            print("Users should perform further validation as appropriate.")
+        else:
+            raise RuntimeError(
+                "Saved Torchscript model is not performing as expected.\n"
+                "Consider using scripting if you used tracing, or investigate further."
+            )
```
jatkinson1000, line 169: Thank you for including this change!!

jatkinson1000 (commented):
Thanks @jwallwork23 

This looks sensible to me, but please do double check the data copy aspect as you mentioned.
I can confirm things run as expected.

Remaining things would be to add documentation, and then make a final decision on what we want to to do with the API (probably in the opposite order!!) which we can discuss. Let me know if you want me to lend a hand with documentation.

I'm also noting that I can't approve this PR as I am a co-author, so will also need @TomMelt to take a look.

src/ctorch.cpp
TomMelt: Hi @jatkinson1000 , according to the PyTorch API docs I don't think that the `toX()` methods create a copy.

> IValues are used as inputs to and outputs from the TorchScript interpreter. To retrieve the value contained within an [IValue](https://pytorch.org/cppdocs/api/structc10_1_1_i_value.html#structc10_1_1_i_value), use the .toX() methods, where X is the type you are trying to get. Note that neither the .toX() methods nor the templated .to<T> functions do any kind of casting, they only unwrap the contained value.

see https://pytorch.org/cppdocs/api/structc10_1_1_i_value.html#struct-documentation for more details.

jatkinson1000: Perfect, in that case I think we are all good!

examples/4_MultiIO/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+  # 4. Check the model can be loaded from file and run in Fortran and that its outputs
+  #    meet expectations
+  add_test(NAME simplenet_infer_fortran
+    COMMAND simplenet_infer_fortran_multiio
+    ${PROJECT_BINARY_DIR}/saved_multiio_model_cpu.pt # Command line argument: model file
+    WORKING_DIRECTORY ${PROJECT_BINARY_DIR})
+  set_tests_properties(simplenet_infer_fortran PROPERTIES PASS_REGULAR_EXPRESSION
+    "   0.00000000       2.00000000       4.00000000       6.00000000")
+  set_tests_properties(simplenet_infer_fortran PROPERTIES PASS_REGULAR_EXPRESSION
+    "   0.00000000      -3.00000000      -6.00000000      -9.00000000")
```
TomMelt, line 59: these tests will most likely fail with the intel compiler and possibly different version of `gfortran`. The default output is quite different for both. We should think of another way to do this.

We could add something like the resnet test?

https://github.com/Cambridge-ICCS/FTorch/blob/b40b03d4c39f607de754d34991a92703ccc36eb3/examples/2_ResNet18/resnet_infer_fortran.f90#L186C15-L214

examples/4_MultiIO/pt2ts.py
```diff
@@ -0,0 +1,174 @@
...
+    # Switch off specific layers/parts of the model that behave
+    # differently during training and inference.
+    # This may have been done by the user already, so just make sure here.
+    trained_model.eval()
+
+    # =====================================================
+    # Prepare dummy input and check model runs
+    # =====================================================
+
+    # FPTLIB-TODO
```
TomMelt, line 96: what do these comments refer to (`FPTLIB-TODO`). Is it for documentation?

examples/4_MultiIO/pt2ts.py
```diff
@@ -0,0 +1,174 @@
...
+    dummy_input : torch.Tensor
+        appropriate size Tensor to act as input to model
+    filename : str
+        name of file to save to
+    """
+    print("Saving model using tracing...", end="")
+    traced_model = torch.jit.trace(model, dummy_input)
+    frozen_model = torch.jit.freeze(traced_model)
+    ## print(frozen_model.graph)
+    ## print(frozen_model.code)
```
TomMelt, line 56: can we remove these comments

examples/4_MultiIO/pt2ts.py
```diff
@@ -0,0 +1,174 @@
...
+    Parameters
+    ----------
+    model : torch.NN.Module
+        a PyTorch model
+    filename : str
+        name of file to save to
+    """
+    print("Saving model using scripting...", end="")
+    scripted_model = torch.jit.script(model)
+    # print(scripted_model.code)
```
TomMelt, line 30: same here

examples/4_MultiIO/simplenet_infer_fortran.f90
```diff
@@ -0,0 +1,62 @@
...
+   out_tensors(1) = torch_tensor_from_array(out_data1, tensor_layout, torch_kCPU)
+   out_tensors(2) = torch_tensor_from_array(out_data2, tensor_layout, torch_kCPU)
+
+   ! Load ML model
+   model = torch_module_load(args(1))
+
+   ! Infer
+   call torch_module_forward(model, in_tensors, out_tensors)
+   write (*,*) out_data1
+   write (*,*) out_data2
```
TomMelt, line 53: if you want to keep the regex check for CMake test we can force the output formatting here. See e.g., https://web.stanford.edu/class/me200c/tutorial_77/17_format.html

Perhaps this?
```suggestion
   write (*,'(4(f8.3))') out_data1
   write (*,'(4(f8.3))') out_data2
```

TomMelt: This will (should) give the same output and be compiler agnostic.

TomMelt (changes requested):
Looks good. Just a couple of things to discuss

examples/4_MultiIO/pt2ts.py
jatkinson1000: They signify to the user where they need to edit the guidance script `pt2ts.py` to export their particular net to TorchScript. I don't think they need changing here.

examples/4_MultiIO/pt2ts.py
jatkinson1000: I think they are a pull over from the `pt2ts.py` script where they are an option for the user to check that they are running things correctly. I would erro on leaving them in for consistency with the base `pt2ts.py`, but open to a discussion.

examples/4_MultiIO/pt2ts.py
jatkinson1000: As above.

examples/4_MultiIO/simplenet_infer_fortran.f90
jatkinson1000: @jwallwork23 This falls under testing so do you want to cover this?
I would also consider opening a separate issue and PR to enact this in the other examples as it is not directly related to this PR.

examples/4_MultiIO/CMakeLists.txt
jatkinson1000: Setting the output to `PASS` or `FAIL` and reading that would probably be a sensible approach, I agree.
I would encourage doing this for the other exercises in a separate PR, however.

jwallwork23: I agree, that would be a better approach indeed. I opened a new issue for this: #142.

examples/4_MultiIO/README.md
jwallwork23: Apologies - done now in d3abc449bd3dafbcee18fd8ac402a67341551008.

jwallwork23:
Updated docs - ready for re-review.

examples/4_MultiIO/CMakeLists.txt
TomMelt: ok we can fix this at a later stage :ok_hand:

examples/4_MultiIO/pt2ts.py
TomMelt: My general opinion is don't leave commented code in repos but in this case it's not a deal breaker.

I guess the ideal scenario is it's either worth keeping (i.e., refactored with a debug flag for example) or it's junk. The problem is, commented code doesn't get tested or updated. So it can get stale quickly :grimacing:

TomMelt (approved):
Thanks @jatkinson1000 and @jwallwork23 for this PR. It is a great addition to the code.

jatkinson1000:
@TomMelt It seems like this was merged somewhat prematurely.
As a result I am having to update the README over in #143 to cover some of the changes made here which is not ideal.

PR #139, opened 2024-06-21
Overload elementary operations

### LINKED ISSUE #138: Overload elementary operations for `torch_tensor`
Related to #111.

Recently we've discussed implementing automatic differentiation (AD) in FTorch using the underlying autograd AD tool built into Torch. The first step to achieve this would be to overload elementary operations for the Fortran `torch_tensor` class: assignment, addition, multiplication of tensors, multiplication of tensor by scalar, etc.

jwallwork23:
Closes #138.

Note that about 1,000 lines of changes are automatically generated from the `ftorch.fypp` changes. So this PR isn't as enormous as it looks.


jatkinson1000:
This seems sensible, and impressive 😄
What is the argument for making `torch_tensor_from_array()` no longer return a value but instead operate internally?
We can chat over lunch.

jwallwork23:
> This seems sensible, and impressive 😄 What is the argument for making `torch_tensor_from_array()` no longer return a value but instead operate internally? We can chat over lunch.

Thanks!

If you overload the assignment operator then `tensor = torch_tensor_from_array(...)` will call `torch_tensor_from_array` with the provided arguments and then call `torch_tensor_assign`, copying that into `tensor`. If this and other constructor-type procedures go from being functions to subroutines then we avoid unnecessary data copies.

examples/5_Autograd/autograd.f90
```diff
@@ -0,0 +1,50 @@
...
+  ! FIXME: requires_grad=.true.
+  call torch_tensor_from_array(a, in_data1, tensor_layout, torch_kCPU)
+  call torch_tensor_from_array(b, in_data2, tensor_layout, torch_kCPU)
+  call torch_tensor_from_array(Q, out_data, tensor_layout, torch_kCPU)
+
+  ! Check arithmetic operations work for torch_tensors
+  write (*,*) "a = ", in_data1(:)
+  write (*,*) "b = ", in_data2(:)
+  Q = 3 * a ** 3 - b ** 2
+  write (*,*) "Q = 3 * a ** 3 - b ** 2 =", out_data(:)
```
jwallwork23, line 36: In the first example, the Fortran array underlying the output tensor gets modified. However, it doesn't here. Is there something obvious I missed?

jwallwork23: The overloaded `=` creates a copy so we shouldn't expect `Q` to point to `out_data` any more. To solve this, we first need to implement `torch_tensor_to_array` (#151).

jwallwork23:
I created a new branch [`autograd_toarray`](https://github.com/Cambridge-ICCS/FTorch/tree/autograd_toarray) off this one and merged in changes from #161. Can confirm the tests pass!

examples/6_Autograd/autograd.f90
```diff
@@ -26,45 +27,78 @@ program example
...
 
-  ! initialize in_data with some fake data
-  do j = 1, m
-    do i = 1, n
-      in_data(i,j) = ((i-1)*m + j) * 1.0_wp
-    end do
-  end do
+  ! Initialise input arrays as in Python example
+  in_data1(:,1) = [2.0, 3.0]
+  in_data2(:,1) = [6.0, 4.0]
```
jwallwork23, line 34: @TomMelt I decided to go back to the original setup, for consistency with the Python version of the example. I made the arrays shape (2, 1) rather than just (2) so we still test the `get_rank` and `get_shape` utils you added but ideally these should be moved into separate unit tests with #4. Once we do that, the arrays here can be switched back to shape (2).

examples/6_Autograd/autograd.f90
```diff
@@ -26,45 +27,78 @@ program example
...
 
   ! Construct a Torch Tensor from a Fortran array
-  call torch_tensor_from_array(tensor, in_data, tensor_layout, torch_kCPU)
+  ! TODO: Implement requires_grad=.true.
```
jwallwork23, line 37: This TODO won't be addressed in this PR.

examples/6_Autograd/autograd.f90
```diff
@@ -26,45 +27,78 @@ program example
...
+  expected(:,1) = [6.0, 4.0]
+  test_pass = assert_real_array_2d(in_data2, expected, test_name="torch_tensor_to_array", rtol=1e-5)
+  if (.not. test_pass) then
+    call clean_up()
+    print *, "Error :: in_data2 was changed during arithmetic operations"
+    stop 999
+  end if
+
+  ! Back-propagation
+  ! TODO: Requires API extension
```
jwallwork23, line 88: Neither will this one. (See #158.)

jwallwork23:
Got round to updating this branch. The simplest thing to do was to merge `main` into it, so it should be merged with squash.

jwallwork23:
As discussed in today's meeting, we should check what happens if an operator that hasn't been implemented is used. I switched from `Q = 3 * a ** 3 - b ** 2` to `Q = 3 * a ** b - b ** 2`, i.e., take a tensor to the power of another tensor, rather than to the power of a scalar. I get the following compile error, which seems clear enough:
```
Scanning dependencies of target autograd
[ 91%] Building Fortran object test/examples/6_Autograd/CMakeFiles/autograd.dir/autograd.f90.o
/home/joe/software/FTorch/src/test/examples/6_Autograd/autograd.f90:54:14:

   54 |   Q = 3 * a ** b - b ** 2
      |              1
Error: Unexpected derived-type entities in binary intrinsic numeric operator ‘**’ at (1)

```

src/ftorch.fypp
```diff
@@ -115,6 +115,36 @@ module ftorch
...
+    module procedure torch_tensor_multiply
+    #:for PREC in PRECISIONS
+    module procedure torch_tensor_premultiply_${PREC}$
+    module procedure torch_tensor_postmultiply_${PREC}$
+    #:endfor
+  end interface
+
+  interface operator (/)
+    module procedure torch_tensor_divide
+  end interface
```
jwallwork23, line 140: Might as well pre/post-divide by scalar, too.

examples/6_Autograd/autograd.f90
```diff
@@ -4,8 +4,9 @@ program example
   use, intrinsic :: iso_fortran_env, only : sp => real32
 
   ! Import our library for interfacing with PyTorch's Autograd module
-  use ftorch, only : torch_tensor, torch_kCPU, &
-                     torch_tensor_from_array, torch_tensor_to_array, torch_tensor_delete
+  use ftorch, only: assignment(=), operator(+), operator(-), operator(*), &
+    operator(**), torch_kCPU, torch_tensor, torch_tensor_delete, &
```
jwallwork23, line 8: Needs to be documented in developer docs.

jwallwork23: Could potentially link to C++ API docs if they're sufficiently stable.

src/ftorch.fypp
jwallwork23: Post-divide implemented in b804d54335989820846326fce92ac633fd1c2b03. Left pre-division (i.e., scalar divided by tensor) for now as I don't think it's something people use often and I need to check what that actually does on the Torch side.

jwallwork23: In f0ed978de73d86b4167daa8af2ddd9ef3563e0b2 I reworked the autograd test so that it tests multiply and divide without changing expected values.

jwallwork23:
Merged in the Windows fix from #207.

examples/6_Autograd/autograd.f90
jwallwork23: Unfortunately the individual functions in the API docs include version number so I don't think we can expect them to point to the latest version? (e.g., `v4` in https://pytorch.org/cppdocs/api/classat_1_1_tensor.html#_CPPv4NR2at6TensoraSERK6Tensor?)

However, I added some links to the Torch Tensor C++ API docs page https://pytorch.org/cppdocs/api/classat_1_1_tensor.html#exhale-class-classat-1-1-tensor in
https://github.com/Cambridge-ICCS/FTorch/pull/139/commits/c0a3fe49c73f130cf51ab6438add5469c93027b2

examples/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
 if(CMAKE_BUILD_TESTS)
   add_subdirectory(1_SimpleNet)
   add_subdirectory(2_ResNet18)
-  # add_subdirectory(3_MultiGPU)
+  if(ENABLE_CUDA)
+    add_subdirectory(3_MultiGPU)
+  endif()
```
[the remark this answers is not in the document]
TomMelt: I guess this is not really related to this PR though. Maybe we can create a separate issue for this?

TomMelt (approved):
Thnaks @jwallwork23 , this looks great. Just a single comment really and I think it should possibly be a separate issue anyway. lmk what you think. I am happy to merge as is.

examples/CMakeLists.txt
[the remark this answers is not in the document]
jwallwork23: Thanks @TomMelt, good point. I reverted this in 8dfa92c and opened #208.

examples/6_Autograd/autograd.f90
```diff
@@ -27,45 +29,78 @@ program example
...
+  contains
+
+    ! Subroutine for freeing memory and nullifying pointers used in the example
+    subroutine clean_up()
+      nullify(out_data)
+      call torch_tensor_delete(a)
+      call torch_tensor_delete(b)
+      call torch_tensor_delete(Q)
+    end subroutine clean_up
+
```
jatkinson1000, line 105: Any particular reason for placing cleanup in a separate subroutine?

pages/autograd.md
```diff
@@ -0,0 +1,42 @@
+title: Online training
+
+[TOC]
+
+## Current state
+
+FTorch has supported offline training of ML models for some time. We are
+currently working on extending its functionality to support online training,
+too. This will involve exposing the automatic differentiation and
+back-propagation functionality in PyTorch/libtorch.
```
jatkinson1000, line 10:
```suggestion
back-propagation functionality in PyTorch/LibTorch.
```

pages/developer.md
```diff
@@ -77,6 +77,14 @@ and many of our users wish to _"clone-and-go"_ rather than develop, we provide b
 Development should only take place in `ftorch.fypp`, however._
 
 
+### Torch C++ API
+
+When extending or modifying functionality related to C++ header and/or source
+files `src/ctorch.h` and `src/ctorch.cpp`, we refer to the Torch C++
+[API documentation](https://pytorch.org/cppdocs/api/classat_1_1_tensor.html#exhale-class-classat-1-1-tensor)
```
jatkinson1000, line 84:
```suggestion
[API documentation](https://pytorch.org/cppdocs/api/classat_1_1_tensor.html)
```

Simplify the link slightly.
I was wondering why you specifically chose the tensor docs rather than the general API docs, but I suppose anything autograd related will be in the tensor?

However, since this section is more generic I wonder if we should instead link to https://pytorch.org/cppdocs/ or https://pytorch.org/cppdocs/api/library_root.html ?

pages/examples.md
```diff
@@ -187,9 +187,16 @@ data to multiple GPU devices.
...
 multiple input tensors and multiple output tensors.
 
-#### 7) Autograd
+#### 5) Looping
 
-[This worked example](https://github.com/Cambridge-ICCS/FTorch/tree/main/examples/5_Autograd)
+[This worked example](https://github.com/Cambridge-ICCS/FTorch/tree/main/examples/5_Looping)
+demonstrates best practices for performing inference on the same network with
+different input multiple times in the same workflow.
+
```
jatkinson1000, line 195: This is probably my bad, thanks for catching and updating.

src/ctorch.cpp
```diff
@@ -1,3 +1,8 @@
+/*
+ * See
+ * https://pytorch.org/cppdocs/api/classat_1_1_tensor.html#exhale-class-classat-1-1-tensor
+ * for more details on the Torch Tensor C++ API.
+ */
```
jatkinson1000, line 5: Again, do we want to point at the tensor docs, or would we be better off pointing at the generl C++ API docs.
This file handles torch_module and other things besides tensors.

jatkinson1000 (approved):
Thanks @jwallwork23 
This is a huge step, and really great work.

I have left a couple of queries/suggestions, but the code itself looks good.
I will approve, so feel free to look at my comments as you please and merge.

examples/6_Autograd/autograd.f90
jatkinson1000: Yeah, that's a pain.
See my review below, but I wonder if pointing at the generic API docs rather specifically tensor would be preferable.

examples/6_Autograd/autograd.f90
jwallwork23: Because it gets called upon hitting an error code as well as upon normal termination of the program.

pages/developer.md
jwallwork23: Ah yes, this is my bad. The tensor link was going to go in the docstrings for the overloaded operators but then I decided to just put one link (because of the discussion above).

Addressed in e82a9dbf3fb9090e64cb47402ed206b9a924dc75.

src/ctorch.cpp
jwallwork23: Addressed in e82a9dbf3fb9090e64cb47402ed206b9a924dc75. (As discussed above.)

examples/6_Autograd/autograd.f90
jwallwork23: Addresssed in e82a9dbf3fb9090e64cb47402ed206b9a924dc75.

PR #143, opened 2024-06-26
Adjust the API in anticipation of future overloading

jatkinson1000:
Change the API to make assignment of data to a `torch_tensor` a subroutine.

This is in anticipation of the AutoGrad wotk led by @jwallwork23 in #139 
This work may take a while to complete, but we need to stablise the API in advance of the summer school and as we move towards a first release and JOSS.

This makes the changes to the API that are required for this work without changing anything underneath.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,8 @@ type(torch_tensor), dimension(n_inputs) :: model_inputs_arr
 type(torch_tensor) :: model_output
 ...
 model = torch_module_load("/my/saved/TorchScript/model.pt")
-model_inputs_arr(1) = torch_tensor_from_array(input_fortran, in_layout, torch_kCPU)
-model_output = torch_tensor_from_array(output_fortran, out_layout, torch_kCPU)
+call torch_tensor_from_array(model_inputs_arr(1), input_fortran, in_layout, torch_kCPU)
+call torch_tensor_from_array(model_output, output_fortran, out_layout, torch_kCPU)
 
 call torch_module_forward(model, model_input_arr, n_inputs, model_output)
 ```
--- a/README.md
+++ b/README.md
@@ -38,15 +38,15 @@ examples for performing coupling.
 ```fortran
 use ftorch
 ...
-type(torch_module) :: model
-type(torch_tensor), dimension(n_inputs) :: model_inputs_arr
-type(torch_tensor) :: model_output
+type(torch_model) :: model
+type(torch_tensor), dimension(n_inputs)  :: model_inputs_arr
+type(torch_tensor), dimension(n_outputs) :: model_output_arr
 ...
-model = torch_module_load("/my/saved/TorchScript/model.pt")
+call torch_model_load(model, "/my/saved/TorchScript/model.pt")
 call torch_tensor_from_array(model_inputs_arr(1), input_fortran, in_layout, torch_kCPU)
-call torch_tensor_from_array(model_output, output_fortran, out_layout, torch_kCPU)
+call torch_tensor_from_array(model_output_arr(1), output_fortran, out_layout, torch_kCPU)
 
-call torch_module_forward(model, model_input_arr, n_inputs, model_output)
+call torch_model_forward(model, model_input_arr, model_output_arr)
 ```
 
 The following presentations provide an introduction and overview of _FTorch_:
--- a/examples/1_SimpleNet/simplenet_infer_fortran.f90
+++ b/examples/1_SimpleNet/simplenet_infer_fortran.f90
@@ -21,7 +21,7 @@ program inference
 
    ! Set up Torch data structures
    ! The net, a vector of input tensors (in this case we only have one), and the output tensor
-   type(torch_module) :: model
+   type(torch_model) :: model
    type(torch_tensor), dimension(1) :: in_tensors
    type(torch_tensor), dimension(1) :: out_tensors
 
@@ -37,17 +37,17 @@ program inference
 
    ! Create Torch input/output tensors from the above arrays
    call torch_tensor_from_array(in_tensors(1), in_data, tensor_layout, torch_kCPU)
-   call torch_tensor_from_array(out_tensor, out_data, tensor_layout, torch_kCPU)
+   call torch_tensor_from_array(out_tensors(1), out_data, tensor_layout, torch_kCPU)
 
    ! Load ML model
-   model = torch_module_load(args(1))
+   call torch_model_load(model, args(1))
 
    ! Infer
-   call torch_module_forward(model, in_tensors, out_tensors)
+   call torch_model_forward(model, in_tensors, out_tensors)
    write (*,*) out_data(:)
 
    ! Cleanup
-   call torch_module_delete(model)
+   call torch_model_delete(model)
    call torch_tensor_delete(in_tensors(1))
    call torch_tensor_delete(out_tensors(1))
 
--- a/examples/1_SimpleNet/simplenet_infer_fortran.f90
+++ b/examples/1_SimpleNet/simplenet_infer_fortran.f90
@@ -36,8 +36,8 @@ program inference
    in_data = [0.0, 1.0, 2.0, 3.0, 4.0]
 
    ! Create Torch input/output tensors from the above arrays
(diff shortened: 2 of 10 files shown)


jatkinson1000:
One thought @jwallwork23 and @TomMelt:

After this most things will be subroutines to be `call`ed from Fortran.

Do we want to update the module loading to be:
```
torch_module_load(torch_model, filename, device_type)
```
rather than:
```
torch_model = torch_module_load(filename, device_type)
```
?

And do we want to make it `torch_model_load()` rather than `torch_module_load`?
I have had a couple of comments in the past about people misunderstanding what is meant by 'module' in this context (A Neural Net subclassed from PyTorch's `nn.Module`).

src/ftorch.f90
```diff
@@ -249,7 +251,7 @@ end subroutine torch_tensor_print
   !> Determines the device index of a tensor.
   function torch_tensor_get_device_index(tensor) result(device_index)
     use, intrinsic :: iso_c_binding, only : c_int
-    type(torch_tensor), intent(in) :: tensor  !! Input tensor
+    type(torch_tensor), value, intent(in) :: tensor  !! Input tensor
```
dorchard, line 254: Why do we need this to be pass-by-value here for get_device_index? Won't this create an unnecessary copy. I don't think torch_tensor_get_device_index would do any modification but is there another reason?

dorchard:
I like the naming of `torch_model_load` - it might indeed be more clear.

src/ftorch.f90
jatkinson1000: This was a change @jwallwork23 made so he can advise.
It may be that is is unnecessary for this part but required in the overloading for some reason.

jwallwork23: You're right, this isn't needed. Thanks @dorchard.

jwallwork23:
Thanks very much for this @jatkinson1000! I was going to suggest doing this separately, too.

> Do we want to update the module loading to be:
> 
> ```
> call torch_module_load(torch_model, filename, device_type)
> ```
> 
> rather than:
> 
> ```
> torch_model = torch_module_load(filename, device_type)
> ```

Once the assignment operator is overloaded for tensors, any function that returns a tensor will implicitly do an additional copy, so it's probably best to turn any functions that return tensors into subroutines.
 
> And do we want to make it `torch_model_load()` rather than `torch_module_load`?

I agree with @dorchard that `torch_model_load` is clearer.

src/ftorch.f90
jatkinson1000: Coolio, I'll remove in due course.

jatkinson1000: Should be covered in https://github.com/Cambridge-ICCS/FTorch/pull/143/commits/7a35805a56139a2209301fd31c619dce9608913a

jatkinson1000:
The latest commits refer to the model as a '`model`' rather than a `module` (note that it still has to be referred to as a module when binding to C++ as this is what the Torch API uses).

It also makes the module loading a subroutine rather than a function.
I have left the get device as a function as this feels more appropriate in this instance.

README.md
```diff
@@ -38,15 +38,15 @@ examples for performing coupling.
...
-type(torch_tensor), dimension(n_inputs) :: model_inputs_arr
-type(torch_tensor) :: model_output
+type(torch_model) :: model
+type(torch_tensor), dimension(n_inputs)  :: model_inputs_arr
+type(torch_tensor), dimension(n_outputs) :: model_output_arr
 ...
-model = torch_module_load("/my/saved/TorchScript/model.pt")
-model_inputs_arr(1) = torch_tensor_from_array(input_fortran, in_layout, torch_kCPU)
-model_output = torch_tensor_from_array(output_fortran, out_layout, torch_kCPU)
+call torch_module_load(model, "/my/saved/TorchScript/model.pt")
```
dorchard, line 45: should be `torch_model_load`

README.md
```diff
@@ -38,15 +38,15 @@ examples for performing coupling.
...
+call torch_tensor_from_array(model_inputs_arr(1), input_fortran, in_layout, torch_kCPU)
+call torch_tensor_from_array(model_output_arr(1), output_fortran, out_layout, torch_kCPU)
 
-call torch_module_forward(model, model_input_arr, n_inputs, model_output)
+call torch_module_forward(model, model_input_arr, model_output_arr)
```
dorchard, line 49: should be `torch_model_forward`

dorchard (changes requested):
Looks good but a couple of documentation updates needed. Also [pages/troubleshooting.md] has a left over `torch_module_forward`

README.md
dorchard, line 45:
```suggestion
call torch_model_load(model, "/my/saved/TorchScript/model.pt")
```

README.md
dorchard:
```suggestion
call torch_model_forward(model, model_input_arr, model_output_arr)
```

dorchard (changes requested):
Update API in documentation

README.md
jatkinson1000: Should be covered in a sneaky rebase: https://github.com/Cambridge-ICCS/FTorch/pull/143/commits/8c1c98b68990c68202fb8a5e7e4fe1dffb46d96e

README.md
jatkinson1000: Should be covered in a sneaky rebase: https://github.com/Cambridge-ICCS/FTorch/pull/143/commits/8c1c98b68990c68202fb8a5e7e4fe1dffb46d96e

jatkinson1000:
Thanks @dorchard I have addressed those in a rebase to keep history clean.
Ready for a re-review I think.

dorchard (approved):
lgtm.

examples/1_SimpleNet/simplenet_infer_fortran.f90
```diff
@@ -36,18 +36,18 @@ program inference
    in_data = [0.0, 1.0, 2.0, 3.0, 4.0]
 
    ! Create Torch input/output tensors from the above arrays
-   in_tensors(1) = torch_tensor_from_array(in_data, tensor_layout, torch_kCPU)
-   out_tensors(1) = torch_tensor_from_array(out_data, tensor_layout, torch_kCPU)
+   call torch_tensor_from_array(in_tensors(1), in_data, tensor_layout, torch_kCPU)
+   call torch_tensor_from_array(out_tensors(1), out_data, tensor_layout, torch_kCPU)
 
    ! Load ML model
-   model = torch_module_load(args(1))
```
jwallwork23, line 43: This wouldn't be affected by overloading of tensors, but I guess it's good to be consistent.

examples/3_MultiGPU/simplenet_infer_fortran.f90
```diff
@@ -48,30 +48,30 @@ program inference
    ! Create Torch input tensor from the above array and assign it to the first (and only)
    ! element in the array of input tensors.
    ! We use the torch_kCUDA device type with device index corresponding to the MPI rank.
-   in_tensors(1) = torch_tensor_from_array(in_data, tensor_layout, torch_kCUDA, &
-                                          device_index=rank)
+   call torch_tensor_from_array(in_tensors(1), in_data, tensor_layout,         &
+                           torch_kCUDA, device_index=rank)
```
jwallwork23, line 52: Inconsistent indentation

jwallwork23 (approved):
Approved with one minor optional request and a comment.

examples/1_SimpleNet/simplenet_infer_fortran.f90
jatkinson1000: Yes, see discussion further up.

Also see that I have left `get_device_index` as a function as that feels more appropriate.

examples/3_MultiGPU/simplenet_infer_fortran.f90
jatkinson1000: Should be addressed in https://github.com/Cambridge-ICCS/FTorch/pull/143/commits/bfbbce9e3976df19615315589e3f163aa04cc0e7

PR #150, opened 2024-07-03
Fix `requires_grad` optional argument

### LINKED ISSUE #149: Fix `requires_grad` optional argument
Currently the `requires_grad` optional argument to various tensor constructors expects `requires_grad_opt`. We should use `requires_grad`, for consistency with PyTorch.

jwallwork23:
Closes #149.

Needed to switch `requires_grad_opt` -> `requires_grad` and use `requires_grad_value` to handle defaults (as with `device_index`).

Tested in my autograd example.

### DIFF
--- a/examples/2_ResNet18/resnet_infer_fortran.f90
+++ b/examples/2_ResNet18/resnet_infer_fortran.f90
@@ -95,7 +95,7 @@ contains
       probability = maxval(probabilities)
 
       ! Check top probability matches expected value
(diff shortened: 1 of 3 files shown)


jwallwork23:
Similarly for `is_training`.

jwallwork23:
Oh and for `rtol` in the ResNet example, for consistency.

jatkinson1000 (approved):
Thanks @jwallwork23 
This is all sensible, thanks.

I suggest you add docs for these as part of #139 or similar.

PR #155, opened 2024-07-05
add subroutine to delete arrays of tensors

TomMelt:
This PR adds a subroutine (`torch_tensor_array_delete`) to delete arrays of tensors. This eliminate the need to manually loop over all elements of the array calling `torch_tensor_delete`, see e.g.,

https://github.com/Cambridge-ICCS/FTorch/blob/490272ef5662936be26697d496b14f6a281214fa/examples/4_MultiIO/multiionet_infer_fortran.f90#L57-L60

Which now becomes,

https://github.com/Cambridge-ICCS/FTorch/blob/e24c38024e0ec8b03ecd0f8a3e6c269e42ac2a0c/examples/4_MultiIO/multiionet_infer_fortran.f90#L57-L58

- [x] add `torch_tensor_array_delete` to delete arrays of tensors
- [x] update tests
- [x] update docs

### DIFF
--- a/src/ftorch.f90
+++ b/src/ftorch.f90
@@ -270,9 +270,20 @@ contains
     device_index = torch_tensor_get_device_index_c(tensor%p)
   end function torch_tensor_get_device_index
 
+  !> Deallocates an array of tensors.
+  subroutine torch_tensor_array_delete(tensor_array)
+    type(torch_tensor), dimension(:), intent(inout) :: tensor_array
+    integer :: i
+
+    ! use bounds rather than (1, N) because it's safer
+    do i = lbound(tensor_array, dim=1), ubound(tensor_array, dim=1)
+      call torch_tensor_delete(tensor_array(i))
+    end do
+  end subroutine torch_tensor_array_delete
+
   !> Deallocates a tensor.
   subroutine torch_tensor_delete(tensor)
-    type(torch_tensor), intent(in) :: tensor     !! Input tensor
+    type(torch_tensor), intent(inout) :: tensor
 
     interface
       subroutine torch_tensor_delete_c(tensor) &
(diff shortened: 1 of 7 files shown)


jatkinson1000 (commented):
@TomMelt First glance looks good.

However, we'll need any examples in the documentation and readme also updating if you can go through these as well please?

src/ftorch.f90
```diff
@@ -270,9 +270,20 @@ end function torch_tensor_get_device_index_c
...
+    ! use bounds rather than (1, N) because it's safer
+    do i = lbound(tensor_array, dim=1), ubound(tensor_array, dim=1)
+      call torch_tensor_delete(tensor_array(i))
+    end do
+  end subroutine torch_tensor_array_delete
+
   !> Deallocates a tensor.
   subroutine torch_tensor_delete(tensor)
-    type(torch_tensor), intent(in) :: tensor     !! Input tensor
+    type(torch_tensor), intent(inout) :: tensor
```
TomMelt, line 286: Ah, I should also point out that this was technically an error. It should have been `inout` because we technically alter the contents.

TomMelt:
> @TomMelt First glance looks good.
> 
> However, we'll need any examples in the documentation and readme also updating if you can go through these as well please?

done

jatkinson1000:
@TomMelt Thanks - I wonder if an addition to the API changes page to point out this new option might be useful, though not essential as the old function is preserved?

Also I added a comment above with an edit for your thoughts:

> I also wonder if your original suggestion of having an interface to delete both an array or a single tensor is a better to simplify the subroutine options. However, I also see that the explicitness of two different routines is more 'Fortranic'.

jwallwork23 (approved):
Nice. Neatens things up a bit.

jatkinson1000:
A new thought on looking at some code today - perhaps we could have an interface `torch_delete()` that can take a model, tensor, or array of tensors and call the relevant deletion subroutine?

PR #156, opened 2024-07-10
add interface to delete all torch objects

TomMelt:
This PR builds on #155 

This PR adds one centralized interface that is capable of deleting multiple torch objects:

https://github.com/Cambridge-ICCS/FTorch/blob/2aae63ab73b8166db7eca55574e580a199629619/src/ftorch.f90#L77-L82


jatkinson1000 (approved):
Thanks @TomMelt This all looks good to be merged, thank you.

PR #159, opened 2024-07-22
Update README.md with citations

jatkinson1000:
Update README.md with citations for Mansfield and Sheshadri and Heuer et al papers using FTorch.


jwallwork23 (approved):
New URLs work and all DOIs are correct.

PR #161, opened 2024-07-22
Implement `torch_tensor_to_array`

### LINKED ISSUE #151: Implement `torch_tensor_to_array`
FTorch implements `torch_tensor_from_array` to map Fortran arrays to (F)Torch tensors. It would be useful to have the reverse mapping, too.

This would provide a way to get around the issue mentioned in #139.

jwallwork23:
Closes #151.

Implements `torch_to_blob` on the C++ side and `torch_tensor_to_array` on the Fortran side.

Updates example 1 to use this.


jatkinson1000 (commented):
Looking good @jwallwork23 

I have a question about the update to example 1 - presumably the use of `torch_tensor_to_array` is redundant here if you had not also changed the `out_data` to `pointer`?

My feeling is that this adds additional layers of confusion and is not what this example is intended to show as people's first entry point to FTorch. Perhaps there could be a separate example for introducing the autograd features?

Apologies if I have misunderstood and this is a change that is required across all of the code!

jwallwork23:
> I have a question about the update to example 1 - presumably the use of `torch_tensor_to_array` is redundant here if you had not also changed the `out_data` to `pointer`?

Agreed. I was planning to switch this to `torch_tensor_empty`, but this hasn't been implemented yet.

> My feeling is that this adds additional layers of confusion and is not what this example is intended to show as people's first entry point to FTorch. Perhaps there could be a separate example for introducing the autograd features?

Yeah, fair enough. Perhaps this PR could start the autograd demo as just "to tensor" and then "to array".

jwallwork23:
This is now ready for review, with thanks to @TomMelt for input on the C++ side.

src/ctorch.cpp
```diff
@@ -7,7 +7,8 @@ constexpr auto get_dtype(torch_data_t dtype)
 {
   switch (dtype) {
   case torch_kUInt8:
-    return torch::kUInt8;
+    std::cerr << "[WARNING]: uint8 not supported" << std::endl;
+    exit(EXIT_FAILURE);
```
jwallwork23, line 11: This is assuming the `not supported in Fortran` comment on the Fortran side is accurate.

src/ctorch.cpp
```diff
@@ -17,7 +18,8 @@ constexpr auto get_dtype(torch_data_t dtype)
   case torch_kInt64:
     return torch::kInt64;
   case torch_kFloat16:
-    return torch::kFloat16;
+    std::cerr << "[WARNING]: float16 not supported" << std::endl;
+    exit(EXIT_FAILURE);
```
jwallwork23, line 22: Similarly here.

src/ctorch.cpp
jatkinson1000: Consider making the error message more specific: "uint8 not supported in Fortran"?

src/ctorch.cpp
jatkinson1000: Consider making the error message more specific: "float16 not supported in Fortran"?

src/ctorch.cpp
```diff
@@ -173,6 +175,42 @@ torch_tensor_t torch_from_blob(void* data, int ndim, const int64_t* shape,
...
+      break;
+    case torch_kFloat16:
+      std::cerr << "[WARNING]: float16 not supported" << std::endl;
+      exit(EXIT_FAILURE);
+    case torch_kFloat32:
+      raw_ptr = (void*) t->data_ptr<float>();
+      break;
+    case torch_kFloat64:
+      raw_ptr = (void*) t->data_ptr<double>();
+      break;
```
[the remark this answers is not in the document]
jatkinson1000: Is it worth a comment somewhere to note this in case we end up having a use case in future?
I agree at the moment we don't want C++23!

examples/5_Autograd/requirements.txt
jatkinson1000, line 1: This file is empty, can it be removed, or should it have `torch` in there?

examples/5_Autograd/autograd.f90
```diff
@@ -0,0 +1,35 @@
...
+
+  ! Set up Torch data structures
+  type(torch_tensor) :: a
+
+  ! Construct a Torch Tensor from a Fortran array
+  in_data(:) = [2.0, 3.0]
+  call torch_tensor_from_array(a, in_data, tensor_layout, torch_kCPU)
+
+  ! Extract a Fortran array from a Torch tensor
+  allocate(out_data(2))
```
jatkinson1000, line 27: I wonder if it is possible/sensible to hide away this `allocate` within `torch_tensor_from_array`?
Or whether users should be expected to provide `out_data` to be the correct size as expected.

examples/5_Autograd/README.md
jatkinson1000, line 1: Sounds like a good plan!

jatkinson1000 (commented):
Thanks @jwallwork23 left you a couple of comments on there.

examples/5_Autograd/requirements.txt
jwallwork23: It's required by the `run_integration_tests.sh` script and the `.github/workflows/test_suite.yml` workflow, both of which attempt to pip install the dependencies.

Either we keep this empty file or we change those to only install if the file exists.

examples/5_Autograd/autograd.f90
jwallwork23: IIRC with `pointer` arrays you can't provide a size at declaration - they need to be allocated.

Yes, we could handle the allocation inside `torch_tensor_to_array` (presumably you meant this subroutine?). It will mean adding a `layout` argument, which is a good idea anyway as it improves consistency with `torch_tensor_from_array`. I'll implement it using an `if (.not. allocated(data_out)) ...` condition and a shape check, in case a user already allocated the array themselves.

src/ctorch.cpp
jwallwork23: Addressed in 806730c55e03b06ff48aa96ec13e4467d9b2a99e.

src/ctorch.cpp
jwallwork23: Addressed in 806730c55e03b06ff48aa96ec13e4467d9b2a99e.

src/ctorch.cpp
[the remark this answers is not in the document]
jwallwork23: Good idea. Addressed in 0f6ee05722b051f0a989dd0f535e4584efd7ff60.

examples/5_Autograd/autograd.f90
jwallwork23: Addressed in https://github.com/Cambridge-ICCS/FTorch/pull/161/commits/e2f6a9d473af461540de51ec64544d53728a9e92.

src/ftorch.fypp
```diff
@@ -489,4 +508,42 @@ contains
...
+      else if (any(shape(data_out) /= sizes)) then
+        write (*,*) "[ERROR]: Array allocated with wrong shape"
+        stop
+      end if
+    else
+      if ((.not. associated(data_out)) .or. (all(shape(data_out) == 0))) then
+        write (*,*) "[ERROR]: Pointer array has not been allocated"
+        stop
+      end if
+    end if
```
jatkinson1000, line 538: As discussed in-person, this is somewhat nuanced from the way we interact with libtorch, and potential for having data that will be pointed to by an currently unallocated variable.

Slightly more detail in the comments to explain what/why we are doing will help when coming back to this in future.

jatkinson1000 (changes requested):
Thanks, @jwallwork23 

Code all looks good, but as discussed in-person a little more explanation in the comments around the tensor_to_array function would be useful please.

src/ftorch.fypp
jwallwork23: Thanks, should be addressed in 570bf21.

jatkinson1000 (approved):
Yep, this is very clear now. Thanks @jwallwork23

PR #166, opened 2024-09-13
Use assertions rather than print statements in tests

### LINKED ISSUE #142: Use assertions rather than printing in tests
> these tests will most likely fail with the intel compiler and possibly different version of `gfortran`. The default output is > quite different for both. We should think of another way to do this.
> 
> We could add something like the resnet test?>
> 
> https://github.com/Cambridge-ICCS/FTorch/blob/b40b03d4c39f607de754d34991a92703ccc36eb3/examples/2_ResNet18/resnet_infer_fortran.f90#L186C15-L214

_Originally posted by @TomMelt in https://github.com/Cambridge-ICCS/FTorch/pull/136#discussion_r1652313394_

jwallwork23:
Closes #142.

As part of this PR, we introduce a module for test utils, e.g., printing test results and making assertions.


jwallwork23:
One discussion to have about these changes is whether we're happy that the examples are becoming more verbose. The versions without assertions will still be used in the docs but (e.g.) `SimpleNet` now has some extra stuff in there that might confuse things for beginners.

src/utils.f90
```diff
@@ -0,0 +1,115 @@
...
+      else
+        print_result_value = print_result
+      end if
+
+      test_pass = (abs(got - expect) <= rtol_value * abs(expect))
+
+      if (print_result_value) then
+        write(message,'("relative tolerance = ", E11.4)') rtol_value
+        call test_print(test_name, message, test_pass)
+      end if
```
jwallwork23, line 65: We refactored the relative error check to avoid divide-by-zero type errors. However, a downside is that we can no longer print the relative error values upon error without further computation and special case handling that we wanted to avoid. Instead, I just print the relative tolerance used.

Perhaps we could just print the two values being compared?

jwallwork23:
Huh! The CI has spontaneously hit a shape error again
https://github.com/Cambridge-ICCS/FTorch/actions/runs/10880645825/job/30187925536

As suggested by @TomMelt, we should use valgrind to see if there's a memory leak.

jatkinson1000:
I'm getting a similar issue over in #173 
I haven't touched the test, but ab getting`[ERROR]: Array allocated with wrong shape` from the `torch_tensor_to_array()` subroutine.

Digging into this the error seems to occur when processing the shape of an unallocated Fortran array.
It seems that ` all(shape(data_out) == 0)` returns `F`, but not because `shape(data_out)` is all zeros - indeed, calling `shape(data_out)` following this leads to a segfault due to an invalid memory reference for me.

I think I have found a fix which is to test `(.not. associated(data_out))` instead.
I have implemented this in https://github.com/Cambridge-ICCS/FTorch/pull/173/commits/e2b581df20156ce4cce4e71965625e35c2bc3fee as part of #173 which this may need rebasing on if merged.
Hopefully it will help here, though I am not 100% certain as the errors seem like they could be subtly different.

jatkinson1000:
Over in #174 I have modified the workflow to dump the error log from the failed test and output the arguments causing it to fail in the fortran.

See attached screenshot - despite sizes being 2 out_data seems to be allocated to have shape 95:

Screenshot 2024-10-01 at 10 50 45

Whereas a 'random success' correctly sets both to 2:

image

jwallwork23:
Merged in @TomMelt's fixes from #175 and extended the test utils to consider other shapes than just 1D arrays. Now ready for review!

examples/1_SimpleNet/simplenet_infer_fortran.f90
```diff
@@ -46,9 +53,17 @@ program inference
    call torch_model_forward(model, in_tensors, out_tensors)
    write (*,*) out_data(:)
 
+   ! Check output tensor matches expected value
+   expected = [0.0, 2.0, 4.0, 6.0, 8.0]
+   test_pass = assert_real_array_1d(out_data, expected, test_name="SimpleNet", rtol=1e-5)
```
jatkinson1000, line 58: I wonder if it is worth having some stdout for user running this as an example, e.g. "Result of NN matches expected value"?

Same applies to other examples, but not a blocker.

jatkinson1000: e.g. I see in the autograd example it finishes with "Autograd example ran successfully".

jatkinson1000 (approved):
Looking good @jwallwork23 

Left a comment about clarifying the results of "tests" when users run them as examples, but it isn't a blocker.

I'm still not 100% on having the `ftorch_test_utils` module as part of the standard build, but we have discussed this together and I have no other great alternative suggestions.

jatkinson1000:
Also looks like you may need want to do a squash and merge (or interactive rebase your end, but squash acceptable here I think, and probably causes fewer tears).

examples/1_SimpleNet/simplenet_infer_fortran.f90
jwallwork23: Good point, added in e4b515c.

jatkinson1000: Haha, I hadn't even thought about putting a regex check in the CMake.
Happy for you to merge.

jwallwork23: Can't take credit for that - Tom had already introduced such a check. Okay cool, will merge!

PR #168, opened 2024-09-13
Windows CI build

TomMelt:
This PR adds windows CI running on latest version of windows with intel compilers version `2023.2`.

For green computing reasons I have only prepared an intel build. My plan is that we have:
- `gcc` on linux
- `clang` on mac; and
-  `intel` on windows

That way we can get nice coverage without hammering compute resources.


jatkinson1000 (commented):
Thanks @TomMelt
Looks sensible to me (as much as any windows code does...)

I think we need some documentation adding about how to run the test suite for Windows users (definitely online and maybe README, though perhaps online only to avoid pollution)

I know nothing about bat scripts, but is there a reason we apply it in a for loop for each example as part of the workflow rather than running all examples from the bat script as we do in `run_integration_tests.sh`?

.github/workflows/test_suite_windows.yml
```diff
@@ -0,0 +1,58 @@
+name: TestSuiteWindows
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
```
jwallwork23, line 6: Please could you add comments consistently as in the other workflows? My recent workflow PR made them all similar in this respect.

.github/workflows/test_suite_windows.yml
```diff
@@ -0,0 +1,58 @@
...
+          cmake --install build
+
+      - name: run tests
+        shell: cmd
+        run: |
+          cd src
+          set PATH=C:\hostedtoolcache\windows\Python\3.12.7\x64\Lib\site-packages;%PATH%
+          set PATH=C:\Program Files (x86)\FTorch\bin;%PATH%
+          set PATH=C:\hostedtoolcache\windows\Python\3.12.7\x64\Lib\site-packages\torch\lib;%PATH%
+          for /d %%i in (1_SimpleNet 2_ResNet18 4_MultiIO) do ..\run_integration_tests.bat build\test\examples\%%i
```
jwallwork23, line 58: Same question as @jatkinson1000 - can this loop not go inside the `run_integration_tests.bat` script?

run_integration_tests.bat
```diff
@@ -0,0 +1,4 @@
+pushd %1
+ctest
+if %errorlevel% neq 0 exit /b %errorlevel%
+popd
```
jwallwork23, line 4: This file needs some comments at the top to say what it is (i.e., that it's the Windows version of `run_integration_tests.sh`.

jwallwork23 (changes requested):
Thanks for putting in the effort to get this working @TomMelt. Just a few minor comments from me.

.github/workflows/test_suite_windows.yml
jwallwork23, line 1: Perhaps rename the other one to `test_suite_ubuntu.yml` for clarity. We could use reusable workflows to make it easier if we ever want to add another test workflow. I can look into this as a follow-up.

jwallwork23: Opened #201 and self-assigned.

.github/workflows/test_suite_windows.yml
TomMelt: fixed by [d177685](https://github.com/Cambridge-ICCS/FTorch/pull/168/commits/d177685da3263399d49e3b0deb2c57d4dcf66fde)

.github/workflows/test_suite_windows.yml
TomMelt: fixed by [a6ec9c2](https://github.com/Cambridge-ICCS/FTorch/pull/168/commits/a6ec9c2ef7d8e880b5d0efabee8737d24ff5f3e1)

TomMelt:
> I know nothing about bat scripts, but is there a reason we apply it in a for loop for each example as part of the workflow rather than running all examples from the bat script as we do in `run_integration_tests.sh`?

fixed by [a6ec9c2](https://github.com/Cambridge-ICCS/FTorch/pull/168/commits/a6ec9c2ef7d8e880b5d0efabee8737d24ff5f3e1)

.github/workflows/test_suite_windows.yml
```diff
@@ -0,0 +1,83 @@
...
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on pushes to the "main" branch, i.e., PR merges
+  push:
+    branches: [ "main" ]
+
+  # Triggers the workflow on pushes to open pull requests with code changes
+  pull_request:
+    paths:
+      - '.github/workflows/test_suite.yml'
```
jwallwork23, line 13: This should be specific to the `test_suite_windows.yml` config.

Similar change in `test_suite_ubuntu.yml` if it's renamed as such.

src/CMakeLists.txt
```diff
@@ -102,6 +102,12 @@ install(FILES "${CMAKE_BINARY_DIR}/modules/ftorch_test_utils.mod"
 
 # Build integration tests
 if(CMAKE_BUILD_TESTS)
+
+  if (WIN32)
```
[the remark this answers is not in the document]
TomMelt: resolved by #203

run_integration_tests.bat
TomMelt: resolved by b38397a

jwallwork23 (approved):
Looks good!

src/test/README.md
```diff
@@ -24,8 +24,11 @@ which tests passed and which failed for your build. Note that the examples have
...
-```
-This will automatically install any additional dependencies for the examples.
+Alternatively, run the helper script in the root FTorch directory. Depending on
+which operating system you are running, you will need:
+
+- `./run_integration_tests.sh` for unix (mac and linux)
+- `run_integration_tests.bat` for windows
+
+This will automatically install any additional dependencies for the examples and
+run the tests.
```
jatkinson1000, line 34: I wonder if we should clarify _Python_ dependencies (don't think there are any others??) and so it is advisable to be in a venv?

jatkinson1000 (changes requested):
Thanks @TomMelt This looks great, I think we just need the online documentation updating to match (in `pages/` - definitely `testing.md`, but worth a quick scan of anywhere else, including the Windows specific docs).

TomMelt:
> Thanks @TomMelt Looks sensible to me (as much as any windows code does...)
> 
> I think we need some documentation adding about how to run the test suite for Windows users (definitely online and maybe README, though perhaps online only to avoid pollution)
> 
> I know nothing about bat scripts, but is there a reason we apply it in a for loop for each example as part of the workflow rather than running all examples from the bat script as we do in `run_integration_tests.sh`?

solved by #203 and https://github.com/Cambridge-ICCS/FTorch/commit/b38397a8f3c7311fc1466f4490dd1d34665e2acb

src/test/README.md
TomMelt: resolved by 4a23616

src/CMakeLists.txt
```diff
@@ -101,6 +101,10 @@ install(FILES "${CMAKE_BINARY_DIR}/modules/ftorch_test_utils.mod"
 
 # Build integration tests
 if(CMAKE_BUILD_TESTS)
+
+  set (Python3_FIND_VIRTUALENV FIRST)
```
jatkinson1000, line 105: This is a CMake 3.15 feature, so perhaps we need to bump the dependency across FTorch to CMake 3.15?

I think this is still sufficiently old not to be an issue for any users.

jatkinson1000 (changes requested):
Thanks @TomMelt 

This is really great, and a significant hurdle - hopefully you learnt something useful, even if it is at the cost of becoming the "Windows guy". ;P 

The blocker for me is not only just to bump the version of CMake required by FTorch to be 3.15 (for use of the Python VIRTUALENV) feature (see specific comment).
Consider also updating in the example CMakeLists for consistency, and in the documentation. From past experience the easiest way to do this is to grep the full source for `3.5`.

Once that is done I am happy to approve.

Other question is whether there is anything you feel is particularly useful or unintuitive to highlight for Windows users within the workflow. Otherwise we can just update the Windows docs piece by piece as people raise things in future.

src/CMakeLists.txt
TomMelt: done in https://github.com/Cambridge-ICCS/FTorch/pull/168/commits/17d8db2589376ee4c0e5d92c9f99c38cba5720ba

TomMelt:
> Thanks @TomMelt
> 
> This is really great, and a significant hurdle - hopefully you learnt something useful, even if it is at the cost of becoming the "Windows guy". ;P
> 
> The blocker for me is not only just to bump the version of CMake required by FTorch to be 3.15 (for use of the Python VIRTUALENV) feature (see specific comment). Consider also updating in the example CMakeLists for consistency, and in the documentation. From past experience the easiest way to do this is to grep the full source for `3.5`.
> 
> Once that is done I am happy to approve.
> 
> Other question is whether there is anything you feel is particularly useful or unintuitive to highlight for Windows users within the workflow. Otherwise we can just update the Windows docs piece by piece as people raise things in future.

Thanks @jatkinson1000 , I have added a [concrete example](https://github.com/Cambridge-ICCS/FTorch/pull/168/commits/e7ebd4d9db1d8505566f574fb4a00840620db9da) to the online docs just in case anyone has difficulty (most likely my future self)

TomMelt:
> Thanks @TomMelt I think README.md and /pages/cmake.md still need the CMake versioin updating from 3.5 to 3.15.

done in https://github.com/Cambridge-ICCS/FTorch/pull/168/commits/df5340a595e9d4dad767aac2631c4d7e9ed339ca

jatkinson1000 (approved):
I'm happy with this now.

Thanks @TomMelt it has been a mammoth task you've done here!

PR #170, opened 2024-09-18
Add comment to readme.md for SimpleNet about setting compiler

dorchard:
A note to remind how to set the compiler in the examples if you built FTorch using a different compiler to that found automatically by CMake (this just bit me, bizarrely only after upgrading mac OS).


jatkinson1000 (commented):
The fact it bit you is probably justification to include it here as well as the documentation, but I think the same sentence should probably be added to the documentation for all examples, not just this one.

Thank you.

dorchard:
Slightly amended the text and added it across all examples.

jatkinson1000 (approved):
Punch it Chewie!

PR #173, opened 2024-10-01
Extend FTorch to cover 5-dimensional tensors.

### LINKED ISSUE #172: torch_tensor_from_array cannot convert 5D array into torch tensor
I'm working with ConvLSTM. It normally receives input data in 5 dimensions: `batch_size x sequence_length x channels x height x width`. when I use the following snippet:
      
```
      type(torch_model) :: model
      type(torch_tensor), dimension(1) :: in_tensors

      ! Set up the model inputs and output as Fortran arrays (data structures)
      real(wp), dimension(:,:,:,:,:), allocatable, target :: in_data ! 5D input tensor (BS, SL, C, H, W)

      integer, parameter :: in_dims = 5
      integer :: in_shape(in_dims) = [1, 48, 5, 100, 154]
      integer :: in_layout(in_dims) = [1, 2, 3, 4, 5]

      integer, parameter :: input_tensor_length = 3696000
      inp_filename = "./data/input_tensor.dat"

      allocate(in_data(in_shape(1), in_shape(2), in_shape(3), in_shape(4), in_shape(5)))

      call load_data(inp_filename, input_tensor_length, in_data)
      call torch_tensor_from_array(in_tensors(1), in_data, in_layout, torch_kCPU)
```

I run into this error:

`Error: There is no specific subroutine for the generic ‘torch_tensor_from_array’ at (1)`

jatkinson1000:
Closes #172 

We have had a request for 5D tensor support.
This is from the ConvLSTM: https://paperswithcode.com/method/convlstm which is a type of NN used to analyse gridded data.
Given our initial application (climate sciences) it would make sense to support this as standard.

I have extended the ranks argument in the fypp code to now support 5-D (rank 5) tensors.

One point to discuss is whether we want to support slightly higher in anticipation of other applications (Rank 6 or perhaps 7) bearing in mind that doing so adds quite a lot of code/functions in the f90 file.

We should also perhaps add documentation noting the limit in Rank, and perhaps describing how to extend it for specific cases.

### DIFF
--- a/src/ftorch.f90
+++ b/src/ftorch.f90
@@ -55,26 +55,32 @@ module ftorch
     module procedure torch_tensor_from_array_int8_2d
     module procedure torch_tensor_from_array_int8_3d
     module procedure torch_tensor_from_array_int8_4d
+    module procedure torch_tensor_from_array_int8_5d
     module procedure torch_tensor_from_array_int16_1d
     module procedure torch_tensor_from_array_int16_2d
     module procedure torch_tensor_from_array_int16_3d
     module procedure torch_tensor_from_array_int16_4d
+    module procedure torch_tensor_from_array_int16_5d
     module procedure torch_tensor_from_array_int32_1d
     module procedure torch_tensor_from_array_int32_2d
     module procedure torch_tensor_from_array_int32_3d
     module procedure torch_tensor_from_array_int32_4d
+    module procedure torch_tensor_from_array_int32_5d
     module procedure torch_tensor_from_array_int64_1d
     module procedure torch_tensor_from_array_int64_2d
     module procedure torch_tensor_from_array_int64_3d
     module procedure torch_tensor_from_array_int64_4d
+    module procedure torch_tensor_from_array_int64_5d
     module procedure torch_tensor_from_array_real32_1d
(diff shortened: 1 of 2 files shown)


jatkinson1000:
As [discussed](https://github.com/Cambridge-ICCS/FTorch/pull/166#issuecomment-2385111680) over in #166 The tests for autograd were producing an error after this change, despite not touching them.

This could be fixed locally on my machine by checking for `(.not. associated(out_data))` rather than the shape of unassociated/unallocated `out_data`.

However, the code is still failing the CI test here.
My guess would be that this is perhaps due to a regex mismatch that is being dealt with over in #166 
So question is whether to:
- merge as is here
- turn autograd test off for now until fixed in #166 as not currently an active feature for users

jatkinson1000:
I have just removed the commit that worked around the pointer issues that will be addressed by #175 

This should now be merged after:
- #175 
- #166

jatkinson1000:
Rebased after #175 merged to main.
@TomMelt's changes to regex matching in autograd test as part of #175 mean that #166 is no longer a blocker so this is ready for review.

jwallwork23 (approved):
Single character change PR, love it!

PR #174, opened 2024-10-01
142 test assert rebase

jatkinson1000:
This is a rebase of #166 on top of #173 to exaime if the CI test suite now passes without regex.


jatkinson1000:
See discussion in #175

PR #180, opened 2024-10-17
wrap libtorch rank and shape routines

TomMelt:
the tensor derived type now supports two methods:
- `get_rank`
- `get_shape`

These methods return the shape and rank of the tensor

### DIFF
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -228,6 +228,18 @@ int torch_tensor_get_device_index(const torch_tensor_t tensor)
   return t->device().index();
 }
 
+int torch_tensor_get_rank(const torch_tensor_t tensor)
+{
+  auto t = reinterpret_cast<torch::Tensor*>(tensor);
+  return t->sizes().size();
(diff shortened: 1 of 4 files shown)


jatkinson1000 (approved):
Thanks @TomMelt

This all reads correctly to me.
I have looked at the libtorch API, but will clarify that I have not run it myself.
I presume you have tested these routines yourself locally and they perform as expected?

I don't think we can insist on tests here (as they are lacking elsewhere in the code) but as we bring more of these "utility" routines across we should look at tackling #4 with more priority.
I suppose there will be some integration testing from the autograd example.

jwallwork23 (approved):
Looks good, thanks @TomMelt.

TomMelt:
> all reads correctly to me.

Thanks @jatkinson1000 . Spot on.

- I plan to add unit tests in the near future with `pFUnit`.
- I did check locally by manually modifying the integration test but I will put a proper test in for the `autograd` work

PR #182, opened 2024-10-21
Bump the minimum required CMake in all CMakeLists to 3.5

jatkinson1000:
As discussed in a meeting users currently get a warning that CMake < 3.5 will be deprecated when building our code.

Since the minimum version that ships with many distros is above this (e.g. 3.26 on RHEL8) we can probably safely bump this to suppress the warning without issue.

Note that >3.5 may not be the minimum required version in actuality - we have not done an exhaustive test - but it was agreed that we should proceed this way until we receive an issue stating that it does not work on a particular version.


jwallwork23 (approved):
Seems reasonable!

PR #189, opened 2024-11-12
Remove specification of fortran compiler standard from CMakeLists.txt

### LINKED ISSUE #188: Specification of fortran standard in CMakeLists raises a warning with intel.
As pointed out by @j-emberton the specification of the Fortran Standard in the CMakeLists by adding a `CMAKE_Fortran_FLAGS` flag raises a warning when using intel (and other non-gcc) compilers.

https://github.com/Cambridge-ICCS/FTorch/blob/main/src/CMakeLists.txt#L15-L17

This is because `-std` is gcc specific.
For reference the intel version would be `-stand=f08`.

There is no default way to set the Fortran standard in CMake.
It could be challenging to implement a workaround for every possible compiler that a user may specify (though this is an option).

My suggestion would perhaps be to remove the standard from CMakeLists and leave it only as something that is checked as part of the testing workflow.

This is because standards in Fortran operate differently to those in C++.
Most compilers support the latest standards by default.

See https://gitlab.kitware.com/cmake/cmake/-/issues/25006 and https://gitlab.kitware.com/cmake/cmake/-/issues/22235

jatkinson1000:
Potential to close #188 

Removes the specification of `-std=f2008` as a Fortran compiler flag in CMakeLists.

This is a gcc specific flag that raises warnings for other compilers e.g. intel.

Whilst we could implement if-else blocks for different compilers this may still cause unforeseen issues and I do not believe it is necessary - see discussions at https://gitlab.kitware.com/cmake/cmake/-/issues/25006 and https://gitlab.kitware.com/cmake/cmake/-/issues/22235

Most compilers enable the latest Fortran standards that they support by default (different to C++ and C).
We still enforce this flag during the test suite as a check that code is compliant, but it is more of a guard against the introduction of features from a new spec rather than to enable features that may otherwise be unavailable.

I think specifying that we need a Fortran compiler that supports the 2008 standard as part of the documentation is sufficient for this.

Therefore I propose removing these lines (which would be overwritten by any `CMAKE_Fortran_FLAGS` a user might specify anyway!!).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -12,10 +12,6 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
-if(NOT CMAKE_Fortran_FLAGS)
-    set(CMAKE_Fortran_FLAGS "-std=f2008")
-endif()
-
 include(CheckLanguage)
 if(ENABLE_CUDA)
   check_language(CUDA)


j-emberton (approved):
This seems reasonable to me. Interesting to learn about the differences between the C languages and Fortran, and the behaviour of the intel and gnu compilers.

TomMelt (approved):
Agreed. I don't think we need this for release builds. 

FWIW, for C/C++ if we wanted to set standards in a compiler agnostic way, we could use CMake's [Requiring Language Standards](https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#requiring-language-standards).

However, this is not supported for CMake (at least not presently, see [here](https://gitlab.kitware.com/cmake/cmake/-/issues/25006) for more info).

jatkinson1000:
@TomMelt Do you mean "not supported for Fortran" at present?

I think those are the same conversations and settings linked in my main comment...?

PR #191, opened 2024-11-18
Long long fix

### LINKED ISSUE #183: Incompatibility with MacOS arm libtorch
Building github runners and locally suggests there is an issue with the macOS version of libtorch outputting different types compared to linux for the functions introduced in #175 .

FTorch expects a `long int*` whilst the libtorch binary seems to be returning a `long long int*`.

Example output below.
```
[user@machine build]$ cmake --build .
[  8%] Building Fortran object CMakeFiles/ftorch.dir/ftorch.f90.o
[ 16%] Building Fortran object CMakeFiles/ftorch.dir/ftorch_test_utils.f90.o
[ 25%] Building CXX object CMakeFiles/ftorch.dir/ctorch.cpp.o
/Users/user/FTorch/src/ctorch.cpp: In function 'const long int* torch_tensor_get_sizes(torch_tensor_t)':
/Users/user/FTorch/src/ctorch.cpp:240:25: error: invalid conversion from 'const long long int*' to 'const long int*' [-fpermissive]
  240 |   return t->sizes().data();
      |          ~~~~~~~~~~~~~~~^~
      |                         |
      |                         const long long int*
make[2]: *** [CMakeFiles/ftorch.dir/ctorch.cpp.o] Error 1
make[1]: *** [CMakeFiles/ftorch.dir/all] Error 2
make: *** [all] Error 2
```

TomMelt:
closes #183 

## Problem
This issue also affected windows and is also fixed by this PR.

Tldr; `FTorch` expects a `long int*` whilst the `libtorch` binary seems to be returning a `long long int*`.

Example output below.
```
[ 25%] Building CXX object CMakeFiles/ftorch.dir/ctorch.cpp.o
/Users/user/FTorch/src/ctorch.cpp: In function 'const long int* torch_tensor_get_sizes(torch_tensor_t)':
/Users/user/FTorch/src/ctorch.cpp:240:25: error: invalid conversion from 'const long long int*' to 'const long int*' [-fpermissive]
  240 |   return t->sizes().data();
      |          ~~~~~~~~~~~~~~~^~
      |                         |
      |                         const long long int*
```
## Solution
The solution is to check which system we are running on (OS and architecture) and create a preprocess macro `-DUNIX` that we can use to switch between `long long int` and `long int` versions of the `libtorch` library. 
https://github.com/Cambridge-ICCS/FTorch/blob/5fb08873c1910f0eff7e6b0424e7ce63a365fcde/src/CMakeLists.txt#L57-L62

Then in the source code we check for that macro e.g.,
https://github.com/Cambridge-ICCS/FTorch/blob/5fb08873c1910f0eff7e6b0424e7ce63a365fcde/src/ctorch.h#L128-L132

## TODO
- [x] works on windows
- [x] works on linux
- [ ] works on mac (M1/M2)
- [x] create windows CI (this will be a separate PR) #168

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -52,7 +52,14 @@ set(CMAKE_INSTALL_RPATH $ORIGIN/${relDir})
 find_package(Torch REQUIRED)
 
 # Library with C and Fortran bindings
-add_library(${LIB_NAME} SHARED ctorch.cpp ftorch.f90 ftorch_test_utils.f90)
+add_library(${LIB_NAME} SHARED ctorch.cpp ftorch.F90 ftorch_test_utils.f90)
+
+if(UNIX)
+  if(NOT APPLE) # only add definition for linux (not apple which is also unix)
+    target_compile_definitions(${LIB_NAME} PRIVATE UNIX)
+  endif()
+endif()
+
 # Add an alias FTorch::ftorch for the library
 add_library(${PROJECT_NAME}::${LIB_NAME} ALIAS ${LIB_NAME})
 set_target_properties(${LIB_NAME} PROPERTIES
--- a/src/ctorch.h
+++ b/src/ctorch.h
@@ -42,10 +42,9 @@ typedef enum { torch_kCPU, torch_kCUDA } torch_device_t;
  * @param device index for the CUDA case
  * @param whether gradient is required
  */
-EXPORT_C torch_tensor_t torch_zeros(int ndim, const int64_t *shape,
-                                    torch_data_t dtype,
-                                    torch_device_t device_type,
-                                    int device_index, const bool requires_grad);
+EXPORT_C torch_tensor_t torch_zeros(int ndim, const int64_t *shape, torch_data_t dtype,
+                                    torch_device_t device_type, int device_index,
+                                    const bool requires_grad);
 
 /**
  * Function to generate a Torch Tensor of ones
@@ -56,8 +55,7 @@ EXPORT_C torch_tensor_t torch_zeros(int ndim, const int64_t *shape,
  * @param device index for the CUDA case
  * @param whether gradient is required
  */
-EXPORT_C torch_tensor_t torch_ones(int ndim, const int64_t *shape,
-                                   torch_data_t dtype,
+EXPORT_C torch_tensor_t torch_ones(int ndim, const int64_t *shape, torch_data_t dtype,
                                    torch_device_t device_type, int device_index,
                                    const bool requires_grad);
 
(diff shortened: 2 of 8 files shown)


src/CMakeLists.txt
```diff
@@ -52,7 +52,15 @@ set(CMAKE_INSTALL_RPATH $ORIGIN/${relDir})
 find_package(Torch REQUIRED)
 
 # Library with C and Fortran bindings
-add_library(${LIB_NAME} SHARED ctorch.cpp ftorch.f90 ftorch_test_utils.f90)
+add_library(${LIB_NAME} SHARED ctorch.cpp ftorch.F90 ftorch_test_utils.f90)
+
+if(UNIX)
+  message(STATUS "CMAKE_SYSTEM_PROCESSOR = ${CMAKE_SYSTEM_PROCESSOR}")
```
TomMelt, line 58:
```suggestion
```

Just remembered I left this print statement in for debugging. I should take this out

TomMelt:
> This looks sensible to me, @TomMelt. I haven't tested it on my laptop running Windows but I can do if needed.

Thanks @jwallwork23. Wouldn't hurt if you get time but I did test on windows vm so it should be fine :ok_hand: 

I do plan on setting up a windows CI soon so that should also catch any issues.

jatkinson1000:
It seems that this incompatibility still exists on Mac 12 and 13 - see https://github.com/Cambridge-ICCS/FTorch/pull/192

So the CMake/#ifdef might need some refinement.
IIRC 12 and 13 are X86 images rather than arm, but I'd need to check that.

src/ctorch.h
```diff
@@ -128,7 +128,12 @@ EXPORT_C int torch_tensor_get_rank(const torch_tensor_t tensor);
  * @param Torch Tensor to determine the rank of
  * @return pointer to the sizes array of the Torch Tensor
  */
+#ifdef UNIX
 EXPORT_C const long int *torch_tensor_get_sizes(const torch_tensor_t tensor);
+#else
+EXPORT_C const long long int *
+torch_tensor_get_sizes(const torch_tensor_t tensor);
```
jwallwork23, line 135: I'm surprised the linter likes these not being on the same line.

jwallwork23 (approved):
Approved once, will approve again. Thanks @TomMelt!

jatkinson1000 (approved):
As discussed in-person, I think this can go in now and the issues with mac can be resolved in a separate set of issues.

Thanks @TomMelt!

src/CMakeLists.txt
TomMelt: not yet. I will look at fixing the mac build before I take it out completely

src/ctorch.h
TomMelt: honestly... it makes me sad. I had them on the same line, but the linter forces it to split :(

jatkinson1000: Can maybe use a config file to extend the column limit?
https://clang.llvm.org/docs/ClangFormatStyleOptions.html

jatkinson1000: I'm a little confused, when I run clang-format locally it moves them to be placed on one line...

jwallwork23: I wonder if this is to do with the default settings being different for different clang-format versions? We observed this in nextSIM-DG and so pinned the version used by the CI.

jatkinson1000: I have just opened https://github.com/Cambridge-ICCS/FTorch/pull/199 to here which should resolve this by fixing Cpp Column Limit to 88.

jatkinson1000:
Have rebased #199 on this if you want to merge.

PR #198, opened 2024-12-04
DO NOT MERGE: Break the cpp in ctorch.cpp, shopuld be caught by clang checker.

jatkinson1000:


jatkinson1000:
It was indeed caught.
Closing as served purpose.

PR #199, opened 2024-12-04
Add clang-format config to set columnlimit for Cpp

jatkinson1000:
Sett Cpp column limit to 88 (as per ruff) for a nicer formatting based on observation.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,4 @@
+---
+Language: Cpp
+ColumnLimit: 88
+---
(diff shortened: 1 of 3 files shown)


TomMelt (approved):
thanks @jatkinson1000. I guess the new limit will always be arbitrary. So I guess 88 is as good as any other number. Is there a reason you went for 88? or just that it's visually pleasing :ok_hand:

PR #202, opened 2024-12-09
Add reusable test workflow

### LINKED ISSUE #201: Use reusable workflows
We could use reusable workflows to make it easier if we ever want to add another test workflow. I can look into this as a follow-up.

_Originally posted by @jwallwork23 in https://github.com/Cambridge-ICCS/FTorch/pull/168#discussion_r1875681977_

jwallwork23:
Closes #201.


jwallwork23:
Hm the workflows are sufficiently different that I don't think this would be worth it at this stage.

PR #203, opened 2024-12-09
test: make integration tests cross-platform

TomMelt:
this makes the integration tests more portable.

- [x] change permissions on python files so they are no longer executables
- [x] remove python shebang from said python files
- [x] update `CMakeLists.txt` to require python interpreter (for building tests only) and use `Python_EXECUTABLE` which is set by `find_package (Python COMPONENTS Interpreter REQUIRED)`

### DIFF
--- a/.github/workflows/test_suite_ubuntu.yml
+++ b/.github/workflows/test_suite_ubuntu.yml
@@ -68,7 +68,12 @@ jobs:
           export BUILD_DIR=$(pwd)/src/build
           mkdir ${BUILD_DIR}
           cd ${BUILD_DIR}
-          cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=${BUILD_DIR} -DCMAKE_BUILD_TESTS=TRUE -DCMAKE_Fortran_FLAGS="-std=${{ matrix.std }}"
+          cmake .. \
+            -DPython_EXECUTABLE="$(which python)" \
+            -DCMAKE_BUILD_TYPE=Release \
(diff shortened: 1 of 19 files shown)


jatkinson1000 (commented):
Seems sensible.

So when setting `Python_EXECUTABLE` this does now work OK on Windows?
Since we don't set this now.

GitHub flags the python scripts as changed, but then shows no lines changed listing it as an 'empty file'. Is this because it interprets the removal of a shebang in a strange way?

TomMelt:
> Seems sensible.
> 
> So when searching for python and implicitly setting `Python_EXECUTABLE` this does now work OK on Windows? Since we don't set this now as far as I can see.

It needs a tweak but it will work. It is picking up the wrong python I think. Should be fixed shortly.

> GitHub flags the python scripts as changed, but then shows no lines changed listing it as an 'empty file'. Is this because it interprets the removal of a shebang in a strange way?

No, it because the permissions were changed from executable (755) to read/write (644). There is no way for github to display that info. It's a bit like when you rename a file. There are zero changes to the file contents but the git metadata changes.

jatkinson1000:
@TomMelt OK, so are we leaving the shebangs in the files?
It's just there is a ticked off box saying they have been removed at the top, so I expected to see this in the file diff at the same time as changing the permissions.

TomMelt:
> @TomMelt OK, so are we leaving the shebangs in the files? It's just there is a ticked off box saying they have been removed at the top, so I expected to see this in the file diff at the same time as changing the permissions.

my bad... now I understand you. I hadn't pushed that commit yet :facepalm: 

You can now see the shebang removal :ok_hand:

PR #204, opened 2024-12-10
Apply CMake formatting

jwallwork23:
Today I discovered a CMake formatter: https://github.com/cheshirekow/cmake_format.

Seems like a good idea to add to the static analysis workflow. I did need to disable the formatter in a couple of places with particularly long lines.

(Most of the changes in this PR are auto-formatting changes.)

### DIFF
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -107,11 +107,11 @@ jobs:
           ruff format --diff ./
           ruff check --diff ./
 
-      # Apply CMake formatter, cmake-format
+      # Apply CMake linter, cmake-lint
       - name: cmake
         if: always()
         run: |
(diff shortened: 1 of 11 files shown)


jatkinson1000 (changes requested):
Nice @jwallwork23 

Changes all look good to me.
As ever, needs a short comment/link adding to the documentation ;)

I'd also suggest putting cmakelang in the requirements.txt rather than pip installing it on its own.

jwallwork23:
> Changes all look good to me. As ever, needs a short comment/link adding to the documentation ;)

My New Years resolution will be to write the docs with the code :angel:
 
> I'd also suggest putting cmakelang in the requirements.txt rather than pip installing it on its own.

Good point.

jatkinson1000 (approved):
Send it!

PR #137, opened 2024-06-19
Melt windows fix

TomMelt:
fixes #124 

Windows was not previously exporting the library symbols and creating the `dll` correctly.

This is now resolved by the following addition to `CMakeLists.txt`
https://github.com/Cambridge-ICCS/FTorch/blob/7856a33f9b81bad504da322c26f6f90c9b75a339/src/CMakeLists.txt#L6-L9

Furthemore, there was an issue that the library was built with a different integer size than that used when compiling example programs, despite no difference in compiler option. We have now _forced_ the library to use `int32` explicitly. This can be modified by the user, modifying the following line
https://github.com/Cambridge-ICCS/FTorch/blob/7856a33f9b81bad504da322c26f6f90c9b75a339/src/ftorch.fypp#L34

Tasks:
- [x] fix cmake windows build issue
- [x] fix `int64`/`int32` issue properly (we have decided to build int32 by default. Can be overridden by user.
- [x] update documentation to reflect new changes

PR #205, opened 2024-12-10
Bump to Fortitude 0.6.0

jatkinson1000:
Also pins version in requirements.txt

Mostly whitespace updates, though I addded some alignment of FORD variable docs at the same time.

In future we should keep an eye out for major release updates and do periodic bumps.

This should be merged after #137


jwallwork23 (approved):
Seems reasonable. Given that fortitude is being actively developed, we should remember to update this as things progress.

PR #206, opened 2024-12-16
Linting for shell and GitHub Actions

jwallwork23:
More linting!

Adds linting for shell and GitHub Actions workflows:
* https://github.com/koalaman/shellcheck
* https://woodruffw.github.io/zizmor

Also fixes an oversight in the CMake linting setup.

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -29,6 +29,8 @@ jobs:
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - name: Checkout code
+        with:
+          persist-credentials: false
         uses: actions/checkout@v4
 
       - name: Install Python
--- a/.github/workflows/fypp_checks.yml
+++ b/.github/workflows/fypp_checks.yml
@@ -23,6 +23,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+      with:
+        persist-credentials: false
     - uses: actions/setup-python@v4
       with:
         python-version: "3.11"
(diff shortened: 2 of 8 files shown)


jwallwork23:
My initial push at commit https://github.com/Cambridge-ICCS/FTorch/pull/206/commits/e128fe3a024a0c663580118091e68fba376644d4 shows the static analysis workflow failing because I'd only applied lint fixes to the `static_analysis.yml`. With my second push at commit https://github.com/Cambridge-ICCS/FTorch/pull/206/commits/5858a6be5d4e78db8523fe643145636cc932a8d0 it passes.

jwallwork23:
#207 has been merged into the branch here so should be merged first.

jatkinson1000 (changes requested):
They seem active projects.
I note zizmor is in beta, so perhaps we want to pin the version similar to fortitude, to avoid unexpected complaints.

We also need the docs updating to detail these, and I note that shellcheck is not pip installable. Seems installable via all sensible package managers. I wouldn't know where to start with Windows packaga managers, but don't see this as an issue as our docs state "follow the instructions in the linter docs".

jwallwork23:
> They seem active projects. I note zizmor is in beta, so perhaps we want to pin the version similar to fortitude, to avoid unexpected complaints.
> 
> We also need the docs updating to detail these, and I note that shellcheck is not pip installable. Seems installable via all sensible package managers. I wouldn't know where to start with Windows packaga managers, but don't see this as an issue as our docs state "follow the instructions in the linter docs".

Thanks @jatkinson1000. Addressed in https://github.com/Cambridge-ICCS/FTorch/pull/206/commits/2834064705796bc428b4f6145dda13ffaadbaea7.

jatkinson1000 (approved):
Thanks @jwallwork23 All LGTM now!

PR #215, opened 2025-01-06
Lean on compiler

jwallwork23:
Building in debug mode with gfortran 11.4.0 on Ubuntu 22.04, I came across several warnings, most of which are fixed in this PR. They are mostly minor but the ones to do with overloaded operators are an oversight of mine from #139.

### DIFF
--- a/examples/2_ResNet18/resnet_infer_fortran.f90
+++ b/examples/2_ResNet18/resnet_infer_fortran.f90
@@ -38,7 +38,7 @@ contains
       integer, parameter :: out_layout(out_dims) = [1, 2]
 
       ! Path to input data
-      character(len=100) :: data_dir
+      character(len=128) :: data_dir
       ! Binary file containing input tensor
       character(len=116) :: filename
       ! Text file containing categories
@@ -52,7 +52,7 @@ contains
       integer :: idx(2)
       real(wp), dimension(:,:), allocatable :: probabilities
       real(wp), parameter :: expected_prob = 0.8846225142478943_wp
-      character(len=100) :: categories(N_cats)
+      character(len=128) :: categories(N_cats)
       real(wp) :: probability
 
       ! Flag for testing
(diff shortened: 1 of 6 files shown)


examples/6_Autograd/autograd.f90
```diff
@@ -23,7 +23,6 @@ program example
   real(wp), dimension(:,:), pointer :: out_data
   real(wp), dimension(n,m) :: expected
   integer :: tensor_layout(2) = [1, 2]
-  integer :: i, j
```
jwallwork23, line 26: Unused variables.

src/ftorch.fypp
```diff
@@ -293,9 +293,12 @@ contains
...
 
-    strides(layout(1)) = 1
-    do i = 2, ndims
-      strides(layout(i)) = strides(layout(i - 1)) * tensor_shape(layout(i - 1))
+    do i = 1, ndims
+      if (i == 1) then
+        strides(layout(i)) = 1
+      else
+        strides(layout(i)) = strides(layout(i - 1)) * tensor_shape(layout(i - 1))
+      end if
```
jwallwork23, line 301: The original version is a zero loop if ndims = 1.

examples/2_ResNet18/resnet_infer_fortran.f90
```diff
@@ -38,7 +38,7 @@ subroutine main()
       integer, parameter :: out_layout(out_dims) = [1, 2]
 
       ! Path to input data
-      character(len=100) :: data_dir
```
jwallwork23, line 41: Number of characters was inconsistent with number of characters for command line input.

src/ftorch.fypp
jatkinson1000: Good spot!
I _suppose_ it did what was intended, but the rewrite is much more explicit and doesn't leave us open to unexpected intentions in future.

jatkinson1000 (approved):
Thanks @jwallwork23 all looks sensible to me and some good catches.

Slightly worrying that the C precision was working in the old version...?

Happy for you to merge.

jwallwork23:
> Slightly worrying that the C precision was working in the old version...?
> 
Agreed.

PR #76, opened 2023-11-28
Unit tests (using pFunit)

dorchard:
I have started putting in the infrastructure for us to write some unit tests using pFunit.

Create simple unit tests for main functions
- [x] Make the cmake portable; I fought a bit, but I don't know enough cmake. We need to be able to link the include directories for ftorch. At the moment the library and include paths need to be specified.
- [x] Test `torch_tensor_from_blob`
- [x] Test `torch_tensor_from_array`
- [x] Test `torch_tensor_zeros`
- [x] Test `torch_tensor_ones`
- [x] Test `torch_tensor_empty`

Other things using `torch_module_forward` could be better with an integration test since it will need a lot of setup code anyway.

Additionally, we could make use of fixtures and other features of pFUnit in a follow-up.

Closing this would fix #4

PR #217, opened 2025-01-06
Reorganise main mod

jwallwork23:
To follow #76.

This PR reorganises the main FTorch module to group things more logically:
1. Datatypes
2. Interfaces for core FTorch functionality
3. Interfaces for operator overloading of tensors
4. Procedures for constructing tensors
5. Procedures for interrogating tensors
6. Procedures for deallocating tensors
7. Overloaded operators for tensors
8. Torch model API

This PR also drops unnecessary imports and makes it so that the `device_index` argument can be passed as a Fortran `integer` rather than a `c_int`.

### DIFF
--- a/src/ftorch.fypp
+++ b/src/ftorch.fypp
@@ -26,11 +26,12 @@ $:'integer' if PRECISION[:3] == 'int' else 'real'
 module ftorch
 
   use, intrinsic :: iso_c_binding, only: c_null_ptr, c_ptr
-
-  ! Set integer size for FTorch library
-  use, intrinsic :: iso_fortran_env, only: ftorch_int => int32
+  use, intrinsic :: iso_fortran_env, only: int32
 
   implicit none
+
+  ! Set integer size for FTorch library
+  integer, parameter :: ftorch_int = int32
 
   ! --- Derived types and enums
 
@@ -481,7 +482,7 @@ contains
   !> Determines the rank of a tensor.
   function get_rank(self) result(rank)
     class(torch_tensor), intent(in) :: self
-    integer(kind=ftorch_int) :: rank  !! rank of tensor
+    integer(kind=int32) :: rank  !! rank of tensor
 
     interface
       function torch_tensor_get_rank_c(tensor) result(rank) &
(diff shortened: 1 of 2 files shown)


src/ftorch.fypp
```diff
@@ -404,7 +481,7 @@ contains
   !> Determines the rank of a tensor.
   function get_rank(self) result(rank)
     class(torch_tensor), intent(in) :: self
-    integer(kind=int32) :: rank  !! rank of tensor
+    integer(kind=ftorch_int) :: rank  !! rank of tensor
```
TomMelt, line 484: @jwallwork23 , I think we need to be careful here and in `get_shape`. Due to the C header, `rank` is a "standard" integer i.e.,

```c
EXPORT_C int torch_tensor_get_rank(const torch_tensor_t tensor);
```
I don't know what happens if `ftorch_int` is set to 64 bit. Will it be re-cast or will it turn into garbage? This definitely needs checking.

If we keep these changes we should try and add a unit test to check this. Due to how the library is structured though, I think it would require building twice.

Similarly for `get_shape` I also don't know if it gets handled correctly.

TomMelt (changes requested):
Thanks @jwallwork23 , this is a simple improvement and helps with maintainability.

My only concern is relating to the recently changed `int32` types which have now been converted to `ftorch_int` (which defaults to `int32` but could in theory change).

src/ftorch.fypp
jwallwork23: Good point, thanks @TomMelt. I think it's probably best to just revert this particular change for now.

jwallwork23: Addressed in 8eaf19f.

TomMelt (approved):
Thanks @jwallwork23 , LGTM now :rocket:

jwallwork23:
> Only suggestion would be to perhaps make the section headers even more obvious e.g.
> 
> ```
> ! =============================================
> ! = --- Derived types and enums                                              =
> ! =============================================
> ```
> 
> To stand out when people are browsing the source.

Addressed in c2e9c1b077d42326823a49bac80b4b1130432f4f.

jatkinson1000 (approved):
Go for it.

PR #222, opened 2025-01-07
Build example 3 if CUDA and MPI enabled

### LINKED ISSUE #208: Add `3_MultiGPU` example to build process
Currently this is commented out in `examples/CMakeLists.txt` and `src/CMakeLists.txt`. To add it, we'll need to use both `ENABLE_MPI` and `ENABLE_CUDA`.

jwallwork23:
Closes #208.


jatkinson1000 (commented):
@jwallwork23 

Yes, agree with this.
However, a comment that I have seen before elsewhere - it would be good to avoid "double negatives" - the guard for running _integration_ tests is based on the _unit_ test flag. This requires some mental gymnastics (unless it's just me) and would be a problem if we wanted to add a third set of subtests for some reason in future.

I wonder if moving to having `RUN_UNIT` and `RUN_INTEGRATION` variables that default to `true` and are checked against for running the tests, and setting them to `false` where necessary in setup would be preferable?

jwallwork23:
> Yes, agree with this. However, a comment that I have seen before elsewhere - it would be good to avoid "double negatives" - the guard for running _integration_ tests is based on the _unit_ test flag. This requires some mental gymnastics (unless it's just me) and would be a problem if we wanted to add a third set of subtests for some reason in future.
> 
> I wonder if moving to having `RUN_UNIT` and `RUN_INTEGRATION` variables that default to `true` and are checked against for running the tests, and setting them to `false` where necessary in setup would be preferable?

Opened #259 and self-assigned to address this.

jwallwork23:
Ah, just saw your commit 5935bb645d253d3e1572747a21d570c8cba8d120. Apologies, looks like I updated the README on the wrong branch (see #258).

jatkinson1000:
@jwallwork23 rebased so this can go into maion with #258 being re-targeted.

Happy to approve once you have a quick glance over to check you are happy with the rebase.
I did check it in #262 as you saw.

jwallwork23:
> @jwallwork23 rebased so this can go into maion with #258 being re-targeted.
> 
> Happy to approve once you have a quick glance over to check you are happy with the rebase. I did check it in #262 as you saw.

LGTM, thanks @jatkinson1000

jatkinson1000 (approved):
Reviewed, tested, amended, and rebased.
Approving now @jwallwork23 and happy for you to merge in.

PR #225, opened 2025-01-07
Unit tests for operator overloading; fix scalar bug

### LINKED ISSUE #221: Unit tests for operator overloads
- [ ] `torch_tensor_assign`
- [ ] `torch_tensor_add`
- [ ] `torch_tensor_subtract`
- [ ] `torch_tensor_multiply`
- [ ] `torch_tensor_divide`
- [ ] `torch_tensor_power`

### LINKED ISSUE #224: Operator overloads not working correctly for non-integer scalars
Unfortunately, there's an oversight in my implementation of the operator overloads that involve scalars. They only seem to currently work for integer scalars (only integer scalars are considered in the autograd demo).

For example, changing the divisor with
```diff
diff --git a/examples/6_Autograd/autograd.f90 b/examples/6_Autograd/autograd.f90
index cf60494..9d73853 100644
--- a/examples/6_Autograd/autograd.f90
+++ b/examples/6_Autograd/autograd.f90
@@ -52,7 +52,7 @@ program example
   ! Check arithmetic operations work for torch_tensors
   write (*,*) "a = ", in_data1(:,1)
   write (*,*) "b = ", in_data2(:,1)
-  Q = 3 * (a**3 - b * b / 3)
+  Q = 3 * (a**3 - b * b / 3.0)
 
   ! Extract a Fortran array from a Torch tensor
   call torch_tensor_to_array(Q, out_data, shape(in_data1))
```
on the `main` branch results in the test failure:
```console
test 2
    Start 2: fautograd

2: Test command: /home/joe/software/FTorch/src/build/test/examples/6_Autograd/autograd
2: Test timeout computed to be: 1500
2:  a =    2.00000000       3.00000000    
2:  b =    6.00000000       4.00000000    
2:  Q = 3 * (a ** 3 - b * b / 2) =   24.0000000       81.0000000    
2: FAILED :: [torch_tensor_to_array] relative tolerance =  0.1000E-04
2:  Error :: out_data does not match expected value
2: STOP 999
2/2 Test #2: fautograd ........................***Failed  Required regular expression not found. Regex=[Autograd example ran successfully
]  0.23 sec

```

jwallwork23:
Closes #221.
Closes #224.

This PR implements unit tests for the overloaded operators acting on tensors using funit.

While implementing these tests, I discovered that pre/post multiplication by scalar, post-division by scalar, and scalar exponent don't work if the scalar is non-integer. This was an oversight by me, which is fixed in this PR. It turns out you can multiply/divide any tensor by a [1] tensor and torch treats it as a scalar, so that's what I do here.

In the case of exponentiation, it doesn't seem to be valid to have a tensor exponent. I revised the implementation to use a `torch_scalar_t` type and restricted attention to the integer case for now.

### DIFF
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -277,22 +277,24 @@ torch_tensor_t torch_tensor_multiply(const torch_tensor_t tensor1,
   return output;
 }
 
-torch_tensor_t torch_tensor_premultiply(const torch_data_t scalar,
+// FIXME: Don't cast to float. Now doesn't work for integers
+torch_tensor_t torch_tensor_premultiply(const float scalar,
                                         const torch_tensor_t tensor) {
   auto t = reinterpret_cast<torch::Tensor *const>(tensor);
   torch::Tensor *output = nullptr;
   output = new torch::Tensor;
-  std::cout << "DEBUG: scalar = " << scalar << std::endl;
+  std::cout << "DEBUG: pre-multiplier = " << scalar << std::endl;
   *output = scalar * *t;
   return output;
 }
 
+// FIXME: Don't cast to float. Now doesn't work for integers
 torch_tensor_t torch_tensor_postmultiply(const torch_tensor_t tensor,
-                                         const torch_data_t scalar) {
+                                         const float scalar) {
   auto t = reinterpret_cast<torch::Tensor *const>(tensor);
   torch::Tensor *output = nullptr;
   output = new torch::Tensor;
-  std::cout << "DEBUG: scalar = " << scalar << std::endl;
+  std::cout << "DEBUG: post-multiplier = " << scalar << std::endl;
   *output = *t * scalar;
   return output;
 }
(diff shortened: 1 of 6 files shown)


jwallwork23:
Following our discussion in the meeting, I've now extended the power functionality to account for both integers and floats, replacing `torch_scalar_t` with `torch_int_t` and `torch_float_t`. I covered both cases with the square unit test and also added a square root one.

jwallwork23:
[Rebased on top of `main` to pick up the restructuring changes]

jatkinson1000 (approved):
Hi @jwallwork23 This looks great.
Am I correct that you've restructured the C++ to match the re-organising you did in the Fortran?

One observation (that is not to be resolved here) is that our [online API docs](https://cambridge-iccs.github.io/FTorch/lists/procedures.html) are no longer very useful for basic users and are instead geared at developers.
In reality, all basic users want is the `Interface` docs, and the list they get above is somewhat daunting. There is an option to lost interfaces [here](https://cambridge-iccs.github.io/FTorch/module/ftorch.html) but I don't feel it helps. I'm not sure if there is an easy solution to this and should perhaps create a new issue.

jwallwork23:
> Am I correct that you've restructured the C++ to match the re-organising you did in the Fortran?

Oh, I didn't do that actually. Can do in a follow-up PR, though.

jwallwork23:
> One observation (that is not to be resolved here) is that our [online API docs](https://cambridge-iccs.github.io/FTorch/lists/procedures.html) are no longer very useful for basic users and are instead geared at developers. In reality, all basic users want is the `Interface` docs, and the list they get above is somewhat daunting. There is an option to lost interfaces [here](https://cambridge-iccs.github.io/FTorch/module/ftorch.html) but I don't feel it helps. I'm not sure if there is an easy solution to this and should perhaps create a new issue.

That's a good point. I guess this is a downside of the fypp approach?

jwallwork23:
> One observation (that is not to be resolved here) is that our [online API docs](https://cambridge-iccs.github.io/FTorch/lists/procedures.html) are no longer very useful for basic users and are instead geared at developers. In reality, all basic users want is the `Interface` docs, and the list they get above is somewhat daunting. There is an option to lost interfaces [here](https://cambridge-iccs.github.io/FTorch/module/ftorch.html) but I don't feel it helps. I'm not sure if there is an easy solution to this and should perhaps create a new issue.

Opened #232.

PR #228, opened 2025-01-10
Set upper CMake minimum version in CMakeLists across codebase.

### LINKED ISSUE #226: Set CMake minimum to a range
As raised by @matthewfeickert in #214 the better syntax to use when setting `cmake_minimum_required` is to specify both a minimum and maximum.

This is because `cmake_minimum_required` makes a call to `cmake_policy` which will enforce ONLY the policies of the minimum if no maximum is supplied.

It is preferable to provide both. c.f https://cliutils.gitlab.io/modern-cmake/chapters/basics.html for more info.

A downside of this is that all CMake files will need updating whenever a new version is released as this is a pinning feature and I don't see an option to provide a `latest` as the maximum.

jatkinson1000:
This should close #226 by setting an upper bound on the minimum required version to the most recent version of CMake (`3.31`).

See issue #226 and https://cliutils.gitlab.io/modern-cmake/chapters/basics.html#minimum-version for more explanation on why this is advised.

### DIFF
--- a/examples/1_SimpleNet/CMakeLists.txt
+++ b/examples/1_SimpleNet/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.15...3.31 FATAL_ERROR)
 # policy CMP0076 - target_sources source files are relative to file where
 # target_sources is run
 cmake_policy(SET CMP0076 NEW)
--- a/examples/n_c_and_cpp/CMakeLists.txt
+++ b/examples/n_c_and_cpp/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.15...3.31 FATAL_ERROR)
 set(PROJECT_NAME FTorch)
 set(LIB_NAME ftorch)
 set(PACKAGE_VERSION 0.1)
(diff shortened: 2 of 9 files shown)


examples/n_c_and_cpp/CMakeLists.txt
jwallwork23, line 1: Not for this PR but I wonder if we should remove this demo from the `main` branch and open an issue that would finish it? It's neither complete nor tested so perhaps not suited in v1.0.

jwallwork23 (approved):
Makes sense, thanks @jatkinson1000.

examples/n_c_and_cpp/CMakeLists.txt
jatkinson1000: Agreed.

examples/1_SimpleNet/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.15...3.31 FATAL_ERROR)
```
matthewfeickert, line 1: This isn't a problem by any means, but `FATAL_ERROR` is ignored in CMake `v2.6+`.

From the [docs](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html):

> The `FATAL_ERROR` option is accepted but ignored by CMake 2.6 and higher. It should be specified so CMake versions 2.4 and lower fail with an error instead of just a warning.

and as you're specifying `v3.15+` here it will always be ignored. I'd suggest removing it just so that people don't think it is doing something to help them. Though again, it won't cause you any errors if you chose to leave it.

jwallwork23: Thanks for this @matthewfeickert. I've opened #233 to remove these.

Issue #229, opened 2025-01-13
Remove `n_c_and_cpp` demo until finished


jwallwork23:
Not for this PR but I wonder if we should remove this demo from the `main` branch and open an issue that would finish it? It's neither complete nor tested so perhaps not suited in v1.0.

_Originally posted by @jwallwork23 in https://github.com/Cambridge-ICCS/FTorch/pull/228#discussion_r1910340522_

jatkinson1000:
@jwallwork23 Should this be closed?
Looks like it didn't get tied to #230

jwallwork23:
Ah, I guess it isn't very clear. I meant for this issue to be the one for reintroducing that demo. I'll give it a better title.

jwallwork23:
Updated the description, too.

PR #230, opened 2025-01-13
Drop n_c_and_cpp example for now

jwallwork23:
As agreed in [this thread](https://github.com/Cambridge-ICCS/FTorch/pull/228#discussion_r1910340522).

Opened #229 as a reminder to finish it.


jatkinson1000 (approved):
Thanks @jwallwork23 This looks fine to me.

Done a quick sanity check on the docs and I can't see any reference to it in there.

We can always recover this if we want a C example in future.

Issue #211, opened 2024-12-23
Move autograd/operator overloading tests checks into unit test


jwallwork23:
Cut down the number of checks in `examples/6_Autograd/autograd.f90` into a new unit test once #76 is merged.

PR #231, opened 2025-01-13
Move input array checks out of autograd example

jwallwork23:
Towards #211.
Follows #225.

This PR cuts down the amount of testing done in the autograd example. In particular, it moves the checks that input arrays aren't modified into the unit tests.

The PR also adds a missing summary for the autograd example in `examples/README.md`.


jwallwork23:
Just pushed a couple of commits because I realised that the cleanup subroutines weren't being used everywhere. Further, the shape checks were redundant because shape checking is already done in the `assert_allclose` subroutines.

jatkinson1000 (approved):
Thanks @jwallwork23 
Been through this and the key changes make sense.
I haven't been through every change in the test_tensor_operator_overloads, but will trust that they are all OK since passing.

PR #233, opened 2025-01-13
Minor build tweaks

jwallwork23:
This PR:

* Gives some information on additional build dependencies in the README. In particular that `torchvision` should be installed at the same time as `torch`.
* Drop `FATAL_ERROR` from CMake versioning, which is ignored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,8 @@ FTorch's test suite has some additional dependencies.
   [pFUnit](https://github.com/Goddard-Fortran-Ecosystem/pFUnit).
 * FTorch's test suite requires that [PyTorch](https://pytorch.org/) has been
   installed, as opposed to LibTorch. We recommend installing `torchvision` in
-  the same command (e.g., `pip install torch torchvision`)<sup>*</sup>.
+  the same command (e.g., `pip install torch torchvision`)<sup>*</sup>. Doing so
+  ensures that `torch` and `torchvision` are configured in the same way.
 * Other Python modules are installed automatically by the `run_test_suite.sh`
   test script (or `run_test_suite.bat` on Windows).
 
--- a/README.md
+++ b/README.md
@@ -88,6 +88,21 @@ To install the library requires the following to be installed on the system:
 
 <sup>*</sup> _The minimal example provided downloads the CPU-only Linux Nightly binary. [Alternative versions](https://pytorch.org/get-started/locally/) may be required._
 
+#### Additional dependencies of the test suite
+
+FTorch's test suite has some additional dependencies.
+
+* You will also need to install the unit testing framework
+  [pFUnit](https://github.com/Goddard-Fortran-Ecosystem/pFUnit).
+* FTorch's test suite requires that [PyTorch](https://pytorch.org/) has been
+  installed, as opposed to LibTorch. We recommend installing `torchvision` in
+  the same command (e.g., `pip install torch torchvision`)<sup>*</sup>.
+* Other Python modules are installed automatically by the `run_test_suite.sh`
+  test script (or `run_test_suite.bat` on Windows).
+
+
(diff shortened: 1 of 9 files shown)


ElliottKasoar:
Do you think it might be worth bumping the minimum CMake version to 3.18, as they have in [PyTorch](https://github.com/pytorch/pytorch/blob/main/torch/CMakeLists.txt)?

I think 3.15 still works in most cases, but [CMake 3.18 was the first version to include [C++17] support for any compiler](https://cmake.org/cmake/help/latest/prop_tgt/CUDA_STANDARD.html) (and includes potentially useful things like [CMAKE_CUDA_ARCHITECTURES](https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_ARCHITECTURES.html)), so it may be safer, although C++17 is already in the FTorch requirements separately.

jatkinson1000:
PyTorch generally suggests installing torch, torchvision, and torchaudio as a oneshot.
Whilst looking at the conda side of things I feel we do not need torchaudio and torchvision is only a dependency for resnet in example 2, though should check this to be absolutely sure.

I'm still on the fence about `FATAL ERROR` and whether, given it has no effect beyond CMake >= 2.6, it should be removed.
The docs state that it "should" be included in case anyone tries building with a version lower than 2.6 (unlikely but a possibility). My reading of [the CMake docs](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html), though I have not tried doing this, is that if someone tried to build FTorch using an older version of CMake it would not raise an error about the version and instead fail elsewhere in the script.

matthewfeickert:
> The docs state that it "should" be included in case anyone tries building with a version lower than 2.6 (unlikely but a possibility).

No, it is not supported at all by

https://github.com/Cambridge-ICCS/FTorch/blob/d5046a3688115b41e0051430f08035ba665942d4/src/CMakeLists.txt#L1

The `3.15` is not a suggestion. It is a hard requirement. You can not build the project with a lower policy number and the policy number is the CMake version.

henryiii:
I haven't seen CMake <2.8 in _many_ years. CMake 2.6 came out in 2008. I would not worry about what happens if someone tries to run with 2.6; nothing else will work either, it likely wouldn't even run on modern systems.

matthewfeickert:
(I deleted my previous message as I now understand you were concerned about the effect of CMake `v2.x` which would print message but keep trying to build, ignoring the `cmake_minimum_required` minimum version, instead of failing out before the build. Sorry for misunderstanding.)

Though I agree with Henry that I wouldn't worry about people trying to build with CMake `v2.X`. Do you know of people you work with that are still using CMake `v2`?

jatkinson1000 (approved):
Thanks Both, and @jwallwork23 
In that case I'll trust your judgment, and I also did a quick scan of linux distros and their requirements and concluded anyone using anything old enough to come with CMake 2.x probably has bigger concerns!!

So I'm happy to approve this for merge.

jwallwork23:
> Do you think it might be worth bumping the minimum CMake version to 3.18, as they have in [PyTorch](https://github.com/pytorch/pytorch/blob/main/torch/CMakeLists.txt)?
> 
> I think 3.15 still works in most cases, but [CMake 3.18 was the first version to include support for any compiler](https://cmake.org/cmake/help/latest/prop_tgt/CUDA_STANDARD.html), enabling `CMAKE_CUDA_STANDARD` 17 support (it also includes potentially useful things like [CMAKE_CUDA_ARCHITECTURES](https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_ARCHITECTURES.html)), so it may be safer, although C++17 is already in the FTorch requirements separately.

@ElliottKasoar Thanks for the suggestion. I opened this as a new issue: #234.

PR #235, opened 2025-01-14
Reorganise C/C++ as in Fortran

jwallwork23:
Follows #32.

Applies similar section headings in the C header and C++ source as were introduced in the Fortran source. Involves some reordering of appearance so these line up.


jatkinson1000 (approved):
All reshuffling, no source changes.
Tests are happy so I'm happy.

Issue #236, opened 2025-01-14
Use fixtures in unit tests


jwallwork23:
Currently, the unit tests only use single values for the different input parameters. We should introduce fixtures so that we can cover more cases without duplicating the test code.

PR #237, opened 2025-01-14
Ensure strides array initialised to zero

jwallwork23:
While working on #236, I hit an overflow error in the 3D `torch_tensor_from_array` unit test. Digging into it, I noticed that the `strides` array in `torch_tensor_from_blob` is not initialised to zero!

I also switched `tensor_shape` and `layout` to be assumed shape as I don't think assumed size is necessary.

### DIFF
--- a/src/ftorch.F90
+++ b/src/ftorch.F90
@@ -253,7 +253,7 @@ contains
     use, intrinsic :: iso_c_binding, only : c_bool, c_int, c_int64_t
     type(torch_tensor), intent(out) :: tensor     !! Returned tensor
     integer(c_int), intent(in)      :: ndims      !! Number of dimensions of the tensor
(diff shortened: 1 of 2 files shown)


jwallwork23:
@jatkinson1000 and I had a discussion with @SimonClifford and determined that the same assumed shape change should be made else where - see https://github.com/Cambridge-ICCS/FTorch/pull/237/commits/190782aef685cf7c3e0b3f1161432e9e3a8e2d13. It still needs to be assumed size in the C interfaces because assumed shape is not interoperable - see https://stackoverflow.com/questions/38349435/in-fortran2003-is-1d-assumed-shape-array-interoperable-with-c.

jatkinson1000 (approved):
Discussed in-person and happy with this, thanks @jwallwork23

PR #238, opened 2025-01-14
Make use of pFUnit test fixtures

### LINKED ISSUE #236: Use fixtures in unit tests

jwallwork23:
Closes #236.
To follow #231 and #237 (rebased on top of those branches).

Sets up the pFUnit fixture machinery for the `test_tensor_constructors` unit tests.

To do this, I needed to convert the unit tests to use modules. A benefit of doing so is that common imports and parameters could be moved to the module level. (As such, many of the lines changed are just indentations of the unit tests.)


jwallwork23:
I recommend viewing the changes in this file offline with `git diff --word-diff`. Many of the lines changed are just indentations when the modules were introduced for the tests.

jatkinson1000 (approved):
Yes, this simplifies things a little in-module, and makes it easier to use.

The diff is a little hard to read (lots of changes and moved to a new file) but they seem sensible on looking through and... the tests are still passing...

PR #239, opened 2025-01-14
Add unit tests for tensor interrogation

### LINKED ISSUE #211: Move autograd/operator overloading tests checks into unit test

### LINKED ISSUE #219: Unit tests for procedures that interrogate tensors
- [x] `torch_tensor_get_device_index`
- [x] `get_rank`
- [x] `get_shape`
- [ ] `torch_tensor_to_array`

Also CUDA device unit tests (built only if `ENABLE_CUDA` is true) for
- [x] `torch_tensor_get_device_index`

Tests for other (and multiple) devices to be added later.

jwallwork23:
Closes #219.
Closes #211.

This PR adds unit tests for the `get_rank` and `get_shape` methods of `torch_tensor`, as well as for the `torch_tensor_get_device_index` util on both CPU and GPU.

I decided not to add a unit test for `torch_tensor_to_array` because it's already covered elsewhere (e.g., in the constructors tests).


.github/workflows/test_suite_ubuntu.yml
```diff
@@ -99,7 +99,12 @@ jobs:
...
+      - name: Run unit tests
         run: |
           . ftorch/bin/activate
-          ./run_test_suite.sh -V
+          ./run_test_suite.sh -V -u
+
+      - name: Run integration tests
+        run: |
+          . ftorch/bin/activate
+          ./run_test_suite.sh -V -i
```
jatkinson1000, line 110: Some advice I had from @christopheredsall always use the verbose flags in bash scripts or CI jobs as it is easy in future to see what it is doing, rather than looking up further docs, and reduces risk of wrong flags creeping in.

```suggestion
          ./run_test_suite.sh --verbose --unit-only

      - name: Run integration tests
        run: |
          . ftorch/bin/activate
          ./run_test_suite.sh --verbose --integration-only
```

jatkinson1000 (commented):
Thanks @jwallwork23 that's a nice functionality for future!

One small comment about using verbose flags and then I think you're good to go with some docs!

jwallwork23:
> One small comment about using verbose flags and then I think you're good to go with some docs!

Hopefully the doc changes in 7d98d09ad19f9b1b97967e85a5e1d614735e2b08 read well.

jatkinson1000 (approved):
Thanks @jwallwork23 This looks good now.

When I said docs for the tests I was actually thinking of documenting somewhere what the different unit tests were and how they are broken up and use pFUnit.
However, that is perhaps a bigger job than here ao I will create a separate issue for it.

PR #245, opened 2025-01-20
Fix broken URL in README

dorchard:
Fixes the link to the developer docs which was broken. Points to the build static HTML in this fix.


jatkinson1000 (approved):
Thanks @dorchard

PR #247, opened 2025-01-20
Implement `torch_tensor_minus` for taking negative of a tensor

### LINKED ISSUE #246: Overload minus operator for tensors
While working on implementing #152, I realised that we haven't overloaded the unary minus operator, i.e., taking the negative of a tensor.

jwallwork23:
Closes #246.

Implements and overload for the unary `-` operator: `torch_tensor_minus`, plus unit test.

### DIFF
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -268,6 +268,14 @@ torch_tensor_t torch_tensor_add(const torch_tensor_t tensor1,
   return output;
 }
 
+torch_tensor_t torch_tensor_minus(const torch_tensor_t tensor) {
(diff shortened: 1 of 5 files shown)


jatkinson1000 (commented):
Code wise, this all looks good, but I wonder if '`minus`' could be misleading, and whether '`negative`' might be better?

Plus extending the comments to be a bit more explicit e.g. 'unary minus' as you used when we were discussing.

dorchard:
I concur with `negative` (this naming is used also in NumPy for example).

jatkinson1000 (approved):
Thanks, I'm happy with this.

PR #251, opened 2025-01-24
Add `get_dtype` and `get_device_type` methods for `torch_tensor`

### LINKED ISSUE #248: Implement `get_dtype` and `get_device_type` for tensors
These, in combination with the existing `get_rank`, `get_shape`, and `get_device_index`, will allow us to create one tensor from another using the same setup.

This will be useful for autograd because the default is to compute the gradient w.r.t. a tensor of ones of the appropriate shape.

We should also make `get_device_index` a method of `torch_tensor`, as with `get_rank` and `get_shape`.

jwallwork23:
Closes #248.

This PR adds functions for getting the data type and device type of a tensor, including unit tests. It also improves the consistency of the existing functions so that they are called `torch_tensor_get_X` but are mapped to methods of the `torch_tensor` class as just `get_X`. This makes it clearer what we are getting the rank/shape of when we call as functions, and reduces unnecessary verbosity when calling as methods.

Using the new utility for getting the device type, the operator overloads involving scalars are corrected so that CPU isn't assumed.

### DIFF
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -40,7 +40,8 @@ constexpr auto get_dtype(torch_data_t dtype) {
   }
 }
 
-const auto get_device(torch_device_t device_type, int device_index) {
+// Mapping from FTorch device_type_t to libtorch DeviceType
+const auto get_libtorch_device(torch_device_t device_type, int device_index) {
   switch (device_type) {
   case torch_kCPU:
     if (device_index != -1) {
@@ -65,6 +66,20 @@ const auto get_device(torch_device_t device_type, int device_index) {
   }
 }
 
+// Mapping from libtorch DeviceType to FTorch device_type_t
+const torch_device_t get_ftorch_device(torch::DeviceType device_type) {
+  switch (device_type) {
+  case torch::kCPU:
+    return torch_kCPU;
+  case torch::kCUDA:
+    return torch_kCUDA;
+  default:
+    std::cerr << "[ERROR]: device type " << device_type << " not implemented in FTorch"
+              << std::endl;
+    exit(EXIT_FAILURE);
+  }
+}
+
 // =============================================================================
 // --- Functions for constructing tensors
 // =============================================================================
@@ -79,7 +94,7 @@ torch_tensor_t torch_empty(int ndim, const int64_t *shape, torch_data_t dtype,
     c10::IntArrayRef vshape(shape, ndim);
     tensor = new torch::Tensor;
     *tensor = torch::empty(vshape, torch::dtype(get_dtype(dtype)))
-                  .to(get_device(device_type, device_index));
+                  .to(get_libtorch_device(device_type, device_index));
   } catch (const torch::Error &e) {
     std::cerr << "[ERROR]: " << e.msg() << std::endl;
     delete tensor;
@@ -102,7 +117,7 @@ torch_tensor_t torch_zeros(int ndim, const int64_t *shape, torch_data_t dtype,
     c10::IntArrayRef vshape(shape, ndim);
     tensor = new torch::Tensor;
     *tensor = torch::zeros(vshape, torch::dtype(get_dtype(dtype)))
-                  .to(get_device(device_type, device_index));
+                  .to(get_libtorch_device(device_type, device_index));
   } catch (const torch::Error &e) {
     std::cerr << "[ERROR]: " << e.msg() << std::endl;
     delete tensor;
@@ -125,7 +140,7 @@ torch_tensor_t torch_ones(int ndim, const int64_t *shape, torch_data_t dtype,
     c10::IntArrayRef vshape(shape, ndim);
     tensor = new torch::Tensor;
     *tensor = torch::ones(vshape, torch::dtype(get_dtype(dtype)))
-                  .to(get_device(device_type, device_index));
+                  .to(get_libtorch_device(device_type, device_index));
   } catch (const torch::Error &e) {
     std::cerr << "[ERROR]: " << e.msg() << std::endl;
     delete tensor;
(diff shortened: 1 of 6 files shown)


src/ftorch.F90
```diff
@@ -27,10 +27,15 @@ module ftorch
 
   !> Type for holding a Torch tensor.
   type torch_tensor
-    type(c_ptr) :: p = c_null_ptr  !! pointer to the tensor in memory
+    type(c_ptr) :: p = c_null_ptr    !! pointer to the tensor in memory
+    integer, private :: dtype        !! integer representing the tensor's data type
+    integer, private :: device_type  !! integer representing the tensor's device type
```
jwallwork23, line 32: These are outdated and should be dropped.

jwallwork23: Done in 7e52efe.

src/ctorch.cpp
```diff
@@ -40,7 +41,37 @@ constexpr auto get_dtype(torch_data_t dtype) {
   }
 }
 
-const auto get_device(torch_device_t device_type, int device_index) {
+// Mapping from libtorch Dtype to FTorch device_data_t
+torch_data_t get_ftorch_dtype(caffe2::TypeMeta dtype) {
+  if (dtype == torch::kUInt8) {
```
jatkinson1000, line 46: Any reason to make this if-else rather than the switch-case used by the other functions?

src/test/unit/test_tensor_interrogation_cuda.pf
jatkinson1000, line 1: Do you think it would be worth adding a test for the other functions running on a CUDA device (get rank etc)?

On thinking about this, if they do not map over from CPU to other devices then that is likely an issue with the backends, rather than our code, so would only be serving as a warning that there were problems with the underlying dependencies.

jatkinson1000 (approved):
This all looks good code-wise @jwallwork23 
You are right that it was a tricky review!!

My overarching thought is perhaps we need an example for basic tensor manipulation.
I think that would be a separate task from this PR however. Thoughts?

Now that you explicitly set the device to create the tensor in some overloads I have a question.
What happens if we call this with tensors that are on different devices?
I presume it fails with a meaningful error message from the C++, but does it provide a useful traceback to where the error originated in the Fortran? I recall sometimes libtorch gives an error report, but no code location making it hard to work out where your Fortran is going wrong.

src/test/unit/test_tensor_interrogation_cuda.pf
jwallwork23: Such tests would be duplicating those in the CPU tests and then we'd need to do the same for XPU, etc. I think we can lean on the underlying implementation for this, but can add tests if preferred.

src/ctorch.cpp
jwallwork23: I think it didn't like switching on the `TypeMeta` type but I will check this.

src/test/unit/test_tensor_interrogation_cuda.pf
jatkinson1000: Nah, I think I agree, and am keen to avoid the verbosity until proven otherwise.
Just wondered what your thoughts were. :)

jwallwork23:
> You are right that it was a tricky review!!

Apologies!

> My overarching thought is perhaps we need an example for basic tensor manipulation. I think that would be a separate task from this PR however. Thoughts?

That's a good idea. I'm increasingly thinking we need to rethink the ordering of the examples. (See https://github.com/Cambridge-ICCS/FTorch/pull/258#issuecomment-2616553746.) Such an example should be near the start.

> Now that you explicitly set the device to create the tensor in some overloads I have a question. What happens if we call this with tensors that are on different devices? I presume it fails with a meaningful error message from the C++, but does it provide a useful traceback to where the error originated in the Fortran? I recall sometimes libtorch gives an error report, but no code location making it hard to work out where your Fortran is going wrong.

Will check, thanks for pointing out this case.

jatkinson1000:
Cool, opened #261 
Happy for this to be merged after you check the "different device" query.

src/ctorch.cpp
jwallwork23: Yeah it appears `switch` statements need to switch on integers
```
/home/joe/software/FTorch/src/ctorch.cpp: In function ‘torch_data_t get_ftorch_dtype(caffe2::TypeMeta)’:
/home/joe/software/FTorch/src/ctorch.cpp:46:11: error: switch quantity not an integer
   46 |   switch (dtype) {
      |           ^~~~~

```

jwallwork23:
> Now that you explicitly set the device to create the tensor in some overloads I have a question. What happens if we call this with tensors that are on different devices? I presume it fails with a meaningful error message from the C++, but does it provide a useful traceback to where the error originated in the Fortran? I recall sometimes libtorch gives an error report, but no code location making it hard to work out where your Fortran is going wrong.

@jatkinson1000 hm the error isn't so helpful. Making the modifications in the last commit on [248_get-dtype-devicetype_GPU-test](https://github.com/Cambridge-ICCS/FTorch/tree/248_get-dtype-devicetype_GPU-test) - which attempts to assign a tensor on a CUDA device to a tensor on the CPU - I get the output
```
4: Test command: /home/joewa/software/FTorch/src/build/test/examples/3_MultiGPU/multigpu_infer_fortran "/home/joewa/software/FTorch/src/build/test/examples/3_MultiGPU/saved_multigpu_model_cuda.pt"
4: Working Directory: /home/joewa/software/FTorch/src/build/test/examples/3_MultiGPU
4: Test timeout computed to be: 1500
4: input on rank 0: [  0.0,  1.0,  2.0,  3.0,  4.0]
4: output on rank 0: [*****,  0.0,  0.0,  0.0,*****]
4:  MultiGPU example ran successfully
4/4 Test #4: multigpu_infer_fortran ...........   Passed    8.12 sec
```
That is, it doesn't raise an error at all. So I guess we should build in errors for when you try to apply operator overloads to tensors on different devices.

jwallwork23:
Opened #269. Will merge and follow up there.

PR #254, opened 2025-01-24
Require `device_type` argument for `torch_model_load`

### LINKED ISSUE #252: Make device type required for model load
As per description discussed at hackathon.

Currently it is optional with a default fallback.

jwallwork23:
Closes #252.

The main bug/oversight we hit in the hackathon.


ma595 (approved):

jatkinson1000 (changes requested):
@jwallwork23 Code in examples 5 and 6 will need updates, as will any example snippets in the docs/readmes.

This is also a breaking API change, though hopefully a straightforward enough one that people should be able to figure it out.
I'm not sure we need heavy signposting, but perhaps add something to (or update the model load section of) https://cambridge-iccs.github.io/FTorch/page/updates.html

ma595:
@jatkinson1000 the autograd example has no `torch_model_load` fn, but I think you just noticed this :)

jwallwork23:
> @jwallwork23 Code in examples 5 ~and 6~ will need updates, as will any example snippets in the docs/readmes.

In addition to @ma595's change, I propose building example 5 if `CMAKE_BUILD_TESTS=TRUE`, even though we don't run the example in the integration tests. This would help to catch such incorrect API. (See f5f05cdbabcf8b2e1102c30191219af8757d4320.)
 
> This is also a breaking API change, though hopefully a straightforward enough one that people should be able to figure it out. I'm not sure we need heavy signposting, but perhaps add something to (or update the model load section of) https://cambridge-iccs.github.io/FTorch/page/updates.html

Good point. Done in cbab392b0105633937704d05642f5811de5fbb46, b2c6c74b869ef1682f0099d3fb0e476942d1a019, fc864f44c0dfca31c87a76603ca9d84a6f00d1f1.

jwallwork23:
> One small addition that I opened #256 for if you want to merge that, then I think I am happy.

Thanks, merged.

> Building 5 even if it is not run is a good idea to catch things like this. I wonder if it is worth noting somewhere that it will not be run? - not a requirement though.

I've highlighted this in https://github.com/Cambridge-ICCS/FTorch/pull/254/commits/a2e49e9d36afd56f35089e9df051db878b69161d.

PR #256, opened 2025-01-27
Update example code in readme to specify device when loading model.

jatkinson1000:
Small addition.
Opening as a PR because you are active on the target branch @jwallwork23.


jwallwork23 (approved):
Good spot, thanks!

PR #260, opened 2025-01-27
Avoid double negatives in the run tests shell script to make things clearer

### LINKED ISSUE #259: Avoid double negatives in `run_test_suite.sh` command line options
It would be good to avoid "double negatives" - the guard for running _integration_ tests is based on the _unit_ test flag. This requires some mental gymnastics (unless it's just me) and would be a problem if we wanted to add a third set of subtests for some reason in future.

I wonder if moving to having `RUN_UNIT` and `RUN_INTEGRATION` variables that default to `true` and are checked against for running the tests, and setting them to `false` where necessary in setup would be preferable?

_Originally posted by @jatkinson1000 in https://github.com/Cambridge-ICCS/FTorch/pull/222#pullrequestreview-2575238258_

jatkinson1000:
Should resolve #259

Moves `X_ONLY` to `RUN_X` and now checks:
```sh
if [ "${RUN_X}" = true ]; then
  cd "${BUILD_DIR}/test/x"
  ctest "${CTEST_ARGS}"
  cd -
fi 
```

### DIFF
--- a/run_test_suite.sh
+++ b/run_test_suite.sh
@@ -27,8 +27,8 @@ show_help() {
 
 # Parse command line arguments
 BUILD_DIR="src/build"
-INTEGRATION_ONLY=false
-UNIT_ONLY=false
+RUN_INTEGRATION=true
+RUN_UNIT=true
 VERBOSE=false
 HELP=false
 for ARG in "$@"; do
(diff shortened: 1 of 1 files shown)


jwallwork23:
I think you might want to do an interactive rebase, dropping commits from #222.

jatkinson1000:
I was envisaging it going in after #222, but actually you are correct, could go in now at the expense of you perhaps needing to update 222.

jwallwork23 (approved):
Okay cool, thanks!

PR #263, opened 2025-01-29
Bugfix: mismatch in Fortran array character array lengths in example 2

jatkinson1000:
A mismatch in character array lengths was causing the wrong resnet category to be printed in the Fortran.

At some point lengths in the main routine were updated to 128, but not in the subroutine required to read categories leading to the arrays becoming garbled during the passing.

Also updates the filename strings to 128 for consistency, rather than seemingly arbitrary lengths.


jwallwork23 (approved):
Good catch!

PR #266, opened 2025-01-30
Enforce the use of a virtual environment in the integration testing script.

### LINKED ISSUE #265: Test suite pip installs into current environment
The test script installs the dependencies for all integration tests via pip:

https://github.com/Cambridge-ICCS/FTorch/blob/c21040df1436e23c2cb12fe075cd512f0e6e7495/run_test_suite.sh#L83-L96

but they go into the current environment.

This was fine on GitHub workflows where we reactivate a venv, but we should probably suggest users make sure that they are running it from within a venv to avoid environment pollution.

jatkinson1000:
Closes #265 

Enforces use of a venv through setting `PIP_REQUIRE_VIRTUALENV=true` in the unix tests script.

Running without a virtual environment will give:

```console
./run_test_suite.sh --verbose --integration-only
ERROR: Could not find an activated virtualenv (required).
Error: Process completed with exit code 3.
```

### DIFF
--- a/pages/testing.md
+++ b/pages/testing.md
@@ -39,10 +39,10 @@ See the subsections below for instructions on how to run subsets of the full
 test suite.
 
 > <sup>1</sup> _If you built FTorch against LibTorch (rather than creating a
-virtual environment) then either
+virtual environment) then you will need to
 [create a virtual environment](https://docs.python.org/3/library/venv.html) for
-the purposes of testing, or note that this script may have your Python
-environment install some modules._
+the purposes of integration testing as this script will install packages into your
+Python environment and will check that a virtual environment is in use._
 
 #### Running unit tests on Unix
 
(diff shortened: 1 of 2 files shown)


matthewfeickert:
> I can confirm that this does the trick, although I get a different output to what you mention in the PR because the system `python` first in my path doesn't have pip installed:
> 
> ```
> $ ./run_test_suite.sh --verbose --integration-only
> /usr/local/bin/python: No module named pip
> ```
> 
> If I switch to `python3 -m pip -q install ...` in the script then I get the expected error.

The relevant bit is that your not in a virtual environment and there is no need to specify `python3`. The error that you're getting is a different unrelated one and if you were in a virtual environment `python` would be `python3` already.

You're simply hitting the fact that you're trying to use the system Python for your own personal use, which you should never do. The system Python is its Python and you should act like it doesn't exist to you. If you want to use Python as a developer or user you should install your own.

PR #272, opened 2025-01-30
Checks for inconsistent device types

### LINKED ISSUE #269: Raise error if overloaded operator applied to tensors on different devices
> Now that you explicitly set the device to create the tensor in some overloads I have a question. What happens if we call this with tensors that are on different devices? I presume it fails with a meaningful error message from the C++, but does it provide a useful traceback to where the error originated in the Fortran? I recall sometimes libtorch gives an error report, but no code location making it hard to work out where your Fortran is going wrong.

Making the modifications in the last commit on [248_get-dtype-devicetype_GPU-test](https://github.com/Cambridge-ICCS/FTorch/tree/248_get-dtype-devicetype_GPU-test) - which attempts to assign a tensor on a CUDA device to a tensor on the CPU - I get the output
```
4: Test command: /home/joewa/software/FTorch/src/build/test/examples/3_MultiGPU/multigpu_infer_fortran "/home/joewa/software/FTorch/src/build/test/examples/3_MultiGPU/saved_multigpu_model_cuda.pt"
4: Working Directory: /home/joewa/software/FTorch/src/build/test/examples/3_MultiGPU
4: Test timeout computed to be: 1500
4: input on rank 0: [  0.0,  1.0,  2.0,  3.0,  4.0]
4: output on rank 0: [*****,  0.0,  0.0,  0.0,*****]
4:  MultiGPU example ran successfully
4/4 Test #4: multigpu_infer_fortran ...........   Passed    8.12 sec
```
That is, it doesn't raise an error at all. So I guess we should build in errors for when you try to apply operator overloads to tensors on different devices.

_Originally posted by @jwallwork23 in https://github.com/Cambridge-ICCS/FTorch/issues/251#issuecomment-2624308175_

jwallwork23:
Closes #269.

This PR adds checks that overloaded operators are only applied to tensors that have the same device type.

It's sufficient to add the checks in the assignment operator and the binary operators as the unary operators will go through the assignment anyway.

The PR also adds checks that tensors have been constructed before they are interrogated. Related fix in the autograd example - the output tensor `Q` wasn't set up so didn't have a device type yet.

## Questions:
* Do we also need to check the device indices are consistent?
  * _-> future PR_
* And what about data types?
  * _-> future PR_
* Can we add pFUnit unit tests that check these errors are raised as expected?
  * _->Doesn't seem so (see below)._


jatkinson1000 (approved):
Hey @jwallwork23 I just rebased this off of [current main](https://github.com/Cambridge-ICCS/FTorch/commit/120f1e5ed3bad195b369a19bb7a134f31c731c2c) and then modified an example to check that if we try and add two with different devices we get an error raised as expected.

I used the below:

```fortran
real(wp), dimension(5), target :: a, b, combine_data
type(torch_tensor) :: tensor_a, tensor_b, combine_tensor

a = [0.0_wp, 1.0_wp, 2.0_wp, 3.0_wp, 4.0_wp]
b = [0.0_wp, 1.0_wp, 2.0_wp, 3.0_wp, 4.0_wp]

call torch_tensor_from_array(tensor_a, a, tensor_layout, torch_kCPU)
call torch_tensor_from_array(tensor_b, b, tensor_layout, torch_kMPS)

combine_tensor = tensor_a + tensor_b
call torch_tensor_print(combine_tensor)
```

When tensors a and b are on CPU this passes fine, but in the above state (b on MPS) we get:

```console
 Error :: cannot add tensors with different device types
Note: The following floating-point exceptions are signalling: IEEE_OVERFLOW_FLAG
STOP 1
```

Since we looked together in-person I am happy to approve this now.

For posterity we also discussed that this error warning could be made friendlier (parent routine, line number etc) but this is not a simple task and not to be done here.

jwallwork23:
>     * Can we add pFUnit unit tests that check these errors are raised as expected?

I don't think it currently supports this - as far as I can tell, it can only mark certain tests as expected to fail, not interrogate error messages.

jwallwork23:
[Rebased on top of main, fixing various conflicts]

jatkinson1000:
I think I hadn't re-approved this because of the issues I had seen when trying to print a tensor.
See where I was investigating on [`mix_devices_temp`](https://github.com/Cambridge-ICCS/FTorch/tree/mix_devices_temp) branch.

Did we get to be bottom of that? I know it was mentioned in the meeting this morning.

PR #264, opened 2025-01-29
Add conda environment files and instructions

jatkinson1000:
This should hopefully address some of  #214 

- [x] Add conda environment for CPU-only builds
- [x] Add conda environment for CUDA build
- [x] Add instructions for conda users
- [x] Independent test build by someone else

Queries:

- What is best practice, should we be pinning versions to any extent? Currently I just name the packages
- Should we add a workflow to check anything? Seems like a lot more effort to do properly (without pip) and I do say in the docs this is not our preferred or maintained approach...

PR #275, opened 2025-02-04
Add multiproc build note

jatkinson1000:
As suggested in #264 This adds a note to the README and docs to let users know that they can build in parallel when available with an additional argument to `cmake build`.

I also take the liberty of removing a now defunct note aboiut building using `make` for CMake < 3.15 as we recently set the minimum version to this (see #228).

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,11 @@ To build and install the library:
     cmake --install .
     ```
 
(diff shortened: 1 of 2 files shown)


jwallwork23 (approved):
Nice, worth mentioning indeed.

PR #281, opened 2025-02-10
Avoid copying examples when building tests

### LINKED ISSUE #279: Avoid copying examples when building tests
Following a discussion in the meeting today we decided it might be better to move the main `CMakeLists.txt` down to the root level and have the integration tests build without moving `examples/`. Similarly for `tests/`.

jwallwork23:
Closes #279.

Simplifies the developer experience, as discussed in today's meeting.

* `FTorch/src/CMakeLists.txt` moved to `FTorch/CMakeLists.txt`.
* `FTorch/src/tests/` moved to `FTorch/tests/`.
* No longer any need to copy the examples when building FTorch with test support.
* Updated docs accordingly.
* Added API change note.


jatkinson1000 (changes requested):
Hi @jwallwork23 CMake adjustments look fine, but I think you need to update the installation instructions on the main `README` and in the `pages/` directory to reflect that users no longer build in `src/build/`.

Would be worth a comb through the rest of the READMEs and docs to check for anything else that may need changing.

jwallwork23:
> Hi @jwallwork23 CMake adjustments look fine, but I think you need to update the installation instructions on the main `README` and in the `pages/` directory to reflect that users no longer build in `src/build/`.
> 
> Would be worth a comb through the rest of the READMEs and docs to check for anything else that may need changing.

Thanks @jatkinson1000 - I think I covered everything with e0e85d3. Grepped for `src` and updated anything needed.

jatkinson1000 (approved):
Thanks @jwallwork23 On the assumption you catched 'em all I'm happy to approve this.

Issue #285, opened 2025-02-18
Wrap Torch Scalar


jwallwork23:
The Torch C++ API includes a `Scalar` class, which should streamline handling scalar values in the autograd work.

We should wrap this class as `torch_scalar`, as well as it's core functionality.

jwallwork23:
I spent a while trying to get this set up but found it was very difficult to deal with the different integer and real types getting cast appropriately. It'd probably be better to define `torch_scalar` as a rank-1 `torch_tensor` with a single element, as done in #286.

PR #286, opened 2025-02-18
Implement `torch_tensor_backward`

### LINKED ISSUE #152: Implement `torch_tensor_backward`
Related to #111.

This will be needed to complete the autograd example in #139. Probably best to do this after that PR is merged, otherwise it will become unwieldy.

jwallwork23:
Closes #152.

This PR:
* ~~Sets the `requires_grad` argument to tensor constructors up properly.~~ (Done in #288.)
* Wraps the `tensor->backward` method as `torch_tensor_backward`.
* Provides a method for retrieving the gradient of a tensor (with respect to a tensor on which the user has called `torch_tensor_backward`): `torch_tensor_get_gradient`.
  * For similarity with PyTorch, I also provided a method `torch_tensor%grad()` (c.f. `tensor.grad`).
* Provides unit tests for the gradients of the overloaded operators.
* Updates the autograd example accordingly.
* ~~Revises the scalar multiplication/division approach as described below.~~ (Done in #289.)

~~I spent far too much time trying to get the gradients to be computed correctly for scalar multiplication and division as it is currently implemented. However, I discovered that everything works nicely if we just define scalars as rank-1 tensors with a single value. The revised approach to scalar multiplication/division adds a bit of boilerplate, but perhaps this could be reduced by introducing a `torch_scalar` class (see #285 for discussion).~~


jwallwork23:
I'll break this PR down into more manageable ones as it's got rather unwieldy.

jwallwork23:
Ah, just noticed that I haven't updated `pages/autograd.md` for this PR yet
```
### The `backward` operator

*Not yet implemented.*
```
Will do that.

jwallwork23:
> Ah, just noticed that I haven't updated `pages/autograd.md` for this PR yet
> 
> ```
> ### The `backward` operator
> 
> *Not yet implemented.*
> ```
> 
> Will do that.

Done in 883654a40847017dfd0443557fe26d3ee8c765a1.

src/ftorch.F90
```diff
@@ -36,6 +36,7 @@ module ftorch
     procedure :: get_dtype => torch_tensor_get_dtype
     procedure :: get_device_type => torch_tensor_get_device_type
     procedure :: get_device_index => torch_tensor_get_device_index
+    procedure :: grad => torch_tensor_get_gradient
```
jatkinson1000, line 39: Wonder if this should be `get_grad` for consistency with the other methods.
Or remove `get_` from other methods for consistency with python's `Tensor.dtype` notation.

src/ftorch.F90
```diff
@@ -1815,6 +1816,61 @@ end subroutine torch_tensor_power_float_c
...
+                           device_index=tensor%get_device_index())
+
+    ! Call back-propagation with the provided external gradient
+    call torch_tensor_backward_c(tensor%p, external_gradient%p)
+
+    ! Delete the external gradient tensor
+    call torch_tensor_delete(external_gradient)
+  end subroutine torch_tensor_backward
+
+  !> Retrieves the gradient with respect to a Torch Tensor.
```
jatkinson1000, line 1851: Add a note that if the lhs tensor has not been created already than it will be generated using empty?

jatkinson1000: Further, what happens if we have a `gradient` with inappropriate dimensions?
I presume it complains at the LibTorch level?

examples/7_Autograd/autograd.f90
```diff
@@ -42,17 +42,37 @@ program example
...
-    print *, "Error :: value of Q does not match expected value"
+  if (.not. assert_allclose(out_data1, expected, test_name="autograd_Q")) then
+    write(*,*) "Error :: value of Q does not match expected value"
     stop 999
   end if
 
-  ! Back-propagation
-  ! TODO: Requires API extension
+  ! Run the back-propagation operator
+  call torch_tensor_backward(Q)
```
jatkinson1000, line 55: I would appreciate a little more info on what/how this does.
e.g.
```suggestion
  ! Run the back-propagation operator
  ! This will perform backpropogation on the tensors involved in generating Q (a and b), setting the `grad` property for both of them.
  call torch_tensor_backward(Q)
```

src/ftorch.F90
```diff
@@ -1815,6 +1816,61 @@ end subroutine torch_tensor_power_float_c
...
 
 
+  ! ============================================================================
+  ! --- Procedures related to automatic differentation functionality for tensors
+  ! ============================================================================
+
+  !> Performs back-propagation on a Torch Tensor, given some external gradient.
+  subroutine torch_tensor_backward(tensor)
+    type(torch_tensor), intent(in) :: tensor
+    type(torch_tensor) :: external_gradient
```
jatkinson1000, line 1826: Add variable docs with inline `  !!` like other functions.
Personally `external_gradient` wasn't immediately obvious to me, especially without the `TODO` note.

src/ftorch.F90
```diff
@@ -1815,6 +1816,61 @@ end subroutine torch_tensor_power_float_c
...
+    end interface
+
+    ! External gradient to provide to the back-propagation consisting of a tensor of ones
+    ! TODO: Accept other external gradients as an optional argument
+    call torch_tensor_ones(external_gradient, tensor%get_rank(), tensor%get_shape(), &
+                           tensor%get_dtype(), tensor%get_device_type(), &
+                           device_index=tensor%get_device_index())
+
+    ! Call back-propagation with the provided external gradient
+    call torch_tensor_backward_c(tensor%p, external_gradient%p)
```
jatkinson1000, line 1845: To check my understanding, we call this on one `tensor`, and it propagates to all other tensors previously used to calculate `tensor` that have `requires_grad` through bookkeeping  on the LibTorch side?

pages/autograd.md
```diff
@@ -45,6 +45,45 @@ Torch tensors, see the associated
...
+call torch_tensor_from_array(b, in_data2, tensor_layout, torch_kCPU, &
+                             requires_grad=.true.)
+call torch_tensor_from_array(Q, out_data1, tensor_layout, torch_kCPU)
+
+Q = a * b
+
+call torch_tensor_backward(Q)
+```
+
+In the example code above, we can extract gradients of `Q` with respect to `a`
```
jatkinson1000, line 68:
```suggestion
Following the example code above, we can extract gradients of `Q` with respect to `a`
```

pages/autograd.md
```diff
@@ -45,6 +45,45 @@ Torch tensors, see the associated
...
+
+```fortran
+! Function approach
+call torch_tensor_from_array(dQda, out_data2, tensor_layout, torch_kCPU)
+dQda = torch_tensor_get_gradient(a)
+
+! Method approach
+call torch_tensor_from_array(dQdb, out_data3, tensor_layout, torch_kCPU)
+dQdb = b%grad()
+```
```
jatkinson1000, line 81: So, what happens if a and b are used in two operations:
```fortran
Q = a * b
P = a**2 / b
```
and we then call backward on both:
```fortran
call torch_tensor_backward(Q)
call torch_tensor_backward(P)
```

My gut says that `a%grad` will be `dPda` having been overwritten by second call?

Also raises some memory questions, ~though perhaps these will be answered when I look at the C++ for this~ do gradients that are created by `backward` get deleted at some point? Or do we just assume they are small and are overwritten with each call so won't stack up over time/successive calls?

pages/autograd.md
```diff
@@ -45,6 +45,45 @@ Torch tensors, see the associated
 For Tensors that you would like to differentiate with respect to, be sure to
 set the `requires_grad` optional argument to `.true.` when you construct it.
 
-### The `backward` operator
+### Back-propagation
```
dorchard, line 48:
```suggestion
### Backpropagation
```

jatkinson1000 (commented):
Thanks @jwallwork23 this is fantastic and will be really useful in #320 

A couple of queries for you before I approve.

src/ftorch.F90
jwallwork23: Ah yes. Given that we want to provide an interface that's familiar to both Fortran and PyTorch users, I'd prefer to go with `torch_tensor%dtype` indeed.

Following this line of thought, though, the question of what to do with `get_device_type` and `get_device_index` to get consistency brings us to introducing a `torch_device` derived type that has `type` and `index` (c.f. `Tensor.device.type` and `Tensor.device.index`).

It's starting to feel like a separate discussion/PR?

jatkinson1000: Agree this is something to resolve in a separate PR.
I also think I would aim to match the torch interface meaning we leave the new code as is.

Would be keen not to implement a device type and have `device_type` as a suitably close equivalent to `device.type` I think?

jwallwork23: Yes, I agree that `device_type` and `device_index` are fine.

src/ftorch.F90
jwallwork23: Oops, fixed in 26b4fd0.

src/ftorch.F90
jwallwork23: Ah yes of course - there's no guarantee the two have the same shape.

In the process of switching to raise an error instead I've gone down a bit of a rabbit hole and am now concerned that we are querying the return values of functions... will report back.

jwallwork23: The issue is illustrated in 8e932803b86eabe7b4994036a41743818b4c5423. I get the output
```
4:  Start: <test_tensor_operator_overloads_autograd_suite.test_torch_tensor_assign[F][F]>
4: . associated? T
4:  associated? F
4:  Error :: tensors for holding gradients must be constructed before retrieving values
4: STOP 1
4/4 Test #4: test_operator_overloads_autograd .......***Failed    0.25 sec
```
i.e., that the pointer for the tensor is associated before calling `torch_tensor_get_gradient` but inside the function it doesn't know about it.

jwallwork23: Okay I fixed the issue in c4e2cd8 by turning `torch_tensor_get_gradient` into a subroutine and making use of `std::move`.

However, I have two comments:
1. As mentioned, the operator overloads are querying their `result`s before setting them, although this surprisingly works.
2. The new subroutine notation `call a%grad(dQda)` doesn't read like Fortran or PyTorch so maybe should be dropped, leaving `torch_tensor_get_gradient` as the only interface?

jatkinson1000: Agree that `call a%grad(dQda)` is not a clear syntax.
It's a shame that the original approach didn't work, as this was quite clean and similar to PyTorch, but perhaps a specific ~function~ subroutine to query is the way to go.

Maybe we need to work with a general ethos that any tensors used anywhere in FTorch should be explicitly created before being assigned from anywhere.

jwallwork23: True, although `torch_tensor_X` is the pattern we've been using for procedures operating on tensors more generally than just those bound as methods.

pages/autograd.md
jwallwork23: In order to get the correct gradients, we are going to need to implement zeroing of gradients. See #336.

Personally, I think we can go ahead with merging this long-standing PR and then I'll look into that issue as follow-up. Sound good?

jatkinson1000: Yep, I definitely agree.
Having this merged will make it easier to open up a number of further avenues for development.

src/ftorch.F90
jatkinson1000: For posterity, this was dropped as a tensor method in https://github.com/Cambridge-ICCS/FTorch/pull/286/commits/18f8be1659fd83b6384826e4d3a1a0191d074cf8

jatkinson1000 (approved):
Thanks for the great work @jwallwork23 and for exploring the queries I raised.
I think this is a great addition and sets us up to do some exciting stuff.
Happy for you to merge after resolving the conflicts.

PR #287, opened 2025-02-18
Fix static analysis workflow

jwallwork23:
Remove errant `src` that was missed in #281.


jatkinson1000 (approved):
Tepid fix.
Found this yesterday and checked happy to approve this to go in.

Issue #158, opened 2024-07-15
Implement `torch_model_backward`


joewallwork:
Related to #152.

Required for #111.

PR #288, opened 2025-02-18
Hook up `requires_grad`

jwallwork23:
Towards #158.
(Refined from #286.)

This PR hooks the `requires_grad` option up properly and provides a `torch_tensor_requires_grad` function (as well as a `torch_tensor%requires_grad` method) to query this. Unit tests are written and the option is also used in the autograd example.

### DIFF
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -151,9 +151,12 @@ torch_tensor_t torch_empty(int ndim, const int64_t *shape, torch_data_t dtype,
   try {
     // This doesn't throw if shape and dimensions are incompatible
     c10::IntArrayRef vshape(shape, ndim);
+    auto options = torch::TensorOptions()
+                       .dtype(get_libtorch_dtype(dtype))
+                       .device(get_libtorch_device(device_type, device_index))
+                       .requires_grad(requires_grad);
     tensor = new torch::Tensor;
-    *tensor = torch::empty(vshape, torch::dtype(get_libtorch_dtype(dtype)))
-                  .to(get_libtorch_device(device_type, device_index));
+    *tensor = torch::empty(vshape, options);
   } catch (const torch::Error &e) {
     std::cerr << "[ERROR]: " << e.msg() << std::endl;
     delete tensor;
@@ -174,9 +177,12 @@ torch_tensor_t torch_zeros(int ndim, const int64_t *shape, torch_data_t dtype,
   try {
     // This doesn't throw if shape and dimensions are incompatible
     c10::IntArrayRef vshape(shape, ndim);
+    auto options = torch::TensorOptions()
+                       .dtype(get_libtorch_dtype(dtype))
+                       .device(get_libtorch_device(device_type, device_index))
+                       .requires_grad(requires_grad);
     tensor = new torch::Tensor;
-    *tensor = torch::zeros(vshape, torch::dtype(get_libtorch_dtype(dtype)))
-                  .to(get_libtorch_device(device_type, device_index));
+    *tensor = torch::zeros(vshape, options);
   } catch (const torch::Error &e) {
     std::cerr << "[ERROR]: " << e.msg() << std::endl;
     delete tensor;
@@ -197,9 +203,12 @@ torch_tensor_t torch_ones(int ndim, const int64_t *shape, torch_data_t dtype,
   try {
     // This doesn't throw if shape and dimensions are incompatible
     c10::IntArrayRef vshape(shape, ndim);
+    auto options = torch::TensorOptions()
+                       .dtype(get_libtorch_dtype(dtype))
+                       .device(get_libtorch_device(device_type, device_index))
+                       .requires_grad(requires_grad);
     tensor = new torch::Tensor;
-    *tensor = torch::ones(vshape, torch::dtype(get_libtorch_dtype(dtype)))
-                  .to(get_libtorch_device(device_type, device_index));
+    *tensor = torch::ones(vshape, options);
   } catch (const torch::Error &e) {
     std::cerr << "[ERROR]: " << e.msg() << std::endl;
     delete tensor;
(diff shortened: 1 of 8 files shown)


src/ctorch.cpp
```diff
@@ -310,7 +326,7 @@ torch_tensor_t torch_tensor_assign(const torch_tensor_t input) {
   torch::AutoGradMode enable_grad(in->requires_grad());
   torch::Tensor *output = nullptr;
   output = new torch::Tensor;
-  *output = in->detach().clone();
+  *output = *in;
```
jwallwork23, line 329: Not sure how this change made it into the PR but it's fairly minor so perhaps I'll leave it. This change just copies by value rather than using the `detach` and `clone` methods, which turn out to be unnecessary.

src/ctorch.cpp
```diff
@@ -122,9 +122,12 @@ torch_tensor_t torch_empty(int ndim, const int64_t *shape, torch_data_t dtype,
   try {
     // This doesn't throw if shape and dimensions are incompatible
     c10::IntArrayRef vshape(shape, ndim);
+    auto options = torch::TensorOptions()
+                       .dtype(get_libtorch_dtype(dtype))
+                       .device(get_libtorch_device(device_type, device_index))
+                       .requires_grad(requires_grad);
```
jatkinson1000, line 128: Much nicer than the previous `torch::dtype()` and `.to()` calls of the previous code.

src/ctorch.cpp
```diff
@@ -292,6 +303,11 @@ int torch_tensor_get_device_index(const torch_tensor_t tensor) {
   return t->device().index();
 }
 
+bool torch_tensor_requires_grad(const torch_tensor_t tensor) {
+  auto t = reinterpret_cast<torch::Tensor *>(tensor);
+  return t->requires_grad();
+}
+
```
jatkinson1000, line 310: I trust you are correct, but to check -- we need this because it is carried around at the LibTorch level, not the FTorch level, so cannot be directly interrogated from Fortran?

test/unit/test_tensor_interrogation.pf
```diff
@@ -232,4 +232,30 @@ contains
...
+    ! Check that torch_tensor_get_device_type can get the device type
+    if (expected .neqv. tensor%requires_grad()) then
+      call torch_tensor_delete(tensor)
+      print *, "Error :: tensor%requires_grad() returned incorrect value"
+      stop 999
+    end if
+    call torch_tensor_delete(tensor)
+
+  end subroutine test_torch_tensor_requires_grad
+
```
jatkinson1000, line 260: I suggest adding to this test to check that the default creation of a tensor (no `requires_grad` option) has grad set to false as well as checking that it is true when explicitly specified.

Tried to write as a suggestion, but feel free to refactor if you would prefer it done another way, perhaps as 2 separate tests.

```suggestion
  ! Unit test for the torch_tensor_requires_grad function
  @test
  subroutine test_torch_tensor_requires_grad()

    implicit none

    type(torch_tensor) :: tensor
    integer, parameter :: ndims = 1
    integer(c_int64_t), parameter, dimension(ndims) :: tensor_shape = [1]
    integer, parameter :: dtype = torch_kFloat32
    logical, parameter :: grad_expected = .true.
    logical, parameter :: grad_not_expected = .false.


    ! Create an empty tensor on the CPU with the default device type and no requires_grad
    call torch_tensor_empty(tensor, ndims, tensor_shape, dtype, device_type)

    ! Check that torch_tensor_requires_grad returns the correct setting
    if (grad_not_expected .neqv. tensor%requires_grad()) then
      call torch_tensor_delete(tensor)
      print *, "Error :: tensor%requires_grad() returned incorrect value (expected .false.)"
      stop 999
    end if
    call torch_tensor_delete(tensor)

    ! Create an empty tensor on the CPU with the default device type and requires_grad true
    call torch_tensor_empty(tensor, ndims, tensor_shape, dtype, device_type, &
                            requires_grad=grad_expected)

    ! Check that torch_tensor_requires_grad returns the correct setting
    if (grad_expected .neqv. tensor%requires_grad()) then
      call torch_tensor_delete(tensor)
      print *, "Error :: tensor%requires_grad() returned incorrect value (expected .true.)"
      stop 999
    end if
    call torch_tensor_delete(tensor)

  end subroutine test_torch_tensor_requires_grad
```

jatkinson1000 (changes requested):
Hi @jwallwork23 Generally seems sensible to me.

Main request is to test the defaults for requires_grad, not just the `true` setting.

I realise this probably needs a bit of tweaking/rebasing after a few of the recent changes.

src/ctorch.cpp
jwallwork23: Yeah I considered just tracking this on the Fortran side but thought it would be more consistent and fool-proof if we had the Torch Tensor as the single source of truth and always delegate getting and setting to that.

test/unit/test_tensor_interrogation.pf
jwallwork23: Thanks for pointing this out. Addressed this by introducing fixtures to this test set in 68121359f60b884fb7bf5fbc39ab6cd3cc0aaaa5.

src/ftorch.fypp
```diff
@@ -626,9 +645,9 @@ contains
     end interface
 
     if (.not. c_associated(output%p)) then
-      ! TODO: Pass requires_grad argument
       call torch_tensor_empty(output, input%get_rank(), input%get_shape(), input%get_dtype(), &
-                              input%get_device_type(), input%get_device_index())
+                              input%get_device_type(), device_index=input%get_device_index(), &
+                              requires_grad=input%requires_grad())
     else if (input%get_device_type() /= output%get_device_type()) then
```
jwallwork23, line 651: Addressed this TODO and all other similar ones.

jwallwork23:
[Rebased on top of `main`.]

test/unit/test_tensor_interrogation.pf
jatkinson1000: Yes, I did feel that this would be a better way to do it 😉

src/ctorch.cpp
jatkinson1000: 100% agree.

jatkinson1000 (approved):
Thanks @jwallwork23 an even better fix than I imagined.
Hopefully fixtures should come in useful as we expand things.

jwallwork23:
Thanks for the review @jatkinson1000!

PR #289, opened 2025-02-18
Rework scalar multiplication and division

jwallwork23:
Towards #158.
(Refined from #286.)

This PR revises the scalar multiplication/division approach as described on #286:

> I spent far too much time trying to get the gradients to be computed correctly for scalar multiplication and division as it is currently implemented. However, I discovered that everything works nicely if we just define scalars as rank-1 tensors with a single value. The revised approach to scalar multiplication/division adds a bit of boilerplate, but perhaps this could be reduced by introducing a `torch_scalar` class (see #285 for discussion).

There are lots of deletions because several fypp sections are no longer needed.

The PR also adds some info on this to both the autograd doc page and the autograd example.


jwallwork23:
[Rebased on top of `main`.]

examples/6_Autograd/autograd.f90
```diff
@@ -17,42 +16,37 @@ program example
...
   ! Set up Fortran data structures
-  integer, parameter :: ndims = 2
-  integer, parameter :: n=2, m=1
-  real(wp), dimension(n,m), target :: in_data1
-  real(wp), dimension(n,m), target :: in_data2
-  real(wp), dimension(n,m), target :: out_data
-  real(wp), dimension(n,m) :: expected
-  integer :: tensor_layout(ndims) = [1, 2]
+  integer, parameter :: ndims = 1
+  integer, parameter :: n = 2
```
jwallwork23, line 20: We'd always intended for the Fortran version of the autograd example to mimic the Python version, which just operates on vectors of length 2. However, at some point 2x1 arrays got introduced as pre-unit test code coverage. Given we now have unit tests, let's try to keep the Fortran and Python versions of this demo as similar as possible.

jwallwork23:
I checked what happens if we try to do scalar multiplication in the way that previously would have worked and get a build error
```
/home/joe/software/FTorch/examples/6_Autograd/autograd.f90:44:6:

   44 |   Q = 3.0 * (a**3 - b * b / divisor)
      |      1
Error: Unexpected derived-type entities in binary intrinsic numeric operator ‘*’ at (1)
```
So we don't need to worry about users doing this. Similarly for division:
```
/home/joe/software/FTorch/examples/6_Autograd/autograd.f90:44:34:

   44 |   Q = multiplier * (a**3 - b * b / 3.0)
      |                                  1
Error: Unexpected derived-type entities in binary intrinsic numeric operator ‘/’ at (1)
```

examples/6_Autograd/autograd.f90
```diff
@@ -17,42 +16,37 @@ program example
...
-  Q = 3 * (a**3 - b * b / 3)
+  ! Scalar multiplication and division are not currently implemented in FTorch. However, you can
+  ! achieve the same thing by defining a rank-1 tensor with a single entry, as follows:
+  call torch_tensor_from_array(multiplier, [3.0_wp], tensor_layout, torch_kCPU)
+  call torch_tensor_from_array(divisor, [3.0_wp], tensor_layout, torch_kCPU)
 
-  ! Extract a Fortran array from a Torch tensor
-  write (*,*) "Q = 3 * (a ** 3 - b * b / 2) =", out_data(:,1)
+  ! Compute the same mathematical expression as in the Python example
+  Q = multiplier * (a**3 - b * b / divisor)
```
jatkinson1000, line 44: Since it has come up elsewhere for me I wonder if there is a better naming convention than `divisor` to suggest users adopt. e.g. `torch_scalar_3p0`, though that is not ideal as torch_scalar is a different thing we did not use...

A non-blocking musing.

examples/6_Autograd/autograd.f90
```diff
@@ -17,42 +16,37 @@ program example
...
+  write (*,*) "Q = 3 * (a^3 - b*b/3) = 3*a^3 - b^2 = ", out_data(:)
 
   ! Check output tensor matches expected value
-  expected(:,1) = [-12.0_wp, 65.0_wp]
```
jatkinson1000, line 52: `expected` is now unused.
I would potentially keep it, rather than passing `assert_allclose` a magic number below, to clarify intent.
Otherwise remove the declaration.

examples/6_Autograd/autograd.f90
```diff
@@ -17,42 +16,37 @@ program example
...
-  type(torch_tensor) :: a, b, Q
+  type(torch_tensor) :: a, b, Q, multiplier, divisor
 
   ! Initialise Torch Tensors from input arrays as in Python example
-  in_data1(:,1) = [2.0_wp, 3.0_wp]
-  in_data2(:,1) = [6.0_wp, 4.0_wp]
-  call torch_tensor_from_array(a, in_data1, tensor_layout, torch_kCPU, requires_grad=.true.)
-  call torch_tensor_from_array(b, in_data2, tensor_layout, torch_kCPU, requires_grad=.true.)
+  call torch_tensor_from_array(a, [2.0_wp, 3.0_wp], tensor_layout, torch_kCPU, requires_grad=.true.)
+  call torch_tensor_from_array(b, [6.0_wp, 4.0_wp], tensor_layout, torch_kCPU, requires_grad=.true.)
```
jatkinson1000, line 33: A nifty thing to be aware of, reducing boilerplate when we only need to declare the tensor to be used.

jatkinson1000 (approved):
Thanks for checking that @jwallwork23 
Not 100% explicit, but reasonably helpful I feel, at least for now - we can perhaps revisit error handling more generally in future if we feel it would be useful.

As discussed in-person agree that this is the easiest way to implement this to facilitate backprop in a straightforward manner.

Couple of small points to check but trust you to do so then merge.

Looks like you'll need a rebase to pick up latest changes.

examples/6_Autograd/autograd.f90
jwallwork23: Hm, could do just `f3p0` where the `f` indicates float?

examples/6_Autograd/autograd.f90
jwallwork23: Rebased on top of 8e03e81292980c7616cca740b08199c15cab80c4.

jwallwork23:
[Rebased on top of `main`.]

examples/6_Autograd/autograd.f90
jatkinson1000: Could do, but that makes it less clear that it is a torch/tensor implementation of 3.0.
Not requisite to decide here I feel, but just a passing musing.
Motivated mostly by the fact 'real-world' codes might have several `divisor` etc that they want to track.

PR #296, opened 2025-02-25
Patch docs to provide preprocessor macros to Ford

jatkinson1000:
Definition of macros is required for generating documentation. Currently docs do not build properly as Ford cannot parse ftorch.F90. Also adds a note to the developer docs to remind people this needs keeping in sync.

Note - keep an eye on this in future, if Ford cannot parse a source file it generates a "warning" of the error rather than failing, so future PRs should include a cursory glance over what is generated.

### DIFF
--- a/FTorch.md
+++ b/FTorch.md
@@ -15,7 +15,8 @@ extra_filetypes: c   //
                  h //
                  py  #
 preprocess: true
-macro: GPU_DEVICE_NONE=0
+macro: UNIX
+       GPU_DEVICE_NONE=0
        GPU_DEVICE_CUDA=1
        GPU_DEVICE_XPU=11
        GPU_DEVICE_MPS=12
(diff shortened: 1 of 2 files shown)


jwallwork23 (approved):
Ah okay, thanks for figuring this out @jatkinson1000. Can confirm the docs build on CI doesn't have any warnings other than `Warning: Will not be able to generate graphs. Graphviz not installed.`, which doesn't matter for us.

How about the `UNIX` variable? Do you know what happens to that?

jatkinson1000:
A sensible question @jwallwork23 
Since `UNIX` is a boolean used in an `#ifdef` the preprocessor will read it as 0 (not set equivalent to not present) and proceed, whereas it fails for the GPU devices as this number always needs to be supplied as a preprocessor macro.

However, on reflection given UNIX is the main target we should set this as well, which I have done in https://github.com/Cambridge-ICCS/FTorch/pull/296/commits/c633fbc3c2c5726b4a029c9f8f919f7036ea33a7

Will merge :)

PR #298, opened 2025-02-26
Cut down unit test code using destructor

jwallwork23:
This PR tidies up the unit tests considerably in light of #297.

There's no longer any need to call `torch_tensor_delete` (except for the test of it). I also spotted a way to refactor the error raising so that we don't need all the `clean_up` subroutines.

In some cases, we can use the inbuilt pFUnit assertions, which is nice.

### DIFF
--- a/test/unit/test_tensor_constructors.pf
+++ b/test/unit/test_tensor_constructors.pf
@@ -133,21 +133,14 @@ contains
     ! Check that the tensor values are all zero
     expected(:,:) = 0.0
     test_pass = assert_allclose(out_data, expected, test_name="test_torch_tensor_zeros")
+
+    ! Nullify pointers used in the unit test
+    nullify(out_data)
+
     if (.not. test_pass) then
-      call clean_up()
       print *, "Error :: incorrect output from torch_tensor_zeros subroutine"
       stop 999
     end if
-
-    call clean_up()
-
-    contains
-
-      ! Subroutine for freeing memory and nullifying pointers used in the unit test
-      subroutine clean_up()
-        nullify(out_data)
-        call torch_tensor_delete(tensor)
-      end subroutine clean_up
 
   end subroutine test_torch_tensor_zeros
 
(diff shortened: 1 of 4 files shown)


jatkinson1000 (approved):
Yes, all makes sense to me. Thanks @jwallwork23 
I like that the nullification is now at the same level as declaration of tensors/pointers and agree we should use the pFUnit preprocessor directives where we can.

Couple of questions around `@assertEqual`:

- What is the output of `@assertEqual`? Previously you had very explicit error messages relating to the test/routine/rank whereas I presume they are now more opaque/generic?
- I presume you checked that `@assertEqual` fails as expected when you break something? 😉

jwallwork23:
>     * What is the output of `@assertEqual`? Previously you had very explicit error messages relating to the test/routine/rank whereas I presume they are now more opaque/generic?

So with the diff
```diff
diff --git a/test/unit/test_tensor_constructors.pf b/test/unit/test_tensor_constructors.pf
index 5e0e2c8..3b7d94a 100644
--- a/test/unit/test_tensor_constructors.pf
+++ b/test/unit/test_tensor_constructors.pf
@@ -93,7 +93,7 @@ contains
     call torch_tensor_delete(tensor)
 
     ! Check torch_tensor_delete does indeed free the memory
-    @assertFalse(c_associated(tensor%p))
+    @assertTrue(c_associated(tensor%p))
 
   end subroutine test_torch_tensor_empty
```
we get the failure
```
1: Failure
1:  in: 
1: test_tensor_constructors_suite.test_torch_tensor_empty[F][F]
1:   Location: 
1: [test_tensor_constructors.pf:96]
1: 
1:   
1: Failure
1:  in: 
1: test_tensor_constructors_suite.test_torch_tensor_empty[T][T]
1:   Location: 
1: [test_tensor_constructors.pf:96]
```
Personally, I think the line number is sufficient for reports of failed assertions.
 
>     * I presume you checked that `@assertEqual` fails as expected when you break something? 😉

Yep

PR #303, opened 2025-03-03
Drop `torch_tensor_to_array`

### LINKED ISSUE #293: `torch_tensor_to_array` does not account for index ordering
Running the tensor manipulation example proposed in #291, I notice that `torch_tensor_to_array` does not properly account for index ordering. The output of the example is currently
```
test 1
    Start 1: tensor_manipulation

1: Test command: /home/joe/software/FTorch/build/examples/0_Tensor/tensor_manipulation
1: Test timeout computed to be: 1500
1:  Contents of first input tensor:
1:  1  1  1
1:  1  1  1
1: [ CPUFloatType{2,3} ]
1:  Contents of second input tensor:
1:    1.00000000       2.00000000       3.00000000       4.00000000       5.00000000       6.00000000    
1:  Output:
1:    2.00000000       4.00000000       6.00000000       3.00000000       5.00000000       7.00000000    
1:  Tensor manipulation example ran successfully
1/1 Test #1: tensor_manipulation ..............   Passed    0.23 sec

100% tests passed, 0 tests failed out of 1
```
It takes a $2\times3$ tensor of ones and adds it to a $2\times3$ tensor containing the integers 1-6 that was created using the Fortran `reshape` intrinsic. The array generated by the reshape is printed as the "second input tensor". We do the addition using the overloaded `+` operator but then when we call `torch_tensor_to_array`, the output is not what we'd expect.

I think the issue is that `torch_tensor_from_blob` accounts for strides, whereas `torch_tensor_to_blob` doesn't. Therefore, we switch index orderings from Fortran to C, but then don't switch back.

jwallwork23:
Closes #293.

Note #272 will need to be merged first as this branch builds upon it.

As discussed in today's meeting, `torch_tensor_to_array` is no longer needed. It was added because the overloaded mathematical operators created new `torch:Tensor`s in C++, rather than working with pointers to Fortran arrays. Since that was addressed in #297, the expected (correct) result from `torch_tensor_to_array` can be achieved by creating the tensor from which you extract data from using `torch_tensor_from_array` applied to the array you want to extract into.

This is much neater and allows us to drop all cleanup (the remaining `nullify` statements) from the unit tests.


pages/updates.md
```diff
@@ -1,9 +1,22 @@
...
+[9b90099](9b90099759a4cf0a1844c53327f15c318eba9b7c)
+(March 2025) you might notice the existence of a `torch_tensor_to_array`
+subroutine. It turned out to be unnecessary and so has been deleted. To achieve
+the same effect of extracting a Fortran array from a Torch Tensor, you simply
+need to ensure that it was constructed based off that Fortran array using
+`torch_tensor_from_array`. This is demonstrated in both the
+[SimpleNet example](https://github.com/Cambridge-ICCS/FTorch/tree/main/examples/1_SimpleNet)
+and
+[autograd example](https://github.com/Cambridge-ICCS/FTorch/tree/main/examples/6_Autograd).
+
```
jatkinson1000, line 15: I wonder if this is necessary given we are pretty sure no-one is actually using this functionality at present?

src/ctorch.cpp
```diff
@@ -338,7 +300,7 @@ void torch_tensor_assign(torch_tensor_t output, const torch_tensor_t input) {
   auto out = reinterpret_cast<torch::Tensor *>(output);
   auto in = reinterpret_cast<torch::Tensor *const>(input);
   torch::AutoGradMode enable_grad(in->requires_grad());
-  *out = *in;
+  std::move(*out) = *in;
```
jatkinson1000, line 303: My very limited C++ I usually see `std::move` on the rhs.
But I can imagine this is different when we overload `=`.
Can you explain what this is doing for me?

jatkinson1000 (approved):
Thanks @jwallwork23 The code all looks sensible to me, but I have a couple of things I would appreciate if you can explain.

To summarise my understanding: instead of generating a tensor (pointer) that needs to then be mapped to an array, we instead require a user to 'pre-bind' an array to the output tensor of an arithmetic expression that will then be populated with any data resulting from computation. Is that correct?

It occurs to me here that all of our tests are for "correct" outputs, we do not test for errors. As discussed last week, this is not doable in pFUnit, but out of interest do you have an idea of what the outcome is if the user does not "pre-declare" the array associated with the output tensor?

Also, just to check my understanding, we do not call `torch_tensor_delete` in the autograd example at all now because it is cleaned up by the `final` method when going out of scope at the end of the program?

jwallwork23:
> To summarise my understanding: instead of generating a tensor (pointer) that needs to then be mapped to an array, we instead require a user to 'pre-bind' an array to the output tensor of an arithmetic expression that will then be populated with any data resulting from computation. Is that correct?

That's correct.

> It occurs to me here that all of our tests are for "correct" outputs, we do not test for errors. As discussed last week, this is not doable in pFUnit, but out of interest do you have an idea of what the outcome is if the user does not "pre-declare" the array associated with the output tensor?

Well you can't use `torch_tensor_from_array` without having declared the array, no?

> Also, just to check my understanding, we do not call `torch_tensor_delete` in the autograd example at all now because it is cleaned up by the `final` method when going out of scope at the end of the program?

Yes, I dropped it for simplicity since we can do that now.

pages/updates.md
jwallwork23: Fair, dropped in edd0aa8f021d00af29f70fa0cdad56beb2a794c3.

src/ctorch.cpp
jatkinson1000: Ah, thanks.
If it warranted a comment on the PR might be worth a brief comment in the code to clarify that it is correct and why needed?

jwallwork23: Good point. Done in 7eca225.

jatkinson1000 (approved):
Thanks. Already approved, but happy to re-approve after changes and feel free to merge as you please.

PR #304, opened 2025-03-03
Add a standalone `SimpleNet` example build to the CI

jwallwork23:
This PR adds an additional build step for one of the examples to check that we can build it with CMake separately from FTorch without needing to set `LD_LIBRARY_PATH` etc.

### DIFF
--- a/.github/workflows/test_suite_ubuntu.yml
+++ b/.github/workflows/test_suite_ubuntu.yml
@@ -108,3 +108,17 @@ jobs:
         run: |
           . ftorch/bin/activate
           ./run_test_suite.sh --verbose --integration-only
+
+      - name: Standalone SimpleNet example build
+        run: |
+          . ftorch/bin/activate
+          export FTORCH_BUILD_DIR="$(pwd)/build"
+          cd examples/1_SimpleNet
+          export BUILD_DIR="$(pwd)/build"
+          mkdir "${BUILD_DIR}"
+          cd "${BUILD_DIR}"
+          cmake .. \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_INSTALL_PREFIX="${BUILD_DIR}" \
+            -DCMAKE_Fortran_FLAGS="-std=${{ matrix.std }}"
+          cmake --build .
--- a/.github/workflows/test_suite_ubuntu.yml
+++ b/.github/workflows/test_suite_ubuntu.yml
@@ -83,16 +83,15 @@ jobs:
         run: |
           . ftorch/bin/activate
           VN=$(python -c "import sys; print('.'.join(sys.version.split('.')[:2]))")
-          export Torch_DIR=${VIRTUAL_ENV}/lib/python${VN}/site-packages
-          export BUILD_DIR=$(pwd)/build
+          export Torch_DIR="${VIRTUAL_ENV}/lib/python${VN}/site-packages"
+          export BUILD_DIR="$(pwd)/build"
           # NOTE: The pFUnit version (pinned during installation above) is used in the install path.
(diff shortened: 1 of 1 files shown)


jwallwork23:
Huh, surprised that this fails as the following script works fine for me locally
```sh
#!/bin/bash

FTORCH_BUILD_DIR="$(pwd)/build"
cd examples/1_SimpleNet
export BUILD_DIR="$(pwd)/build"
rm -rf "${BUILD_DIR}"
mkdir "${BUILD_DIR}"
cd "${BUILD_DIR}"
cmake .. \
  -DPython_EXECUTABLE="$(which python)" \
  -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_PREFIX_PATH="${FTORCH_BUILD_DIR}" \
  -DCMAKE_BUILD_TESTS=TRUE
cmake --build .
ctest -V
```

jatkinson1000:
Yep, pretty sure this matches what I was getting.
Interesting.

jatkinson1000 (changes requested):
The content looks fine @jwallwork23 but I think it needs a comment in there somewhere about what this is checking for and why we want it, given the same code is run as part of the bulk testing CI above.

Is there anywhere that now needs updating in the rest of the docs/codebase to tell users that they need to do this?

jwallwork23:
@TomMelt is there any chance you could take a look at this? I had something working in https://github.com/Cambridge-ICCS/FTorch/pull/304/commits/8e5f7a62f909c4b5673df44d9b5b45fed8af383d (by passing arguments to CMake) but I can't seem to get it to work in the CMakeLists.txt itself.

jwallwork23:
> My feeling is that it's not a bad suggestion, following that the lack of this is what was causing @jfdev001 issues in the first place and uncaught by us.

Good point.

> My one query is whether this is better to be placed in a separate workflow/test to avoid the testing workflow becoming too bloated (indeed, we could consider separating unit and integration tests as well).
> 
> The flipside argument of course is that the bulk of these workflows are spend in setup and install, not the actual tests themselves.

I think there's sufficient time taken to install everything that my preference would be to tack this on the end (as on the branch currently) rather than having a separate workflow.

CMakeLists.txt
```diff
@@ -87,6 +87,7 @@ endif()
 
 # Set RPATH behaviour
 set(CMAKE_SKIP_RPATH FALSE)
+set(CMAKE_BUILD_RPATH "${INSTALL_PREFIX}")
```
jfdev001: Why do you set this variable? What was breaking without it being set like this?

joewallwork: Good question. I think it's probably lying around from some previous WIP. I removed it in 16e692c, let's see what happens.

joewallwork: Tests pass so it seems it was redundant. Thanks for flagging this.

.github/workflows/test_suite_ubuntu_cpu_gnu.yml
```diff
@@ -217,3 +217,23 @@ jobs:
           . ftorch/bin/activate
           cd build
           ctest --verbose --tests-regex example
+
+      - name: Standalone SimpleNet example
```
jatkinson1000, line 222: Could add a comment what this is for?

```suggestion
      - name: Standalone SimpleNet example
      # check that we can successfully build and run an example outside the main FTorch build process.
```

jatkinson1000: And similarly below.

joewallwork: Done in 58e042b.

PR #310, opened 2025-03-06
Fortitude v0.7

jatkinson1000:
This PR updates to use the latest version of the Fortran fortitude linter which included a few API changes

Notably it adds:

- Ignoring of `C003` 'implicit none' missing 'external' as this is a f2013 feature and we want to support back to f2008.
- Adding an explicit `public` statement to the ftorch and test utils modules.
  - This is correct for the test utils, but I think we should have a discussion about whether everything in the ftorch module should be public.

### DIFF
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -99,7 +99,7 @@ jobs:
         run: |
           cd ${{ github.workspace }}
           . ftorch_venv/bin/activate
-          fortitude check --ignore=E001,T041 src/ftorch.F90
+          fortitude check src/ftorch.F90
           fortitude check src/ftorch_test_utils.f90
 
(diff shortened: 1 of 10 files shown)


jwallwork23 (approved):
>     * Ignoring of `C003` 'implicit none' missing 'external' as this is a f2013 feature and we want to support back to f2008.

That's fine.

>     * Adding an explicit `public` statement to the ftorch and test utils modules.
>       
>       * This is correct for the test utils, but I think we should have a discussion about whether everything in the ftorch module should be public.

I agree this is the right thing to do for now to maintain the current API. But yes, we should discuss whether some of the procedures should be made private in a future meeting.

jatkinson1000:
Thanks @jwallwork23 opened #311 to capture this for a future discussion and will merge this now.

Issue #300, opened 2025-02-27
Windows CI failing to find TorchScript during build


jwallwork23:
```
-- Building with Fortran PyTorch coupling
-- Configuring done (20.6s)
-- Generating done (0.2s)
-- Build files have been written to: D:/a/FTorch/FTorch/build
[  5%] Building Fortran object CMakeFiles/ftorch.dir/src/ftorch.F90.obj
[ 11%] Building Fortran object CMakeFiles/ftorch.dir/src/ftorch_test_utils.f90.obj
[ 16%] Building CXX object CMakeFiles/ftorch.dir/src/ctorch.cpp.obj
In file included from D:\a\FTorch\FTorch\src\ctorch.cpp:6:
In file included from C:\hostedtoolcache\windows\Python\3.13.2\x64\Lib\site-packages\torch\include\torch/script.h:3:
In file included from C:\hostedtoolcache\windows\Python\3.13.2\x64\Lib\site-packages\torch\include\torch/csrc/api/include/torch/types.h:3:
In file included from C:\hostedtoolcache\windows\Python\3.13.2\x64\Lib\site-packages\torch\include\ATen/ATen.h:7:
In file included from C:\hostedtoolcache\windows\Python\3.13.2\x64\Lib\site-packages\torch\include\ATen/Context.h:3:
In file included from C:\hostedtoolcache\windows\Python\3.13.2\x64\Lib\site-packages\torch\include\ATen/BlasBackend.h:3:
In file included from C:\hostedtoolcache\windows\Python\3.13.2\x64\Lib\site-packages\torch\include\c10/util/Exception.h:5:
In file included from C:\hostedtoolcache\windows\Python\3.13.2\x64\Lib\site-packages\torch\include\c10/macros/Macros.h:3:
In file included from C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.43.34808\include\cassert:7:
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.43.34808\include\yvals_core.h(927,1): error: static assertion failed: error STL1000: Unexpected compiler version, expected Clang 18.0.0 or newer.
_EMIT_STL_ERROR(STL1000, "Unexpected compiler version, expected Clang 18.0.0 or newer.");
[...]

PR #315, opened 2025-03-10
Temporarily disable Windows CI

jwallwork23:
The current issue with the Windows CI (#300) is causing those tests to fail. Rather than having those tests run and fail with each commit pushed to a PR, let's just disable the Windows CI for now.

### DIFF
--- a/.github/workflows/test_suite_windows.yml
+++ b/.github/workflows/test_suite_windows.yml
@@ -3,25 +3,26 @@ name: TestSuiteWindows
 
 # Controls when the workflow will run
 on:
-  # Triggers the workflow on pushes to the "main" branch, i.e., PR merges
-  push:
-    branches: [ "main" ]
-
-  # Triggers the workflow on pushes to open pull requests with code changes
-  pull_request:
-    paths:
-      - '.github/workflows/test_suite_windows.yml'
-      - '**.bat'
-      - '**.c'
-      - '**.cpp'
-      - '**.fypp'
-      - '**.f90'
-      - '**.F90'
-      - '**.pf'
-      - '**.py'
-      - '**CMakeLists.txt'
-      - '**requirements.txt'
-      - '**data/*'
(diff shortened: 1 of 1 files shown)


jatkinson1000 (approved):
Hi @jwallwork23 Yes, agree with doing this for the time being.

Tracked by #300 so we can keep an eye.
Other option would be to reduce frequency so we still run it periodically to see if a new image with a fix comes along...?

TomMelt:
> Hi @jwallwork23 Yes, agree with doing this for the time being.
> 
> Tracked by #300 so we can keep an eye. Other option would be to reduce frequency so we still run it periodically to see if a new image with a fix comes along...?

I think it's fine to turn it off for now. I will first try running in VM on my machine. We can re-evaluate after that if we want to try periodically and hope for the best :crossed_fingers:

.github/workflows/test_suite_windows.yml
```diff
@@ -3,25 +3,26 @@ name: TestSuiteWindows
...
+  #     - '**.c'
+  #     - '**.cpp'
+  #     - '**.fypp'
+  #     - '**.f90'
+  #     - '**.F90'
+  #     - '**.pf'
+  #     - '**.py'
+  #     - '**CMakeLists.txt'
+  #     - '**requirements.txt'
+  #     - '**data/*'
```
TomMelt, line 25: I haven't done this before. Does GitHub have defaults? will this actually disable the CI or just change our configuration?

Hope my question makes sense. lmk if not.

jwallwork23: I think it does disable the CI. The job doesn't show up on the checks run. Also note that I left `workflow_dispatch` as enabled so I think this is basically just a manual-trigger only setup.

jwallwork23: Can confirm it looks okay on #317 rebased on top of `main` after merging this.

PR #291, opened 2025-02-25
Tensor manipulation demo

jwallwork23:
Closes #261.

Adds an introductory example to demonstrate how to construct, interrogate, manipulate, and delete `torch_tensor`s.

PR #317, opened 2025-03-10
Reorder examples

joewallwork:
FTorch's examples have been added with a sequential numbering that suggests which order they should be done, but is actually more to do with the order they were written. I think if we're going to continue having numbered examples then this would make for a more logical numbering:

0. Tensor manipulation (upcoming, see #291)
1. Simplenet
2. ResNet
3. Multi-I/O
4. Looping
5. Multi-GPU
6. MPI
7. Autograd

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -21,7 +21,7 @@ This directory contains a number of examples of how to use the library:
       in of a net from the call to the forward pass.
 
 5. MultiGPU
-	  - Revisits the SimpleNet example but considering multiple GPUs.
+    - Revisits the SimpleNet example but considering multiple GPUs.
 
 6. MPI
    - Demonstrates how to use FTorch in the context of an MPI program.


jatkinson1000 (commented):
Discussed in-person and you've already incorporated my feedback here.

Numbering starts at 1 though, surely 😉

jwallwork23:
> Discussed in-person and you've already incorporated my feedback here.
> 
> Numbering starts at 1 though, surely 😉

Haha true. We can bump the numbers upon merging #291 I guess.

Sorry, hadn't noticed your review and rebased on top of `main` to check that #315 worked as expected.

examples/README.md
```diff
@@ -10,27 +10,27 @@ This directory contains a number of examples of how to use the library:
...
-5. Looping
+4. Looping
     - In most applications we do not just call a net once, but many times.
       To do this in a computationally efficient manner requires some thought.
       This example demonstrates how to structure code in these cases, separating reading
       in of a net from the call to the forward pass.
 
-6. Autograd
+5. MultiGPU
+	  - Revisits the SimpleNet example but considering multiple GPUs.
```
jatkinson1000, line 24:
```suggestion
    - Revisits the SimpleNet example but considering multiple GPUs.
```

jatkinson1000 (approved):
Typo suggestion, rest LGTM @jwallwork23

PR #319, opened 2025-03-10
Add tests for constructing after destructing

### LINKED ISSUE #318: Tests for calling destructor and then constructor
We have unit tests that check that the `torch_tensor_delete` destructor works when called manually or automatically via the finalizer. However, we do not consider the case where a user calls the destructor and then applies a constructor to the same tensor. @ma595 showed an example where this could happen in a loop over vertical levels, for example.

jwallwork23:
Closes #318.

This test extends the fixtures in the existing unit tests for the constructors and destructors to cover the case where `torch_tensor_delete` is called and then a constructor is applied to the same tensor.

### DIFF
--- a/test/unit/test_tensor_constructors_destructors.pf
+++ b/test/unit/test_tensor_constructors_destructors.pf
@@ -45,20 +45,23 @@ contains
     test_case_ctor%param = param
   end function test_case_ctor
 
-  ! A fixture comprised of a full list of parameter sets
-  function get_parameters_full() result(params)
+  ! A fixture comprised of parameter sets for destructor tests
+  function get_parameters_destruction() result(params)
     type(TestParametersType), allocatable :: params(:)
     params = [ &
       TestParametersType(.false.,.false.,1), &
       TestParametersType(.false.,.false.,2), &
       TestParametersType(.false.,.true.,1), &
-      TestParametersType(.false.,.true.,2), &
-      TestParametersType(.true.,.false.,1), &
-      TestParametersType(.true.,.false.,2), &
-      TestParametersType(.true.,.true.,1), &
-      TestParametersType(.true.,.true.,2) &
+      TestParametersType(.false.,.true.,2) &
     ]
-  end function get_parameters_full
+  end function get_parameters_destruction
+
+  ! A fixture comprised of a short list of parameter sets
+  function get_parameters_requires_grad() result(params)
+    type(TestParametersType), allocatable :: params(:)
+    params = [TestParametersType(.false.,.false.,1)]
+    params = [TestParametersType(.true.,.false.,1)]
+  end function get_parameters_requires_grad
 
   ! A fixture comprised of a short list of parameter sets
   function get_parameters_short() result(params)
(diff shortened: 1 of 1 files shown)


test/unit/test_tensor_constructors_destructors.pf
```diff
@@ -22,8 +22,9 @@ module test_tensor_constructors_destructors
   ! Typedef holding a set of parameter values
   @testParameter
   type, extends(AbstractTestParameter) :: TestParametersType
-    logical :: requires_grad
-    logical :: auto_delete
+    logical :: requires_grad  ! Value used for the requires_grad argument
+    logical :: auto_delete    ! torch_tensor_delete is called when .false., otherwise the finalizer
```
jatkinson1000, line 26: Otherwise the finalizer what?

test/unit/test_tensor_constructors_destructors.pf
```diff
@@ -44,37 +45,32 @@ contains
...
-      TestParametersType(.false.,.false.), &
-      TestParametersType(.true.,.false.), &
-      TestParametersType(.false.,.true.), &
-      TestParametersType(.true.,.true.) &
+      TestParametersType(.false.,.false.,1), &
+      TestParametersType(.false.,.false.,2), &
+      TestParametersType(.false.,.true.,1), &
+      TestParametersType(.true.,.false.,1), &
+      TestParametersType(.true.,.false.,2), &
+      TestParametersType(.true.,.true.,1) &
```
jatkinson1000, line 58: Oh, pFUnit, this is a horrible way to do this 😅

test/unit/test_tensor_constructors_destructors.pf
```diff
@@ -91,29 +87,34 @@ contains
...
-    end if
+      if (.not. this%param%auto_delete) then
+        ! Call torch_tensor_delete manually
+        call torch_tensor_delete(tensor)
+
+        ! Check torch_tensor_delete does indeed free the memory
+        @assertFalse(c_associated(tensor%p))
+      end if
+
+    end do
```
jatkinson1000, line 112: I see this tests manual deletion, but I am not sure this test actually tests automatic deletion as the docstring on ll 77-78 implies to me?

jatkinson1000 (commented):
Thanks @jwallwork23 Generally looks sensible, couple of queries in there which may just be me not knowing pFUnit as well as you do.

test/unit/test_tensor_constructors_destructors.pf
jwallwork23: Yup

test/unit/test_tensor_constructors_destructors.pf
jwallwork23: Apologies, should be clearer in 1644b6f.

test/unit/test_tensor_constructors_destructors.pf
jwallwork23: Well it at least checks that the finalizer doesn't raise an error. I'm not sure how else we could test it, given Fortran doesn't support arbitrary scopes like C does.

Expanded the checks in 8e84109.

jwallwork23:
Simplified unit testing in a5db1d6.

test/unit/test_tensor_constructors_destructors.pf
```diff
@@ -75,14 +78,39 @@ contains
...
+
+    ! Check the tensor pointer is not associated
+    @assertFalse(c_associated(tensor%p))
+
+    ! Create a tensor without any data values assigned
+    call torch_tensor_empty(tensor, ndims, tensor_shape, dtype, device_type, device_index, &
+                            this%param%requires_grad)
+
+    ! Check the tensor pointer is associated
+    @assertTrue(c_associated(tensor%p))
```
jatkinson1000, line 102: I wonder if it's worth testing anything else other than whether a pointer is associated? e.g. shape rank?

jatkinson1000 (approved):
Thanks @jwallwork23 that is a lot clearer I think.
And I like the improved fixtures use - I had been wondering if something like that was possible actually!

Will approve, only final pedantic point is consider moving `test_torch_tensor_delete` to the end of the tests rather than between `empty` and `zeros` - to me the tests for empty, zeroes, and ones should be together. Matter of taste and not going to make it blocking, however.

test/unit/test_tensor_constructors_destructors.pf
jwallwork23: I started implementing this but then realised it's already done in `test_tensor_interrogation.pf`.

jwallwork23:
> Will approve, only final pedantic point is consider moving `test_torch_tensor_delete` to the end of the tests rather than between `empty` and `zeros` - to me the tests for empty, zeros, and ones should be together. Matter of taste and not going to make it blocking, however.

Good point. Done in https://github.com/Cambridge-ICCS/FTorch/pull/319/commits/bdcf3cd2947059310f5726ed7480a395f2a0514a.

jatkinson1000:
Thanks, feel free to merge.

PR #320, opened 2025-03-12
Add optimisers

jatkinson1000:
This PR seeks to add optimisers functionality to the code.

- [ ] FTorch
    - [ ] Expose optimisers in FTorch
    - [ ] Write unit tests to cover these
- [ ] Exercise:
    - [ ] README
    - [ ] requirements
    - [x] python version of the exercise
    - [ ] Fortran version of the exercise

Notes whilst in progress:

- It may be useful to expose a few more tensor functions such as `.mean()`
  - it is handy to implicitly initialise a gradient of ones in `.backward()` by contracting a (loss) tensor into a scalar. Should this be added to the autograd example?
- `ftorch.F90` is going to grow with this. Perhaps now is the time to break apart into sub-modules?


.github/workflows/static_analysis.yml
```diff
@@ -100,6 +100,7 @@ jobs:
           cd ${{ github.workspace }}
           . ftorch_venv/bin/activate
           fortitude check src/ftorch.F90
+          fortitude check src/ftorch_optim.F90
```
jwallwork23, line 103: At some point it might make sense to separate out modules for `tensor`, `model`, `optim`.

jatkinson1000: Yeah, since each optimiser (Adam, SGD, etc.) seems to be its own function I thought I'd do this here, for the Fortran at least.
I had been hoping that there was a general optimiser function specified by an enum, but alas no.

Others feel OK for now, but definitely something I was thinking about, especially as module functions might grow soon.

jwallwork23: Agreed. This was more in anticipation of when the torch model/module API starts growing.

jatkinson1000:
Starting with trying to simply bring across SGD.

There is a `torch::optim::Optimizer` class, but it is never directly used, instead being subclassed by `torch::optim::SGD` etc.
For now I will just try and bring over `torch::optim::SGD`, and we can see if it is possible to generalise in future.

There is an associated `torch::optim::SGDOptions` class (subclassed from `torch::optim::OptimizerOptions` (or maybe `torch::optim::OptimizerCloneableOptions`??)) that is passed into the SGD class to provide options.

The other required input is a list of tensors: `std::vector<torch::Tensor>` which I will use in the early examples.
More generally we can use `torch::nn::Module::parameters()` to provide this list, so we probably want to expose that method for models somewhere down the line.

We will need to expose the `.zero_grad()` method and the `.step()` method which are common for all optimizers.

Still not clear what unit tests should be.
My plan is to write some C++ code, then move to write using ctorch code, then to FTorch code.

jatkinson1000:
> The other required input is a list of tensors: `std::vector<torch::Tensor>` which I will use in the early examples.
More generally we can use `torch::nn::Module::parameters()` to provide this list, so we probably want to expose that method for models somewhere down the line.
  - On reflection I think we want to make the default input a `vector`, and perhaps handle enforcing this on the FTorch side if possible?
  - For model forward we require a user to supply tensors as an array of tensors for which each `IValue` in C++ is pushed on to a vector. This can also be implemented.
  - For model parameters we could either pass around a pointer to a vector of tensors as arises from the C++ `.parameters` and check the type, or manually extract into an array of tensors in Fortran. The former might be 'cleaner', but the latter would be useful should we ever eant to probe things on ther Fortran side...

jatkinson1000:
Checkpointing here:

- Functions for:
  - Creating SGD optimizer
  - stepping an optimizer
  - zeroing gradients associated with an optimizer
- Python and fortran examples to 'train' a single tensor to scale an input vector

Further steps:

- [ ] Implement more optimizers beyond SGD
- [ ] Tests for optimizer functionalities
- [ ] Better generation of input `parameters`
  - [ ] Including getting outputs of a model, though this may be another PR

.github/workflows/static_analysis.yml
jatkinson1000: This has come up for me just now.
I looked at adding `torch_optim_delete` to the generic `torch_delete` interface, but cannot as this would create a cyclic module dependency (I use `ftorch_int` in the optimizers).

Perhaps the route would be for a 'parent' module that provides interfaces and key routines from sub-modules for specific items. Also perhaps a constants module.

I'm pausing this for now, but we can perhaps discuss.
Again, would unfortunately be a breaking API change I think.

jwallwork23: If we were to have modules `ftorch_tensor`, `ftorch_model`, `ftorch_optim` then we could avoid an API break by providing interfaces to items from the different modules inside a core `ftorch` module, no? The `torch_delete` interface would be an example of something living in the core module and be defined based off `torch_tensor_delete`, `torch_model_delete`, and `torch_optim_delete` from the three modules, resp..

jatkinson1000: Yes, would be verbose, but should work, I think...
Perhaps something for the hackathon?
I guess this is the challenge of maintaining vs. developing software.

jwallwork23: I think it'd be a good hackathon exercise, yeah. I wouldn't have thought it'd be too verbose but maybe I'm missing something.

jwallwork23:
> A useful thing to note is that if we want to set values of a tensor (e.g. loss, or a gradient) it must NOT have `requires_grad=.true.` or it will produce a segfault.
> 
> This has caught me out a couple of times, so probably need some clear docs as to where users should and should not be enabling gradient calculations.

Good to know. It seems you need to use `requires_grad=.true.` only for the arrays you want to differentiate with respect to.

examples/n_Optimizers/optimizers.f90
```diff
@@ -0,0 +1,115 @@
...
+
+    ! Create an empty loss tensor and populate with mean square error (MSE) between target and input
+    ! Then perform backward step on loss to propogate gradients using autograd
+    !
+    ! We could use the following lines to do this by explicitly specifying a
+    ! gradient of ones to start the process:
+    call torch_tensor_empty(loss, ndims, tensor_shape, &
+                           torch_kFloat32, torch_kCPU)
+    call torch_tensor_empty(scaling_grad, ndims, tensor_shape, &
+                           torch_kFloat32, torch_kCPU)
```
jatkinson1000, line 74: Some strange behaviour that needs digging into here.
This code seems to work, but I create a new tensor to store the gradient here every time.

If I take this outside the loop we get segmentation faults with the gradient seeming to be undefined when I subsequently try and fetch it.

The gradient of `scaling_tensor` should be intrinsically attached to the tensor itself and not affected by `torch_tensor_get_gradient` or anything else other than `zero_grad` which should set its values to `0.0`, not leave it undefined.
I tried changing `get_gradient` to take a `.clone()` in the cpp but this did not help.
It seems strange that `zero_grad` would do this since it is just a direct call to the Optimizer object method. I think a fresh pair of eyes a different day is needed for now.

jwallwork23:
FYI I rebased this branch on top of `main` on [`joe/optim`](https://github.com/Cambridge-ICCS/FTorch/tree/joe/optim) in case it's useful. Really cool functionality!

examples/n_Optimizers/optimizers.f90
jwallwork23: Hm I've seen similar issues while working on #336 so will take a look.

jwallwork23: On branch [`joe/optim`](https://github.com/Cambridge-ICCS/FTorch/tree/joe/optim) I found that we can pull the tensor constructor calls for the loss tensor and scaling tensor outside the loop if we pass `true` for the `retain_graph` argument in the backward operator. I'll open a separate PR to enable this.

jatkinson1000:
Rebased on main

src/ftorch_optim.F90
```diff
@@ -0,0 +1,182 @@
...
+        type(c_ptr), value, intent(in) :: optim_c
+      end subroutine torch_optim_zero_grad_c
+    end interface
+
+    call torch_optim_zero_grad_c(optim%p)
+  end subroutine torch_optim_zero_grad
+
+  !> Step a Torch optimizer
+  subroutine torch_optim_step(optim)
+    type(torch_optim), intent(in) :: optim  !! Optimizer to step
```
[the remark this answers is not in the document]
jatkinson1000: This has been done (and updated in pFUnit tests and examples) in https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/65d51eaaaa86caba37cad2bb70fa254b644f2517

jatkinson1000:
Note: Found during this that one cannot use `tensor_a = tensor_b(requires_grad=.true.)` when `tensor_a` is only defined as a `torch_tensor`. This is because the `required_grad` property gets copied across, and then we get errors resulting from trying to modify the values of a tensor with `requires_grad=.true.`. The solution is to declare `tensor_a` as empty.

See pFUnit tests for optimizers, specifically the zero_grad test.

cc @jwallwork23

examples/CMakeLists.txt
```diff
@@ -17,4 +17,5 @@ if(CMAKE_BUILD_TESTS)
     add_subdirectory(7_MPI)
   endif()
   add_subdirectory(8_Autograd)
+  add_subdirectory(n_Optimizers)
```
joewallwork:
```suggestion
  add_subdirectory(9_Optimizers)
```

examples/10_Optimizers/CMakeLists.txt
joewallwork, line 1: Subdir needs renaming to `9_Optimizers`

examples/n_Optimizers/README.md
```diff
@@ -0,0 +1,116 @@
+# Example n - Optimizers
```
joewallwork:
```suggestion
# Example 9 - Optimizers
```

src/ctorch.cpp
```diff
@@ -539,6 +539,142 @@ void torch_tensor_get_gradient(const torch_tensor_t tensor, torch_tensor_t gradi
...
+
+// Check if an optimizer is valid
+void validate_optimizer(const torch_optim_t optim, const std::string &name) {
+  if (!optim) {
+    throw std::invalid_argument(name + " is null.");
+  }
+}
+
+// Function to create an SGD optimizer and return a pointer to it
+// TODO: Wrap additional Options beyond learning rate and momentum
```
joewallwork: It looks like this has already been done? There's already weight decay, too?

src/ctorch.cpp
```diff
@@ -539,6 +539,142 @@ void torch_tensor_get_gradient(const torch_tensor_t tensor, torch_tensor_t gradi
...
+    auto optimizer_SGD = new torch::optim::SGD(parameters_vec, options);
+    return static_cast<torch_optim_t>(optimizer_SGD);
+  } catch (const std::exception &e) {
+    ctorch_error(std::string(e.what()) + " in torch_optim_SGD");
+  }
+  return nullptr; // Return null in case of error
+}
+
+// Function to create an Adam optimizer and return a pointer to it
+// TODO: Wrap additional Options beyond learning rate
```
joewallwork: Here too?

src/ctorch.cpp
```diff
@@ -539,6 +539,142 @@ void torch_tensor_get_gradient(const torch_tensor_t tensor, torch_tensor_t gradi
...
+    auto optimizer_Adam = new torch::optim::Adam(parameters_vec, options);
+    return static_cast<torch_optim_t>(optimizer_Adam);
+  } catch (const std::exception &e) {
+    ctorch_error(std::string(e.what()) + " in torch_optim_Adam");
+  }
+  return nullptr; // Return null in case of error
+}
+
+// Function to create an AdamW optimizer and return a pointer to it
+// TODO: Wrap additional Options beyond learning rate
```
joewallwork: And here?

test/unit/optim/test_optim.py
joewallwork, line 1: It'd be nice to use Pytest here. The syntax would be more concise and easily extensible and it'd be possible to run more tests after one fails.

examples/CMakeLists.txt
jatkinson1000: Should now be number 10 in https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/789598dc8d56864f9936e6fc1fae8c2e9820f598

examples/10_Optimizers/CMakeLists.txt
jatkinson1000: Should now be named 10 in https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/789598dc8d56864f9936e6fc1fae8c2e9820f598

examples/n_Optimizers/README.md
jatkinson1000: Now named 10 in https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/789598dc8d56864f9936e6fc1fae8c2e9820f598

examples/CMakeLists.txt
```diff
@@ -1,21 +1,22 @@
...
-  add_subdirectory(4_Batching)
-  add_subdirectory(5_MultiIO)
-  add_subdirectory(6_Looping)
+  add_subdirectory(01_Tensor)
+  add_subdirectory(02_SimpleNet)
+  add_subdirectory(03_ResNet)
+  add_subdirectory(04_Batching)
+  add_subdirectory(05_MultiIO)
+  add_subdirectory(06_Looping)
   # NOTE: Example 6_Looping is built here but not run as part of the integration
```
joewallwork:
```suggestion
  # NOTE: Example 06_Looping is built here but not run as part of the integration
```

pages/usage/online.md
```diff
@@ -1,17 +1,68 @@
...
+
+#### 2. pt2ts
+
+The scripting section comes earlier in the online workflow. Having written a
+model to a file with `.pt` extension, use the `pt2ts.py` utility Python script
+to convert it to TorchScript format. A template `pt2ts.py` script can be found
+in the [`utils`](https://github.com/Cambridge-ICCS/FTorch/tree/main/utils)
+subdirectory. See the
+[README](https://github.com/Cambridge-ICCS/FTorch/blob/main/utils/README.md)
+there for more details on how to use the script.
```
joewallwork, line 37: Note to self to update this in #555

src/ctorch.cpp
```diff
@@ -537,6 +537,142 @@ void torch_tensor_get_gradient(const torch_tensor_t tensor, torch_tensor_t gradi
   }
 }
 
+// =============================================================================
+// --- Torch optimisers API
```
joewallwork: It seems this is your preference
```suggestion
// --- Torch optimizers API
```

pages/usage/online.md
```diff
@@ -150,13 +201,32 @@ gradient for the same tensor is being extracted into the same array. This is due
...
+follows:
+
+* `torch_optim%zero_grad` ([[ftorch_optim(module):torch_optim_zero_grad(subroutine)]]),
+  which zeroes all tensors associated with the
+  optimizer. This should be called at the beginning of every step of the
+  optimization loop.
+* `torch_optim%step` ([[ftorch_optim(module):torch_optim_step(subroutine)]]),
+  which takes an iteration of the optimization method.
+
+The [optimizer worked example](https://github.com/Cambridge-ICCS/FTorch/tree/main/examples/n_Optimizers)
```
joewallwork:
```suggestion
The [optimizer worked example](https://github.com/Cambridge-ICCS/FTorch/tree/main/examples/10_Optimizers)
```

pages/usage/online.md
```diff
@@ -150,13 +201,32 @@ gradient for the same tensor is being extracted into the same array. This is due
...
 
-<!- Write as part of the API Documentation Pages -->
-*Not yet implemented.*
+The loss function classes defined in PyTorch/LibTorch have not yet been exposed
+in FTorch. However, the [[ftorch_tensor(module):torch_tensor_sum(subroutine)]] and
+[[ftorch_tensor(module):torch_tensor_mean(subroutine)]] reduction
+operators have been provided, which should be sufficient for simple loss
+functions such as the mean-square-error (MSE).
+
+See the [optimizer worked example](https://github.com/Cambridge-ICCS/FTorch/tree/main/examples/n_Optimizers)
```
joewallwork:
```suggestion
See the [optimizer worked example](https://github.com/Cambridge-ICCS/FTorch/tree/main/examples/10_Optimizers)
```

test/unit/optim/unittest_optim.pf
```diff
@@ -0,0 +1,337 @@
...
+    ! Repeat the operation but check gradients no longer stacked (back to 2*tensor)
+    tensor_sq = tensor ** 2
+    call torch_tensor_backward(tensor_sq, retain_graph=.true.)
+    call torch_tensor_get_gradient(grad, tensor)
+    @assertTrue(assert_allclose(grad_data, 2.0*tensor_data, test_name="test_torch_optim_zero_grad_3"))
+
+    ! Clean up
+    call torch_delete(tensor)
+    call torch_delete(tensor_sq)
+    call torch_delete(grad)
```
joewallwork, line 191: Do we still want to include these? (Here and elsewhere)

src/ftorch_optim.f90
```diff
@@ -0,0 +1,286 @@
...
+    end if
+
+    if (.not. present(beta_2)) then
+      beta_2_value = 0.999_real64
+    else
+      beta_2_value = beta_2
+    end if
+
+    if (.not. present(weight_decay)) then
+      weight_decay_value = 0.01_real64
```
joewallwork: I feel that non-obvious default values such as these should be documented in the docstring.

src/ctorch.cpp
jatkinson1000: There are [several more options available](https://docs.pytorch.org/cppdocs/api/structtorch_1_1optim_1_1_adam_options.html#exhale-struct-structtorch-1-1optim-1-1-adam-options), but at diminishing returns.

However, Matt suggests exposing the full list if we can so I'll do that in a subsequent commit.

src/ctorch.cpp
jatkinson1000: There are [several more options available](https://docs.pytorch.org/cppdocs/api/structtorch_1_1optim_1_1_s_g_d_options.html#exhale-struct-structtorch-1-1optim-1-1-s-g-d-options), but at diminishing returns.

However, Matt suggests exposing the full list if we can so I'll do that in a subsequent commit.

src/ctorch.cpp
jatkinson1000: There are [several more options available](https://docs.pytorch.org/cppdocs/api/structtorch_1_1optim_1_1_adam_w_options.html#exhale-struct-structtorch-1-1optim-1-1-adam-w-options), but at diminishing returns.

However, Matt suggests exposing the full list if we can so I'll do that in a subsequent commit.

src/ctorch.cpp
jatkinson1000: See https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/e211295aca37806c98920b7c690567f5dae90cb1 through https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/c0ecc3e661b4775cc667371de587a863075be03f

src/ctorch.cpp
jatkinson1000: See https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/e211295aca37806c98920b7c690567f5dae90cb1 through https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/c0ecc3e661b4775cc667371de587a863075be03f

src/ctorch.cpp
jatkinson1000: See https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/e211295aca37806c98920b7c690567f5dae90cb1 through https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/c0ecc3e661b4775cc667371de587a863075be03f

src/ctorch.cpp
jatkinson1000: See https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/db88f84e07ce11da8878c28cb6a04c4ad9e6175d, also in the header file.

test/unit/optim/unittest_optim.pf
jatkinson1000: Did we settle on guidance in the examples?
I can see an argument either way, happy to remove to keep tests clearer and more concise.

src/ftorch_optim.f90
jatkinson1000: Agreed, will do that to hopefully be close to wrapping this up!
Values are selected to match PyTorch.

jatkinson1000: Done in https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/9b484a9148dbd1f61c312468293e141aad4f4d37

examples/10_Optimizers/README.md
```diff
@@ -0,0 +1,117 @@
+# Example 10 - Optimizers
+
+**This example is currently under development.** Eventually, it will demonstrate
```
joewallwork: Can this disclaimer be dropped now?

examples/10_Optimizers/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+By exposing optimizers in Fortran, FTorch will be able to compute optimisation
+steps to update models as part of a training process.
+
+## Description
+
+A Python demo is copied from the PyTorch documentation as `optimizers.py`, which
+shows how to use an optimiser in PyTorch.
+
+The demo will be replicated in Fortran as `optimizers.f90`, to show how to do the
```
joewallwork:
```suggestion
The demo is replicated in Fortran as `optimizers.f90`, to show how to do the
```

pages/usage/optimizers.md
```diff
@@ -0,0 +1,124 @@
...
+
+Whilst different subroutines exist for creating different kinds of optimizer, they all
+have common core methods that are used.
+
+##### Zero Gradients
+
+[[ftorch_optim(module):torch_optim_zero_grad(subroutine)]] clears the gradients of all
+parameters managed by the optimizer. This should be called at the beginning of each
+iteration during training.
+The method is implemented as a bound procedure on the
```
joewallwork:
```suggestion
The method is implemented as a procedure bound to the
```

pages/usage/optimizers.md
```diff
@@ -0,0 +1,124 @@
...
+`optimizer%zero_grad()`.
+
+##### Step
+
+[[ftorch_optim(module):torch_optim_step(subroutine)]] performs a single optimization
+step, updating all parameters managed by the optimizer based on their gradients.
+It should be called after backpropogation has been performed following a forward pass
+during a training iteration.
+
+The method is implemented as a bound procedure on the
```
joewallwork:
```suggestion
The method is implemented as a procedure bound to the
```

test/unit/optim/test_optim_sgd.py
joewallwork, line 1: A docstring would be helpful to explain what this is for. It seems to be fore generating the expected values for the Fortran tests.

joewallwork: Similarly for the others.

test/unit/optim/unittest_optim.pf
joewallwork: I think leave for now and we can follow up to do that across all examples and tests later. This PR is already rather large.

test/unit/optim/unittest_optim_adam.pf
joewallwork, line 1: I like the way you defined the parameter sets and use the description in the `toString`.

jatkinson1000: Thanks! The idea came to me when writing: https://jackatkinson.net/post/fortran_parameterizing_pfunit/

test/unit/optim/test_optim_sgd.py
jatkinson1000: See https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/99b1f1fbd715b1d3e62b82b6358a5b23fa6c16d2

pages/usage/optimizers.md
jatkinson1000: See https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/b46fb8206f97d9179e4f81bb9285f20428c77008

pages/usage/optimizers.md
jatkinson1000: See https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/b46fb8206f97d9179e4f81bb9285f20428c77008

examples/10_Optimizers/README.md
jatkinson1000: See https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/d9d29e6bc0548d72e9a14379a18a9dd2f677439e

examples/10_Optimizers/README.md
jatkinson1000: See https://github.com/Cambridge-ICCS/FTorch/pull/320/commits/d9d29e6bc0548d72e9a14379a18a9dd2f677439e

PR #324, opened 2025-03-14
Typo correction of is_training variable description in the ftorch API.

jatkinson1000:
As per the description, I spotted this on a stroll through ftorch.F90 so thought I'd fix.

Seems to have been a copy-paste error.


niccolozanotti (approved):
Yes, that does look like a typo.

PR #326, opened 2025-03-17
Mention test changes in changelog

jwallwork23:
I forgot to mention the changes that came with #319 in the newly added changelog.

Going forward, we should add changes to the changelog in the PR that introduces them, where possible.


jatkinson1000 (approved):
Haha, I did actually think of requesting this, but decided that were maybe small enough internal changes that they could be skipped 

Definitely should be part of a PR, and I'll be enforcing this in reviews going forward 😜

PR #330, opened 2025-03-18
Ignore `*.DS_Store` and fix `pre-commit` hook path in the docs

niccolozanotti:
As per title, fixes two minor issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -196,4 +196,4 @@ cython_debug/
 *.pt
 
 # Apple Desktop Services Store
-*.DS_Store
+**.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -194,3 +194,6 @@ cython_debug/
 
 # TorchScript model
 *.pt
+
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -194,3 +194,6 @@ cython_debug/
 
 # TorchScript model
 *.pt
+
+# Apple Desktop Services Store
+*.DS_Store
```
jatkinson1000, line 199: IIRC this will ignore anything ending in .DS_Store in the top level directory, but `**` will ignore in every directory. see https://git-scm.com/docs/gitignore#_pattern_format

```suggestion
**.DS_Store
```

jatkinson1000 (changes requested):
Yes, thanks for spotting this @niccolozanotti 

I have a suggestion on the globbing in .gitignore, but am otherwise happy.

niccolozanotti:
Happy to bring pair of fresh 👀 to the project.

PR #331, opened 2025-03-19
Add `fypp` to requirements 

### LINKED ISSUE #329: Add fypp to requirements
@niccolozanotti and I realised that the `requirements.txt` file is missing `fypp`. We were discussing whether it could be separated out into user vs. developer dependencies like Python packages often are nowadays, but then realised that all the dependencies in `requirements.txt` are already developer ones anyway. Not sure if this is mentioned in the docs but perhaps it should be.

niccolozanotti:
This PR
- renames `requirements.txt`  to `requirements-dev.txt` to clarify these are developer dependencies rather than user ones as one might have been tricked to think following the naming convention;
- adds pip-installable `fypp` to the requirements and mentions that in the docs.
 
Closes #329.


jwallwork23 (approved):
Reads well to me, thanks for this.

pages/developer.md
jatkinson1000, line 1: I would perhaps be in favour of consolidating the install instructions into one place under an "Installing developer requirements" header, explaining that developers should install these tools, and then refer to this instead of repeating the installation instructions twice.

jatkinson1000 (commented):
Thanks @niccolozanotti I think this is fine as you have written, but consider the suggested restructure to perhaps make things a bit more concise.

pages/developer.md
niccolozanotti: [4401f2a](https://github.com/Cambridge-ICCS/FTorch/pull/331/commits/4401f2a7066c8ec2ab5c46d384ac6db2b52758ba)  is supposed to reflect those changes.

jatkinson1000 (approved):
Thanks @niccolozanotti this looks great.

Made a couple of non-blocking style suggestions (I like slightly longer clickable links) but am approving this for you.

PR #333, opened 2025-03-19
Add news section to website

jwallwork23:
As discussed the other day, I thought it'd be a good idea to add a News section to the website mentioning some recent/upcoming items.


jatkinson1000 (approved):
Yes, this looks sensible to me.

We'll just need to make sure we keep on top of clearing out the old things and adding new ones.
Do you think it would be worth having a news 'archive' page for past things?

jwallwork23:
> Yes, this looks sensible to me.
> 
> We'll just need to make sure we keep on top of clearing out the old things and adding new ones. Do you think it would be worth having a news 'archive' page for past things?

Opened #334.

PR #335, opened 2025-03-24
Fix static analysis workflow (failing after #331)

niccolozanotti:
* Renames references to requirements.txt to requirements-dev.txt (needed after #331)
* Adds requirements-dev.txt to triggering changed paths


jatkinson1000 (approved):
🔥 🚒 
Thanks @niccolozanotti I think that fixes everything needed.

Issue #336, opened 2025-03-26
Implement zeroing of gradients


jwallwork23:
Following @jatkinson1000's question about what happens when we run backpropagation on different tensors in https://github.com/Cambridge-ICCS/FTorch/pull/286#discussion_r1998305043, I did some digging.

I think this PyTorch code encapsulates things nicely (print outputs showed as comments):
```python
import torch

a = torch.tensor([1.0], requires_grad=True)
b = torch.tensor([2.0], requires_grad=True)

Q = a * b
print(f"Q = a * b = {Q}")
# Q = a * b = tensor([2.], grad_fn=<MulBackward0>)

P = a * b
print(f"P = a * b = {P}")
# P = a * b = tensor([2.], grad_fn=<MulBackward0>)

Q.backward(torch.ones((1,)), retain_graph=True)
print(f"dQda = {a.grad} ?= b = {b}")
print(f"dQdb = {b.grad} ?= a = {a}")
# dQda = tensor([2.]) ?= b = tensor([2.], requires_grad=True)
# dQdb = tensor([1.]) ?= a = tensor([1.], requires_grad=True)

P.backward(torch.ones((1,)), retain_graph=True)
print(f"dPda = {a.grad} ?= b = {b}")
print(f"dPdb = {b.grad} ?= a = {a}")
# dPda = tensor([4.]) ?= b = tensor([2.], requires_grad=True)
# dPdb = tensor([2.]) ?= a = tensor([1.], requires_grad=True)

R = a * b
print(f"R = a * b = {R}")
# R = a * b = tensor([2.], grad_fn=<MulBackward0>)

R.backward(torch.ones((1,)), retain_graph=True)
print(f"dRda = {a.grad} ?= b = {b}")
print(f"dRdb = {b.grad} ?= a = {a}")
# dRda = tensor([6.]) ?= b = tensor([2.], requires_grad=True)
# dRdb = tensor([3.]) ?= a = tensor([1.], requires_grad=True)

a.grad.zero_()
b.grad.zero_()

S = a * b
print(f"S = a * b = {S}")
# S = a * b = tensor([2.], grad_fn=<MulBackward0>)

S.backward(torch.ones((1,)), retain_graph=True)
print(f"dSda = {a.grad} ?= b = {b}")
print(f"dSdb = {b.grad} ?= a = {a}")
# dSda = tensor([2.]) ?= b = tensor([2.], requires_grad=True)
[...]

PR #338, opened 2025-03-26
Implement `torch_tensor_zero`

jwallwork23:
While working on #336, I wrapped the `Tensor.zero_` method and thought it might be useful.

To support this, I added a unit test module for tensor manipulation.

### DIFF
--- a/pages/tensor.md
+++ b/pages/tensor.md
@@ -71,6 +71,12 @@ this can also be applied to arrays of tensors. Calling this subroutine manually
 is optional as it is called as a destructor when the `torch_tensor` goes out of
 scope anyway.
 
+### Tensor manipulation
+
+We provide a subroutine for manipulating the data values associated with a
+`torch_tensor` object: `torch_tensor_zero_`. It has an alias as a class method:
+`torch_tensor%zero_`.
+
 ### Operator overloading
 
 Mathematical operators involving Tensors are overloaded, so that we can compute
--- a/pages/tensor.md
+++ b/pages/tensor.md
@@ -74,8 +74,8 @@ scope anyway.
 ### Tensor manipulation
 
 We provide a subroutine for manipulating the data values associated with a
-`torch_tensor` object: `torch_tensor_zero_`. It has an alias as a class method:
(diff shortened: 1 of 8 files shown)


jatkinson1000 (commented):
All seems sensible, quick thought is that we might want to drop the trailing `_` from the Fortran API.

I think making it a method of the tensor should be sufficient to prevent people confusing this with the subroutine that creates a tensor of zeros from scratch.

pages/tensor.md
```diff
@@ -71,6 +71,12 @@ this can also be applied to arrays of tensors. Calling this subroutine manually
 is optional as it is called as a destructor when the `torch_tensor` goes out of
 scope anyway.
 
+### Tensor manipulation
+
+We provide a subroutine for manipulating the data values associated with a
+`torch_tensor` object: `torch_tensor_zero`. It has an alias as a class method:
+`torch_tensor%zero`.
```
jatkinson1000, line 78: Maybe start a list of manipulations available as we may add more?

src/ctorch.h
```diff
@@ -173,6 +173,16 @@ EXPORT_C bool torch_tensor_requires_grad(const torch_tensor_t tensor);
  */
 EXPORT_C void torch_tensor_delete(torch_tensor_t tensor);
 
+// =====================================================================================
+// --- Functions for manipulating tensors
+// =====================================================================================
+
+/**
+ * Function to reset the values of a Torch Tensor to zero
+ * @param Torch Tensor to zero
```
jatkinson1000, line 182:
```suggestion
 * @param Torch Tensor to zero the values of
```

jatkinson1000 (approved):
Yes, all seems sensible to me @jwallwork23

pages/tensor.md
jwallwork23: Good idea. See f53740a.

jwallwork23: Sure, done in https://github.com/Cambridge-ICCS/FTorch/pull/338/commits/585c03ac1a0bfdca202ba59afbe1cf010317c77e.

PR #341, opened 2025-03-28
Implement zeroing of gradients

### LINKED ISSUE #336: Implement zeroing of gradients

jwallwork23:
Closes #336.

This PR adds a subroutine (and corresponding class method) for zeroing the gradients associated with the tensor.

The caveat that had been tripping me up until now is that you need to call `torch_tensor_get_gradient` after zeroing the gradients. It'd be nice if we could find a way round this, but I guess it's okay for now so long as it's documented.

### DIFF
--- a/pages/autograd.md
+++ b/pages/autograd.md
@@ -77,6 +77,7 @@ call torch_tensor_get_gradient(a, dQda)
 call torch_tensor_from_array(dQdb, out_data3, tensor_layout, torch_kCPU)
 call torch_tensor_get_gradient(b, dQdb)
 ```
+
 #### Zeroing gradients
 
 Having computed gradients of one tensor with respect to its dependencies,
@@ -102,6 +103,12 @@ call torch_tensor_backward(P)
 
 ! ...
 ```
+
+#### Extracting gradients
+
+Note that `torch_tensor_get_gradient` must be called after every call to
+`torch_tensor_backward` or `torch_tensor_zero_grad`, even if the gradient for
+the same tensor is being extracted into the same array.
 
 ### Optimisation
 
--- a/pages/autograd.md
+++ b/pages/autograd.md
@@ -118,7 +118,8 @@ call torch_tensor_backward(P, retain_graph=.true.)
 
 Note that `torch_tensor_get_gradient` must be called after every call to
 `torch_tensor_backward` or `torch_tensor_zero_grad`, even if the gradient for
-the same tensor is being extracted into the same array.
+the same tensor is being extracted into the same array. This is due to the way
+that pointers are handled on the C++ side.
 
 ### Optimisation
 
--- a/pages/autograd.md
+++ b/pages/autograd.md
@@ -78,6 +78,16 @@ call torch_tensor_from_array(dQdb, out_data3, tensor_layout, torch_kCPU)
 call torch_tensor_get_gradient(b, dQdb)
 ```
 
+#### `retain_graph` argument
+
+If you wish to call the backpropagation operator multiple times then you may
+need to make use of the `retain_graph` argument for `torch_tensor_backward`.
+This argument accepts logical values and defaults to `.false.`, for consistency
+with PyTorch and LibTorch. According to the
+[PyTorch docs](https://pytorch.org/docs/stable/generated/torch.Tensor.backward.html),
+`retain_graph=.true.` will not be needed in most cases, but it's useful to have
+for the cases where it is.
+
 #### Zeroing gradients
 
 Having computed gradients of one tensor with respect to its dependencies,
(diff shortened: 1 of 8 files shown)


pages/autograd.md
```diff
@@ -71,15 +71,45 @@ to `a` and/or `b`. To do this, we can use the `torch_tensor_get_gradient`
...
+call torch_tensor_backward(P)
+
+! ...
+```
+
+#### Extracting gradients
+
+Note that `torch_tensor_get_gradient` must be called after every call to
+`torch_tensor_backward` or `torch_tensor_zero_grad`, even if the gradient for
+the same tensor is being extracted into the same array.
```
jatkinson1000, line 111: Might be useful to add a short "because".
Related to changing pointer or data location by these calls on the C++ side?

src/ctorch.h
```diff
@@ -251,6 +251,12 @@ EXPORT_C void torch_tensor_power_float(torch_tensor_t output,
 // --- Functions related to automatic differentiation functionality for tensors
 // =============================================================================
 
+/**
+ * Function to reset the gradient values of a Torch Tensor to zero
+ * @param Torch Tensor to zero the graident values of
```
jatkinson1000, line 256: small typo
```suggestion
 * @param Torch Tensor to zero the gradient values of
```

src/ftorch.fypp
```diff
@@ -840,6 +841,23 @@ contains
   ! --- Procedures related to automatic differentation functionality for tensors
   ! ============================================================================
 
+  !> Resets a tensor's gradient to zero.
+  subroutine torch_tensor_zero_grad(tensor)
+    use, intrinsic :: iso_c_binding, only : c_associated
+    class(torch_tensor), intent(inout) :: tensor
```
jatkinson1000, line 847:
```suggestion
    class(torch_tensor), intent(inout) :: tensor  !! Tensor to zero the gradient of
```

jatkinson1000 (approved):
All generally seems sensible @jwallwork23

"Request changes" seems a bit heavy, but small typo correction and addition of inline FORD docs for you.
Also a small docs addition to explain why users need to get_gradient after each recompute may be useful if you agree.

src/ctorch.h
jwallwork23: Thanks, I always make this typo!

jatkinson1000: 😂 Yeah, I have certain words that get me every time too!

pages/autograd.md
jwallwork23: I think it's because of the way we use `std::move`. I wonder if we can avoid this.

jatkinson1000: Ah, yes, I actually have this in a PR that I am yet to push changes for.

PR #342, opened 2025-03-28
Expose `retain_graph` argument for `backward`

### LINKED ISSUE #340: Expose `retain_graph` argument for `backward`
As mentioned in #320, we need to expose the `retain_graph` optional argument of the backpropagation operator in order to be able to run it multiple times.

jwallwork23:
Closes #340.

> As mentioned in #320, we need to expose the `retain_graph` optional argument of the backpropagation operator in order to be able to run it multiple times.

Note that the branch used in this PR is based off the one from #341 so that one will need to be merged first.

### DIFF
--- a/pages/autograd.md
+++ b/pages/autograd.md
@@ -78,6 +78,13 @@ call torch_tensor_from_array(dQdb, out_data3, tensor_layout, torch_kCPU)
 call torch_tensor_get_gradient(b, dQdb)
 ```
 
+#### `retain_graph` argument
+
+If you wish to call the backpropagation operator multiple times then it's likely
+you will need to make use of the `retain_graph` argument for
+`torch_tensor_backward`. This argument accepts logical values and defaults to
+`.false.`, for consistency with PyTorch and LibTorch.
+
 #### Zeroing gradients
 
 Having computed gradients of one tensor with respect to its dependencies,
@@ -99,7 +106,7 @@ call torch_tensor_backward(Q)
 call torch_tensor_zero_grad(a)
 call torch_tensor_zero_grad(b)
 
-call torch_tensor_backward(P)
+call torch_tensor_backward(P, retain_graph=.true.)
 
 ! ...
 ```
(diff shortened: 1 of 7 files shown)


pages/autograd.md
```diff
@@ -78,6 +78,13 @@ call torch_tensor_from_array(dQdb, out_data3, tensor_layout, torch_kCPU)
 call torch_tensor_get_gradient(b, dQdb)
 ```
 
+#### `retain_graph` argument
+
+If you wish to call the backpropagation operator multiple times then it's likely
+you will need to make use of the `retain_graph` argument for
+`torch_tensor_backward`. This argument accepts logical values and defaults to
+`.false.`, for consistency with PyTorch and LibTorch.
+
```
jatkinson1000, line 87: Making note that you said it is unclear if this is actually the case or not, and that it will work most of the time without.

test/unit/test_tensor_autograd.pf
```diff
@@ -75,4 +75,50 @@ contains
...
+    expected(:,:) = 1.0
+    if (.not. assert_allclose(out_data, expected, test_name="test_torch_tensor_retain_graph1")) then
+      print *, "Error :: incorrect value for first gradient computation"
+      stop 999
+    end if
+
+    ! Zero the gradient and then call back-propagation again and check the computed gradient still
+    ! takes the expected value
+    call a%zero_grad()
+    call torch_tensor_backward(Q, retain_graph=.true.)
```
jatkinson1000, line 115: Given the note that this probably works even when `retain_graph` is not set, have you checked what happens if you remove this from the test? Does it still pass, or does it fail?

jatkinson1000 (approved):
Yes, this is very tidy and assuming the tests are working this seems correct.

I see no reason not to merge, but had a couple of comments/queries around our discussion earlier and when `retain_graph` is or is not required.

test/unit/test_tensor_autograd.pf
jwallwork23: Yeah it still passes.

jatkinson1000: 🙃 Well at least we know it passes if enforced. Not sure there is much we can do, just a curio I had.

pages/autograd.md
jwallwork23: Reworded in ddac55b.

PR #344, opened 2025-03-28
Overload the sum and mean over tensors

### LINKED ISSUE #240: Overload `sum` intrinsic for tensors
We'll need this for computing cost functions on the FTorch side, e.g., taking $\ell_2$ product.

jwallwork23:
Closes #240.

This PR implements `torch_tensor_sum` and `torch_tensor_mean` to support taking the sum and mean over the entries in a tensor.

Unit tests are added to check both the values and gradients are computed correctly. Docs and changelog are updated accordingly.


jatkinson1000 (commented):
Generally seems sensible.

Queries:

- Are sum and mean operator overloads? Or should there be a new tensor_functionalities or similar section?
- Is it worth adding these to the 1st exercise?

jwallwork23:
>     * Are sum and mean operator overloads? Or should there be a new tensor_functionalities or similar section?

I suppose you're right - they aren't overloads. Originally I'd planned to overload the `SUM` intrinsic but decided against it. Reworked in a61095d3478266731d8168bbea0a5c29a3fc786d and 09a7af7cedb25f50c38d39fe77f585b49eb0b526.

>     * Is it worth adding these to the 1st exercise?

Ooh yeah, good idea. I added a mean value computation in e6c8cee1b9a16487ee9f35af7f324c21b2f66ab1, which gives the output
```
1:  Contents of first input tensor:
1:  1  1  1
1:  1  1  1
1: [ CPUFloatType{2,3} ]
1:  Contents of second input tensor:
1:    1.00000000       2.00000000       3.00000000       4.00000000       5.00000000       6.00000000    
1:  Sum of input tensors:
1:    2.00000000       3.00000000       4.00000000       5.00000000       6.00000000       7.00000000    
1:  Mean value:
1:    4.50000000    
1:  Tensor manipulation example ran successfully
```

jwallwork23:
[Rebased on top of latest `main`]

jwallwork23:
Oops, should've checked build locally before pushing. Fixed in 7e002da.

examples/1_Tensor/tensor_manipulation.f90
```diff
@@ -60,9 +64,21 @@ program tensor_manipulation
   ! constructed using `torch_tensor_empty` but it won't be possible to extract its data into an
   ! array.
   c = a + b
-  write(*,*) "Output:"
+  write(*,*) "Sum of input tensors:"
   write(*,*) out_data
 
+  ! Summing the contents of a tensor
```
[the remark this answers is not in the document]
jatkinson1000: In retrospect this was probably a sum and was changed to mean at some point?

src/ftorch.fypp
```diff
@@ -909,6 +909,56 @@ contains
 
   #:endfor
 
+  ! ============================================================================
+  ! --- Other operators for computations involving tensors
+  ! ============================================================================
+
+  !> Overloads summation operator over the values in a tensor.
+  subroutine torch_tensor_sum(tensor, output)
```
jatkinson1000, line 917: As an observation the argument order is reversed between Fortran and C++
input, output vs. output, input

It would probably be good to make them consistent.

src/ftorch.fypp
```diff
@@ -973,7 +1023,7 @@ contains
 
   !> Retrieves the gradient with respect to a Torch Tensor.
   subroutine torch_tensor_get_gradient(tensor, gradient)
-    class(torch_tensor), intent(in) :: tensor      !! Tensor to compute the gradient with respect to
+    type(torch_tensor), intent(in) :: tensor       !! Tensor to compute the gradient with respect to
```
jatkinson1000, line 1026: Whoops!!

Though thanks I guess, as TIL a bunch more about Fortran polymorphism.

jatkinson1000 (commented):
Hi @jwallwork23 
This was stuck part-reviewed, apologies.
Have been through and finished off with a couple of comments for you.

examples/1_Tensor/tensor_manipulation.f90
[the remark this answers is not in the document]
jwallwork23: Yeah it was a copy-paste error, thanks.

src/ftorch.fypp
jwallwork23: Well this is the case across many of the functions/subroutines, though. I'll leave this change now for a future PR as it'd be a more widespread change.

jatkinson1000: Ah, that's a pain.
Opened https://github.com/Cambridge-ICCS/FTorch/issues/361 to reflect.

jwallwork23: Thanks

jatkinson1000: I went through the rest of `ftorch.F90` in #363 and don't think there are any other cases where argument order is inconsistent between the ftorch subroutines and their wrapped C/C++ functions.

Looking at the `=`, `*`, `/` overloads, `torch_tensor_from_array`, `torch_tensor_ones/zeros`, `torch_model_load` etc. the 'result' is always the first argument to the subroutine.
So to be consistent with that here I think it should be `subroutine torch_tensor_mean(output, tensor)` and `subroutine torch_tensor_sum(output, tensor)`.

However, `subroutine torch_tensor_get_gradient(tensor, gradient)` would also then want to become `subroutine torch_tensor_get_gradient(gradient, tensor)`.

jwallwork23: Addressed in be272e55324c30e3c7d69f32b3e8fc66126d6b95 and 1e81c8fd623d5f513d34f695523774a26366b293.

PR #347, opened 2025-03-31
Add error handling 

jatkinson1000:
This PR intends to improve the handling of errors in C++ to make things a little more instructive for users when things go wrong, rather than just having a bare stacktrace.

Currently there is some handling within FTorch to check tensors are created.
This adds handling on the C++ side to try and tell users why something has failed.

It is unclear how to test for this with pFUnit. At the moment I write some code, break it accordingly, and check I see the error expected...

Implemented:

- [x] generic error handling routines to be used throughout `ctorch.cpp`
    - These have not been made public or added to `ctorch.h` as designed to be used internally
- [x] generic routines to check tensors:
  - tensor is defined
  - tensor has requires_grad
  - gradient is defined for a tensor
- [x] usage of these throughout `ctorch.cpp`

Checking is not performed for getters (interrogation functions) or overloads (other than assign) as I felt probably unnecessary here.

This will introduce some overheads, but probably useful to have in for now as not expected to be significant and will help users in the development stage. Could consider a build with error handling off based on preprocessor flags in future if really desired.

### DIFF
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -406,9 +406,38 @@ void torch_tensor_backward(const torch_tensor_t tensor,
 }
 
 void torch_tensor_get_gradient(const torch_tensor_t tensor, torch_tensor_t gradient) {
-  auto t = reinterpret_cast<torch::Tensor *const>(tensor);
-  auto g = reinterpret_cast<torch::Tensor *>(gradient);
-  std::move(*g) = t->grad();
+  try {
+    // Cast the input pointers to torch::Tensor
+    auto t = reinterpret_cast<torch::Tensor *const>(tensor);
+    auto g = reinterpret_cast<torch::Tensor *>(gradient);
+
+    // Check if the tensor is valid and defined
+    if (!t || !t->defined()) {
+      throw std::invalid_argument("Input tensor is null or undefined.");
+    }
+
+    // Check if the output gradient pointer is valid
+    if (!g) {
+      throw std::invalid_argument("Output gradient pointer is null.");
+    }
+
+    // Check if the tensor requires gradients
+    if (!t->requires_grad()) {
+      throw std::runtime_error("Input tensor does not have requires_grad set.");
+    }
+
+    // Check if the gradient is defined
+    if (!t->grad().defined()) {
+      throw std::runtime_error("Gradient is undefined for the input tensor.");
+    }
+
+    // Assign the gradient to the output tensor
+    std::move(*g) = t->grad();
+  } catch (const std::exception &e) {
+    // Print the error message for debugging
+    std::cerr << "Error in torch_tensor_get_gradient: " << e.what() << std::endl;
+    std::terminate(); // Terminate here to avoid undefined behavior
+  }
 }
 
 // =============================================================================
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -179,7 +179,7 @@ torch_tensor_t torch_empty(int ndim, const int64_t *shape, torch_data_t dtype,
                            torch_device_t device_type, int device_index = -1,
                            const bool requires_grad = false) {
   torch::AutoGradMode enable_grad(requires_grad);
-  torch::Tensor *tensor = nullptr;
+  auto tensor = new torch::Tensor;
   try {
     // This doesn't throw if shape and dimensions are incompatible
     c10::IntArrayRef vshape(shape, ndim);
@@ -187,7 +187,6 @@ torch_tensor_t torch_empty(int ndim, const int64_t *shape, torch_data_t dtype,
                        .dtype(get_libtorch_dtype(dtype))
                        .device(get_libtorch_device(device_type, device_index))
                        .requires_grad(requires_grad);
-    tensor = new torch::Tensor;
     *tensor = torch::empty(vshape, options);
   } catch (const torch::Error &e) {
     std::cerr << "[ERROR]: " << e.msg() << std::endl;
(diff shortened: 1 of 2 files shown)


jatkinson1000:
In https://github.com/Cambridge-ICCS/FTorch/pull/347/commits/e9d7e5ea4259ddc7e6c46fa9a817cecc8f339f3a I abstract out the validation code into their own routines for re-use.

Since we do not expect to use them externally I have not added them to `ctorch.h`.

One thing that occurs to me is that when checking gradient is defined we effectively run the grad function in c++ twice (once to check, once to extract): https://github.com/Cambridge-ICCS/FTorch/blob/e9d7e5ea4259ddc7e6c46fa9a817cecc8f339f3a/src/ctorch.cpp#L160-L165

I wonder if we remove this check and replace it with a call to `﻿﻿validate_tensor` after we extract the gradient.
Doubt this is the serious bottleneck for now.
Any thoughts @jwallwork23

Side note, looks like pFUnit is having issues with CMake in GH actions 😢

jwallwork23:
> I wonder if we remove this check and replace it with a call to `﻿﻿validate_tensor` after we extract the gradient. Doubt this is the serious bottleneck for now. One option would be to make the routine return the gradient, though I am wary of doing this with the pointer abstractions we are using. Any thoughts @jwallwork23

Yeah it might be better to take that approach, although to be sure we'd have to check if the Torch `grad` method is just a getter or if it actually computes anything. If it's just a getter than it probably doesn't matter.

> Side note, looks like pFUnit is having issues with CMake in GH actions 😢

Alas!

jatkinson1000:
Rebased on main

jatkinson1000:
Decided not to add tensor validation to the getters as this felt like overkill, but could if desired.
Similarly have not added checks to any overloads other than assignment.

Think this is now ready for review.

src/ctorch.cpp
```diff
@@ -12,6 +12,25 @@
 #define GPU_DEVICE GPU_DEVICE_NONE
 #endif
 
+// =============================================================================
+// --- Functions to aid in consistent error handling
+// =============================================================================
+
+// Accept a string message and handle as error or warning if specified.
+// Can also accept a cleanup function if required.
```
jwallwork23, line 20: Worth noting that this is only for the error case.

src/ctorch.cpp
```diff
@@ -32,25 +50,23 @@ constexpr auto get_libtorch_dtype(torch_data_t dtype) {
...
+    ctorch_error("float16 not supported in Fortran");
     // See https://gcc.gnu.org/onlinedocs/gfortran/ISO_005fFORTRAN_005fENV.html
-    exit(EXIT_FAILURE);
   case torch_kFloat32:
     return torch::kFloat32;
   case torch_kFloat64:
     return torch::kFloat64;
   default:
-    std::cerr << "[WARNING]: unknown data type, setting to torch_kFloat32" << std::endl;
+    ctorch_error("unknown data type, setting to torch_kFloat32", true);
```
jwallwork23, line 60: I find calling `ctorch_error` to raise a warning a bit unclear. Given that the function has different implementations for the error and warning cases, would it not make more sense to just have separate `ctorch_error` and `ctorch_warning` functions? That'd also make it clearer because currently the `cleanup` function gets silently ignored if you are just raising a warning.

src/ctorch.cpp
```diff
@@ -147,24 +200,19 @@ torch_tensor_t torch_empty(int ndim, const int64_t *shape, torch_data_t dtype,
                            torch_device_t device_type, int device_index = -1,
                            const bool requires_grad = false) {
   torch::AutoGradMode enable_grad(requires_grad);
-  torch::Tensor *tensor = nullptr;
+  auto tensor = new torch::Tensor;
```
jwallwork23, line 203: Good catch with these.

src/ctorch.cpp
```diff
@@ -147,24 +200,19 @@ torch_tensor_t torch_empty(int ndim, const int64_t *shape, torch_data_t dtype,
...
                        .dtype(get_libtorch_dtype(dtype))
                        .device(get_libtorch_device(device_type, device_index))
                        .requires_grad(requires_grad);
-    tensor = new torch::Tensor;
     *tensor = torch::empty(vshape, options);
   } catch (const torch::Error &e) {
-    std::cerr << "[ERROR]: " << e.msg() << std::endl;
-    delete tensor;
-    exit(EXIT_FAILURE);
+    ctorch_error(e.msg(), false, [&]() { delete tensor; });
```
jwallwork23, line 213: Neat use of lambdas!

jwallwork23 (changes requested):
This is great, thanks @jatkinson1000. My only request is that you separate the warning functionality out of `ctorch_error` as a separate `ctorch_warning` function.

src/ctorch.cpp
jatkinson1000: Amended in https://github.com/Cambridge-ICCS/FTorch/pull/347/commits/479d39ec0e82b938b16770f2c0193e7deb9e4be8 addressing below.

src/ctorch.cpp
jatkinson1000: Yes, that's fair and I did wonder about doing this as the function developed.
Should be covered in https://github.com/Cambridge-ICCS/FTorch/pull/347/commits/479d39ec0e82b938b16770f2c0193e7deb9e4be8

src/ctorch.cpp
jatkinson1000: If I'm honest, this is not 100% my code ;)
Though I learnt some things, perhaps.

jatkinson1000:
Only final thing to note is that there are a couple of messages I couldn't handle with the new functions as they also output the `Torch::dtype` which I was not able to cast to string easily. For now I have left these as they were, though someone with better C++ may have a bright idea.

jwallwork23 (approved):
> Only final thing to note is that there are a couple of messages I couldn't handle with the new functions as they also output the `Torch::dtype` which I was not able to cast to string easily. For now I have left these as they were, though someone with better C++ may have a bright idea.

I think that's fine, we don't need to standardise everything - these changes already reduce multiplicity significantly.

niccolozanotti (approved):
Thanks @jatkinson1000 this is super useful.

PR #348, opened 2025-04-01
`torch_tensor_from_array` interface with default layout

### LINKED ISSUE #337: Default tensor layout
`torch_tensor_from_array`'s `tensor_layout` argument does have a use case, but in most cases it could just be defaulted to `[1, 2, ..., n]`, where `n` is the number of dimensions.

jwallwork23:
Closes #337.

It'd be convenient to default the tensor layout to `[1, 2, ..., n]`, as mentioned in #337. However, because this argument to `torch_tensor_from_array` comes before `device_type` (which is not optional), we can't simply make the `layout` argument optional because it would break the current API.

Thanks to the magic of `fypp`, this PR takes an alternative approach of providing more interfaces to `torch_tensor_from_array` such that it can be called without the `layout` argument, in which case it just gets the default value.

[As usual, the vast majority of lines added are due to fypp.]

### DIFF
--- a/pages/tensor.md
+++ b/pages/tensor.md
@@ -35,7 +35,7 @@ include:
   outputting data. `torch_tensor_from_array` may be called with or without the
   `layout` argument - an array which specifies the order in which indices should
   be looped over. The default `layout` is `[1,2,...,n]` implies that data will
-  be read into the correct indices by Torch. (See the
+  be read into the same indices by Torch. (See the
   [transposing user guide page](pages/transposing.html) for more details.
 
 It is *compulsory* to call one of the constructors before interacting with it in
--- a/pages/tensor.md
+++ b/pages/tensor.md
@@ -32,7 +32,10 @@ include:
   meaning the array must have been declared with the `target` property. The
   array will continue to be pointed to even when operations are applied to the
   tensor, so this subroutine can be used 'in advance' to set up an array for
-  outputting data.
+  outputting data. `torch_tensor_from_array` may be called with or without the
+  `layout` argument - an array which specifies the order in which indices should
+  be looped over. The default `layout` is `[1,2,...,n]`, i.e., the natural
+  ordering in Fortran.
 
 It is *compulsory* to call one of the constructors before interacting with it in
 any of the ways described in the following. Each of the constructors sets the
--- a/pages/tensor.md
+++ b/pages/tensor.md
@@ -34,8 +34,9 @@ include:
   tensor, so this subroutine can be used 'in advance' to set up an array for
   outputting data. `torch_tensor_from_array` may be called with or without the
   `layout` argument - an array which specifies the order in which indices should
-  be looped over. The default `layout` is `[1,2,...,n]`, i.e., the natural
-  ordering in Fortran.
+  be looped over. The default `layout` is `[1,2,...,n]` implies that data will
+  be read into the correct indices by Torch. (See the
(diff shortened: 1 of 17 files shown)


pages/tensor.md
```diff
@@ -32,7 +32,10 @@ include:
   meaning the array must have been declared with the `target` property. The
   array will continue to be pointed to even when operations are applied to the
   tensor, so this subroutine can be used 'in advance' to set up an array for
-  outputting data.
+  outputting data. `torch_tensor_from_array` may be called with or without the
+  `layout` argument - an array which specifies the order in which indices should
+  be looped over. The default `layout` is `[1,2,...,n]`, i.e., the natural
+  ordering in Fortran.
```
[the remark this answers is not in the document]
jwallwork23: How about what I put in 15524d69fff8dbad448423db8054579314ea48ff?

jwallwork23:
Thanks very much for the review!

>     * This, or something very close, warrants a minor release.

Agreed. Would that be v1.1?
 
>     * The growth of tensor creation routines is another nudge that we should look at breaking apart into smaller modules, as in [Add optimisers #320](https://github.com/Cambridge-ICCS/FTorch/pull/320) for optimizers, but this is for a different PR.

Yeah, I agree with this, too.

jwallwork23:
Hm, it's odd that this pFUnit error sometimes happens and sometimes doesn't. Given that it passed before the last commit, which was docs-only, this PR does pass the test suite.

pages/tensor.md
```diff
@@ -32,7 +32,11 @@ include:
   meaning the array must have been declared with the `target` property. The
   array will continue to be pointed to even when operations are applied to the
   tensor, so this subroutine can be used 'in advance' to set up an array for
-  outputting data.
+  outputting data. `torch_tensor_from_array` may be called with or without the
+  `layout` argument - an array which specifies the order in which indices should
+  be looped over. The default `layout` is `[1,2,...,n]` implies that data will
+  be read into the correct indices by Torch. (See the
```
jatkinson1000, line 38: Similarly, I feel "correct" could be ambiguous depending on how the reader feel's this should be implemented, perhaps "same"?
```suggestion
  be read into the same indices by Torch. (See the
```

pages/transposing.md
```diff
@@ -132,20 +132,23 @@ Not doing so could leave us open to introducing bugs.
...
 It takes the form of an array specifying which order to read the indices in.
 i.e. `[1, 2]` will read `i` then `j`.
 By passing `layout = [1, 2]` the data will be read into the correct indices by
-Torch.
-
-This is achieved by wrapping the `torch_tensor_from_blob` function to automatically
-generate strides assuming that a straightforward conversion between
-row- and column-major is what should happen.
+Torch. The natural ordering `[1, 2, ..., n]` (where `n` is the dimension of the
+array) is the default used by `torch_tensor_from_array`. In most cases, it
```
jatkinson1000, line 145:
```suggestion
array) is the default used by `torch_tensor_from_array`.
In cases where your tensors are indexed the same way in both Fortran and Torch, it
```

jatkinson1000 (approved):
Thanks @jwallwork23 this is great.

2 Small phrasing suggestions up to you.
And a rebase should resolve the workflow problems.

jwallwork23:
> 2 Small phrasing suggestions up to you.

Thanks, both good suggestions.

PR #353, opened 2025-04-02
Variable definitions in docstrings for tensor API

jwallwork23:
Apologies for not adding these when I added these procedures. I also added a few that were missing from other procedures.


jatkinson1000 (approved):
My bad too for missing these in the review 😅 
Go for it @jwallwork23!

PR #356, opened 2025-04-08
Add workflow to catch broken URLs in the docs

### LINKED ISSUE #323: Add CI task to check for broken links in the docs
As per @jwallwork23 suggestion, it could be useful to think of CI task to prevent broken links (especially Markdown internal links) to end up in the docs.  See e.g. commit [e97f424](https://github.com/Cambridge-ICCS/FTorch/commit/e97f424dc4ca3b26a019ec693e6c4793987e9e64).

jatkinson1000:
Spotted by @niccolozanotti and patched here.

I also did a quick scan for anything else that may be outdated.

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -71,7 +71,7 @@ jobs:
           # The URLs regex to check
           # Key is the final negative lookbehind `(?<![).,-])` which describes what should be dropped from the end of a url.
           # Not infallibe, but should catch most expected cases.
-          url-regex: 'http[s]?:\/\/(?:[a-zA-Z0-9$-_@.&+!*(),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+(?<![).,])'
+          url-regex: 'https?:\/\/(?:[^\s()]+|\([^\s()]*\))+'
 
       - name: Build docs with FORD
         run: ford FTorch.md
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -63,15 +63,10 @@ jobs:
           pip install --group docs
 
       - name: Check broken URLs
-        uses: RazcoDev/broken-urls-in-code@v1.0.1
-        with:
-          directory: 'pages/'
-          # The files glob to check the URLs inside
-          files-glob: '**/*.md'
-          # The URLs regex to check
-          # Key is the final negative lookbehind `(?<![).,-])` which describes what should be dropped from the end of a url.
-          # Not infallibe, but should catch most expected cases.
-          url-regex: 'https?:\/\/(?:[^\s()]+|\([^\s()]*\))+'
+        run: |
+          cd pages/
+          chmod +x checklinks.sh
+          ./checklinks.sh "*.md"
 
       - name: Build docs with FORD
         run: ford FTorch.md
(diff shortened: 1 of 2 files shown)


joewallwork (approved):

jatkinson1000:
Tried:

- https://github.com/gaurav-nelson/github-action-markdown-link-check
  Struggles to parse internal links, and reports 403 errors for a number of intel webpages see https://github.com/Cambridge-ICCS/FTorch/pull/356/commits/8a67b930ce93c75d44bf966627506df55118531c
- https://github.com/RazcoDev/broken-urls-in-code
  Struggles with extracting links properly due to markdown syntax (trailing `)` etc.) That was not easily fixed for all edge cases with modification to the regex - see https://github.com/Cambridge-ICCS/FTorch/pull/356/commits/1760a897d5e94e1a55667498d18ff61965adeeec

Will leave for now as non-urgent and taken a fair amount of time today.

jatkinson1000:
Made my own 🙄 

Works fine on my machine, but fails on GitHub CI for the gcc links.
Returns 000 exit code for curl.

Is Gnu blocking requests from GitHub servers?? 🤣 

This was supposed to be simple 😭

jatkinson1000:
Huh, updated but passed fine that time 🤷

jatkinson1000:
TODO:

- [ ] Add some docs
- [ ] Rename to `checkmdlinks.sh`

Mikolaj-A-Kowalski:
I would not be inclined to scrap the automated check completely (as in do we *really* trust ourselves to run the script as often as we should ;-)  ), but I completely agree that it is far to brittle to be blocking for the CI. 

What I am thinking is we could add [`continiue-on-error`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax?versionId=free-pro-team%40latest&productId=actions#jobsjob_idstepscontinue-on-error) to the link check step so it doesn't kill the job. 

Also to make the results more visible we can raise a 'notice' using the appropriate [workflow command](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands?versionId=free-pro-team%40latest&productId=actions&restPage=reference%2Cworkflows-and-actions%2Cworkflow-syntax#setting-a-notice-message), i.e. for the failure echo something like that:     
```
echo "::notice title=Maybe broken link:: '$url' may not be reachable got  (HTTP $status_code). Note this check is brittle. Please check manually and ignore this message if the link seems OK."
```  
This should create a helpful (?) banner. Not entirely sure where though if the file and line is not explicitly specified...    

May also be a better idea to collect all failing links into a single message to avoid getting multiple banners.

My thinking is that we can keep the check running for some time and we will get a sense of how annoying vs useful it will be. If it creates too much noise we can always scrap it.

.github/workflows/build_docs.yml
```diff
@@ -62,6 +62,12 @@ jobs:
           python -m pip install --upgrade pip
           pip install --group docs
 
+      - name: Check broken URLs
+        run: |
+          cd pages/
+          chmod +x checklinks.sh
+          ./checklinks.sh
```
joewallwork, line 69: Is it bad practice to ship scripts that are already executable?

joewallwork (changes requested):
This version of the script mostly works well for me, although I get the rate limit errors mentioned.

> I'm wondering if we just add it as a script and rely on developers to run instead as a useful utility

How about we add it as a CI job whose only trigger is calendar-based? e.g., runs early AM on the first day of the month?

PR #357, opened 2025-04-11
Run test suite with `ctest`

### LINKED ISSUE #249: Pass arguments to `ctest`?
Whilst working on #239, I figured out how to pass unused arguments to `ctest`. This could be useful for, e.g., running specific tests from the `run_test_suite` script.

However, my version of `ctest` (from CMake 3.22.1) silently drops invalid arguments, which I don't think we want (see ). This was fixed for a later version of CMake (see ), though.

jwallwork23:
Closes #249.

Tidies up the test suite to just run with `ctest` directly, rather than navigating into subdirectories.

A big advantage of this approach is that we can use the various useful arguments for `ctest`, such as `ctest -R unittest_tensor_autograd` to run that specific test.

### DIFF
--- a/run_test_suite.sh
+++ b/run_test_suite.sh
@@ -82,16 +82,5 @@ fi
 
 # Run integration tests
 if [ "${RUN_INTEGRATION}" = true ]; then
-  if [ -e "${BUILD_DIR}/examples/6_MultiGPU" ]; then
-    EXAMPLES="1_Tensor 2_SimpleNet 3_ResNet 4_MultiIO 6_MultiGPU 7_MPI 8_Autograd"
-  else
-    EXAMPLES="1_Tensor 2_SimpleNet 3_ResNet 4_MultiIO 7_MPI 8_Autograd"
-  fi
-  export PIP_REQUIRE_VIRTUALENV=true
-  for EXAMPLE in ${EXAMPLES}; do
-    python -m pip -q install -r examples/"${EXAMPLE}"/requirements.txt
-    cd "${BUILD_DIR}"/examples/"${EXAMPLE}"
-    ctest "${CTEST_ARGS}"
-    cd -
-  done
+  ctest -R example ${CTEST_ARGS}
 fi
--- a/run_test_suite.sh
+++ b/run_test_suite.sh
@@ -73,11 +73,11 @@ else
   CTEST_ARGS=""
 fi
 
+cd "${BUILD_DIR}"
+
 # Run unit tests
 if [ "${RUN_UNIT}" = true ]; then
-  cd "${BUILD_DIR}/test/unit"
-  ctest "${CTEST_ARGS}"
-  cd -
+  ctest -R unittest ${CTEST_ARGS}
 fi
 
 # Run integration tests
--- a/examples/1_Tensor/CMakeLists.txt
+++ b/examples/1_Tensor/CMakeLists.txt
@@ -30,6 +30,7 @@ if(CMAKE_BUILD_TESTS)
     NAME example1_tensor_manipulation
     COMMAND tensor_manipulation
     WORKING_DIRECTORY ${PROJECT_BINARY_DIR})
-  set_tests_properties(example1_tensor_manipulation PROPERTIES PASS_REGULAR_EXPRESSION
+  set_tests_properties(example1_tensor_manipulation
+                       PROPERTIES PASS_REGULAR_EXPRESSION
                        "Tensor manipulation example ran successfully")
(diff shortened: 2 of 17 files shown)


run_test_suite.sh
```diff
@@ -73,25 +73,14 @@ else
...
 fi
 
 # Run integration tests
 if [ "${RUN_INTEGRATION}" = true ]; then
-  if [ -e "${BUILD_DIR}/examples/6_MultiGPU" ]; then
-    EXAMPLES="1_Tensor 2_SimpleNet 3_ResNet 4_MultiIO 6_MultiGPU 7_MPI 8_Autograd"
-  else
-    EXAMPLES="1_Tensor 2_SimpleNet 3_ResNet 4_MultiIO 7_MPI 8_Autograd"
-  fi
-  export PIP_REQUIRE_VIRTUALENV=true
```
jatkinson1000, line 90: Think this was supposed to stay in :)

jatkinson1000 (commented):
From Discussion, suggest removing the run_tests script (at least for bash) and telling users how to run tests through `ctest` as this covers the functionalities we built into the script.

Also check to see if this carries across to Windows.

run_test_suite.sh
jwallwork23: Put back in https://github.com/Cambridge-ICCS/FTorch/pull/357/commits/2126cf4b43b42031acffdec89540be991581899f, albeit in a different form.

jatkinson1000: Have not tested, but will trust you that this works.
Good to cover conda as well as venv, and to enforce where requirements are installed.

However, does this now require the `requirements.txt` to be removed from all of the examples and the `README.md`s updating?
Even if installation of dependencies is now done through CMake I think it would still be good to list for the users what is needed for a particular exercise in case they want to replicate it in their own code..

jwallwork23:
> Also check to see if this carries across to Windows.

I made a start on this last week but it became a bit of a headache. Got most of the way to setting things up, though.

jwallwork23:
[Rebased on top of `main`]

jwallwork23:
Perhaps we just merge this, given that the Windows CI is deactivated anyway?

.github/workflows/test_suite_windows.yml
```diff
@@ -92,4 +92,9 @@ jobs:
           for /f "tokens=2*" %%i in ('pip show torch ^| findstr /R "^Location"') do set torch_path=%%i
           set PATH=C:\Program Files (x86)\FTorch\bin;%PATH%
           set PATH=%torch_path%\torch\lib;%PATH%
-          run_test_suite.bat
+          cd build
+          ctest -V -R example1
+          ctest -V -R example2
+          ctest -V -R example3
+          ctest -V -R example4
+          ctest -V -R example8
```
jatkinson1000: Suggest explicit options in static script for clarity:

```suggestion
          ctest --verbose --tests-regex example1
          ctest --verbose --tests-regex example2
          ctest --verbose --tests-regex example3
          ctest --verbose --tests-regex example4
          ctest --verbose --tests-regex example8
```

.github/workflows/test_suite_ubuntu.yml
```diff
@@ -102,9 +102,11 @@ jobs:
       - name: Run unit tests
         run: |
           . ftorch/bin/activate
-          ./run_test_suite.sh --verbose --unit-only
+          cd build
+          ctest -V -R unit
```
jatkinson1000:
```suggestion
          ctest --verbose --tests-regex unit
```

.github/workflows/test_suite_ubuntu.yml
```diff
@@ -102,9 +102,11 @@ jobs:
...
 
       - name: Run integration tests
         run: |
           . ftorch/bin/activate
-          ./run_test_suite.sh --verbose --integration-only
+          cd build
+          ctest -V -R example
```
jatkinson1000:
```suggestion
          ctest --verbose --tests-regex example
```

conda/README.md
```diff
@@ -145,13 +145,15 @@ make tests
 make install
 ```
 
-The unit tests only can be run using the provided script:
+The unit tests can be run by navigating to the FTorch build directory and
+calling `ctest`:
 ```sh
-./run_test_suite.sh --unit-only
+cd build
+ctest -V
```
jatkinson1000: I presume this also needs the `-R` flag?

```suggestion
ctest --verbose --tests-regex
```

PR #359, opened 2025-04-11
Raise an error if user attempts to assign to tensor with `requires_grad=.true.`

### LINKED ISSUE #327: Issue with copying tensor with `requires_grad` set to true
I had some unexpected behaviour using the overloaded assignment (`=`) operator to set a tensor equal to another that had `requires_grad=.true.`.

It resulted in the following stacktrace (reduced output):
```
libc++abi: terminating due to uncaught exception of type c10::Error: a leaf Variable that requires grad is being used in an in-place operation.

...

frame #4: at::_ops::copy_::call(at::Tensor&, at::Tensor const&, bool) + 304 (0x11124c9dc in libtorch_cpu.dylib)
frame #5: torch_tensor_assign + 92 (0x104756100 in libftorch.dylib)
```

It seems perhaps we cannot copy a tensor that has `requires_grad` set?
Happy to try and rewrite a MWE if needed.

Perhaps there are different versions of "copying" as we sometimes see in python?

From what I read leaf tensors cannot be modified as they break the backprop cycle.
It seems there is a need to use `tensor.clone()` rather than `tensor.copy()`.
Though there are some other hacky suggestions out there like creating a new tensor through an operation:
```
tensor = tensor + 0;
tensor.copy_(source_tensor);
```

jwallwork23:
Closes #327.

As discussed in-person, it shouldn't be allowed to modify a tensor that has `requires_grad=true` because this is a tensor that's marked as something we differentiate with respect to. This PR raises an error if someone attempts to do so.

~~Unfortunately, I had to raise the error on the C++ side because it didn't work on the Fortran side in `torch_tensor_assign`. The reason is that the `output` tensor is actually the return value of a function, so we can't query its attributes. It seems to always be set to `.false.`. However, it is possible to do this when its pointer gets passed to C++.~~

Edit: Ignore me! Having set things back up on my fresh OS, I'd set up my editor autocmds incorrectly and was auto-generating `ftorch.f90` rather than `ftorch.F90`. The error is now raised on the Fortran side.

### DIFF
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -325,6 +325,11 @@ void torch_tensor_assign(torch_tensor_t output, const torch_tensor_t input) {
   auto out = reinterpret_cast<torch::Tensor *>(output);
   auto in = reinterpret_cast<torch::Tensor *const>(input);
   torch::AutoGradMode enable_grad(in->requires_grad());
+  if (out->requires_grad()) {
+    std::cerr << "[ERROR]: Cannot modify the values of a tensor with requires_grad=true"
+              << std::endl;
+    exit(EXIT_FAILURE);
+  }
   // NOTE: The following line ensures that the output tensor continues to point to a
   //       Fortran array if it was set up to do so using torch_tensor_from_array. If
   //       it's removed then the Fortran array keeps its original value and is no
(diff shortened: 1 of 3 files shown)


jatkinson1000 (approved):
LGTM, thanks for investigating @jwallwork23.

One point to consider - on the C++ side I have been trying to include the offending function/subroutine name in the error report to provide people with an pointer to where its going wrong. I can't remember offhand how much information the FTorch stacktrace gives, but you may want to do the same here - the current message doesn't say _how_ we are trying to modify the tensor.

jwallwork23:
> LGTM, thanks for investigating @jwallwork23.
> 
> One point to consider - on the C++ side I have been trying to include the offending function/subroutine name in the error report to provide people with an pointer to where its going wrong. I can't remember offhand how much information the FTorch stacktrace gives, but you may want to do the same here - the current message doesn't say _how_ we are trying to modify the tensor.

Clarified that we're working with an assignment in db63bf6.

PR #360, opened 2025-04-15
Adding ability to move `torch_tensor`s between `device`s and `dtypes`

### LINKED ISSUE #299: Add ability to move tensors between devices in Fortran
In reviewing #272 it would be useful to be able to move an existing tensor from one device to another (`to()` method in PyTorch).

This would probably be a method of the `torch_tensor` class wrapping the LibTorch functionality that performs this.

niccolozanotti:
Opening this draft PR as a follow-up to our discussion @jatkinson1000 @jwallwork23. This would close #299.

a2ba778 contains a proposed implementation of the subroutine `torch_tensor_to` which effectively moves a source tensor to a target tensor with a taget `device` and `dtype`.

---
I am actively working on the testing. Right now `torch_kCPU` --> `torch_kCUDA` and `torch_kCPU` --> `torch_kMPS` with dtype `Float64` -->`Float32` seemingly working fine.

Any suggestion on the most appropriate [unit test](https://github.com/Cambridge-ICCS/FTorch/tree/main/test/unit) to add this to?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -197,3 +197,6 @@ cython_debug/
 
 # Apple Desktop Services Store
 **.DS_Store
+
+# fortls LSP config file
+.fortls
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -317,6 +317,17 @@ void torch_tensor_zero(torch_tensor_t tensor) {
   t->zero_();
 }
 
+void torch_tensor_to(const torch_tensor_t source_tensor, torch_tensor_t dest_tensor,
+                     bool non_blocking, bool copy) {
+  auto source_t = reinterpret_cast<torch::Tensor *>(source_tensor);
+  auto dest_t = reinterpret_cast<torch::Tensor *>(dest_tensor);
+
+  torch::Device device_type = dest_t->device();
+  at::ScalarType dtype = dest_t->scalar_type();
+
+  std::move(*dest_t) = source_t->to(device_type, dtype, non_blocking, copy);
+}
+
 // =====================================================================================
 // --- Operator overloads acting on tensors
 // =====================================================================================
--- a/src/ftorch.F90
+++ b/src/ftorch.F90
@@ -163,7 +163,7 @@ module ftorch
     end function torch_from_blob_c
   end interface
 
-  ! ============================================================================
+    ! ============================================================================
   ! --- Interfaces for overloaded operators acting on tensors
   ! ============================================================================
 
@@ -2456,6 +2456,70 @@ contains
     end if
     call torch_tensor_zero_c(tensor%p)
   end subroutine torch_tensor_zero
+
+  !> Moves a source_tensor tensor to a target tensor's device and dtype
+  subroutine torch_tensor_to(source_tensor, target_tensor, non_blocking, copy)
+    use, intrinsic :: iso_c_binding, only : c_bool, c_int, c_int64_t
+    type(torch_tensor), intent(in) :: source_tensor      !! Source tensor to be moved
+    type(torch_tensor), intent(inout) :: target_tensor   !! Target tensor with the desired device and dtype
+    logical, optional, intent(in) :: non_blocking        !! Whether to perform asynchronous copy
+    logical, optional, intent(in) :: copy                !! Whether to force a copy
+  
+    logical(c_bool) :: non_blocking_value, copy_value
+    integer(c_int) :: source_rank, target_rank, i
+    integer(c_int64_t), pointer :: source_shape(:), target_shape(:)
+
+    interface
+      subroutine torch_tensor_to_c(src_tensor, targ_tensor, non_blocking, copy) &
+          bind(c, name = 'torch_tensor_to')
+        use, intrinsic :: iso_c_binding, only : c_bool, c_ptr
+        implicit none
+        type(c_ptr), value, intent(in) :: src_tensor
+        type(c_ptr), value, intent(in) :: targ_tensor
+        logical(c_bool), value, intent(in) :: non_blocking
+        logical(c_bool), value, intent(in) :: copy
+      end subroutine torch_tensor_to_c
+    end interface
+
+    ! Check for rank and shape consistency between the source and target tensors
+    source_rank = source_tensor%get_rank()
+    target_rank = target_tensor%get_rank()
+    
+    if (source_rank /= target_rank) then
+      write(*,*) "Error :: Cannot move source_tensor to target_tensor because the ranks do not match."
+      write(*,*) "Source tensor rank:", source_rank, "Target tensor rank:", target_rank
+      stop 1
+    end if
+    
+    source_shape => source_tensor%get_shape()
+    target_shape => target_tensor%get_shape()
+    
+    do i = 1, source_rank
+      if (source_shape(i) /= target_shape(i)) then
+        write(*,*) "Error :: Cannot move source_tensor to target_tensor because the shapes do not match."
+        write(*,*) "Dimension", i, "mismatch: source_tensor =", source_shape(i), &
+            "Target =", target_shape(i)
+        stop 1
+      end if
+    end do
+  
+    ! Process optional arguments
+    if (present(non_blocking)) then
+      non_blocking_value = non_blocking
+    else
+      non_blocking_value = .false.
+    end if
+  
+    if (present(copy)) then
+      copy_value = copy
+    else
+      copy_value = .false.
+    end if
+
+    
+    call torch_tensor_to_c(source_tensor%p, target_tensor%p, non_blocking_value, copy_value)
+
+  end subroutine torch_tensor_to
 
   ! ============================================================================
   ! --- Overloaded operators acting on tensors
--- a/src/ftorch.fypp
+++ b/src/ftorch.fypp
@@ -132,7 +132,7 @@ module ftorch
     end function torch_from_blob_c
   end interface
 
-  ! ============================================================================
+    ! ============================================================================
   ! --- Interfaces for overloaded operators acting on tensors
   ! ============================================================================
 
@@ -690,6 +690,70 @@ contains
     end if
     call torch_tensor_zero_c(tensor%p)
   end subroutine torch_tensor_zero
+
+  !> Moves a source_tensor tensor to a target tensor's device and dtype
+  subroutine torch_tensor_to(source_tensor, target_tensor, non_blocking, copy)
+    use, intrinsic :: iso_c_binding, only : c_bool, c_int, c_int64_t
+    type(torch_tensor), intent(in) :: source_tensor      !! Source tensor to be moved
+    type(torch_tensor), intent(inout) :: target_tensor   !! Target tensor with the desired device and dtype
+    logical, optional, intent(in) :: non_blocking        !! Whether to perform asynchronous copy
+    logical, optional, intent(in) :: copy                !! Whether to force a copy
+  
+    logical(c_bool) :: non_blocking_value, copy_value
+    integer(c_int) :: source_rank, target_rank, i
+    integer(c_int64_t), pointer :: source_shape(:), target_shape(:)
+
+    interface
+      subroutine torch_tensor_to_c(src_tensor, targ_tensor, non_blocking, copy) &
+          bind(c, name = 'torch_tensor_to')
+        use, intrinsic :: iso_c_binding, only : c_bool, c_ptr
+        implicit none
+        type(c_ptr), value, intent(in) :: src_tensor
+        type(c_ptr), value, intent(in) :: targ_tensor
+        logical(c_bool), value, intent(in) :: non_blocking
+        logical(c_bool), value, intent(in) :: copy
+      end subroutine torch_tensor_to_c
+    end interface
+
+    ! Check for rank and shape consistency between the source and target tensors
+    source_rank = source_tensor%get_rank()
+    target_rank = target_tensor%get_rank()
+    
+    if (source_rank /= target_rank) then
+      write(*,*) "Error :: Cannot move source_tensor to target_tensor because the ranks do not match."
+      write(*,*) "Source tensor rank:", source_rank, "Target tensor rank:", target_rank
+      stop 1
+    end if
+    
+    source_shape => source_tensor%get_shape()
+    target_shape => target_tensor%get_shape()
+    
+    do i = 1, source_rank
+      if (source_shape(i) /= target_shape(i)) then
+        write(*,*) "Error :: Cannot move source_tensor to target_tensor because the shapes do not match."
+        write(*,*) "Dimension", i, "mismatch: source_tensor =", source_shape(i), &
+            "Target =", target_shape(i)
+        stop 1
+      end if
+    end do
+  
+    ! Process optional arguments
+    if (present(non_blocking)) then
+      non_blocking_value = non_blocking
+    else
+      non_blocking_value = .false.
+    end if
+  
+    if (present(copy)) then
+      copy_value = copy
+    else
+      copy_value = .false.
+    end if
+
+    
+    call torch_tensor_to_c(source_tensor%p, target_tensor%p, non_blocking_value, copy_value)
+
+  end subroutine torch_tensor_to
 
   ! ============================================================================
   ! --- Overloaded operators acting on tensors
--- a/src/ctorch.h
+++ b/src/ctorch.h
@@ -183,6 +183,18 @@ EXPORT_C void torch_tensor_delete(torch_tensor_t tensor);
  */
 EXPORT_C void torch_tensor_zero(torch_tensor_t tensor);
 
+/**
+ * Function to move a tensor to a target tensor's device and dtype
+ * @param Tensor to be moved
+ * @param Tensor with the target device and dtype
+ * @param if True and this copy is happening between CPU and GPU, the copy may occur
+ * asynchronously
+ * @param if True, a new Tensor is created even when the Tensor already matches the
+ * desired conversion
+ */
+EXPORT_C void torch_tensor_to(const torch_tensor_t source_tensor,
+                              torch_tensor_t dest_tensor, bool non_blocking, bool copy);
+
 // =============================================================================
 // --- Operator overloads acting on tensors
 // =============================================================================


src/ftorch.fypp
```diff
@@ -132,7 +132,7 @@ module ftorch
     end function torch_from_blob_c
   end interface
 
-  ! ============================================================================
+    ! ============================================================================
```
jatkinson1000, line 135:
```suggestion
  ! ============================================================================
```

src/ftorch.fypp
```diff
@@ -691,6 +691,70 @@ contains
     call torch_tensor_zero_c(tensor%p)
   end subroutine torch_tensor_zero
 
+  !> Moves a source_tensor tensor to a target tensor's device and dtype
+  subroutine torch_tensor_to(source_tensor, target_tensor, non_blocking, copy)
+    use, intrinsic :: iso_c_binding, only : c_bool, c_int, c_int64_t
+    type(torch_tensor), intent(in) :: source_tensor      !! Source tensor to be moved
+    type(torch_tensor), intent(inout) :: target_tensor   !! Target tensor with the desired device and dtype
+    logical, optional, intent(in) :: non_blocking        !! Whether to perform asynchronous copy
+    logical, optional, intent(in) :: copy                !! Whether to force a copy
```
jatkinson1000, line 700: What does it mean to "force a copy"?

src/ftorch.fypp
```diff
@@ -691,6 +691,70 @@ contains
...
+    target_shape => target_tensor%get_shape()
+    
+    do i = 1, source_rank
+      if (source_shape(i) /= target_shape(i)) then
+        write(*,*) "Error :: Cannot move source_tensor to target_tensor because the shapes do not match."
+        write(*,*) "Dimension", i, "mismatch: source_tensor =", source_shape(i), &
+            "Target =", target_shape(i)
+        stop 1
+      end if
+    end do
```
jatkinson1000, line 738: This guarding is really good, thanks for including!

One thing that might improve is to note the subroutine in which the error occurs, e.g. '`Error in torch_tensor_to: ...`'

jatkinson1000 (commented):
This is looking really good @niccolozanotti 

Left a couple of small comments.
Plus some whitespace to clear up in the Fortran.

As for unit tests, I think I would suggest "tensor manipulation".

src/ftorch.fypp
niccolozanotti: From the [python docs](https://pytorch.org/docs/stable/generated/torch.Tensor.to.html) of `torch.Tensor.to`: 
> When copy is set, a new Tensor is created even when the Tensor already matches the desired conversion

jatkinson1000: I was wondering if this was perhaps a redundant argument for us, as we always create a second tensor to which we will copy to with this routine.

However, after some thought I actually think it is valid, as that second tensor is always temporary with its address being reassigned to the item generated by the C++ implementation of `to`.
If I understand correctly.

niccolozanotti: Yes, exactly.

src/ftorch.fypp
niccolozanotti: Addressed in c2951cb.

src/ftorch.fypp
niccolozanotti: Addressed in c2951cb.

niccolozanotti:
Rebased on top of main.

niccolozanotti:
I added the test we've been talking about @jatkinson1000 @jwallwork23 and I think this is ready for review.

src/ftorch.F90
```diff
@@ -2457,6 +2457,69 @@ end subroutine torch_tensor_zero_c
...
   end subroutine torch_tensor_zero
 
+  !> Moves a source_tensor tensor to a target tensor's device and dtype
+  subroutine torch_tensor_to(source_tensor, target_tensor, non_blocking, copy)
+    use, intrinsic :: iso_c_binding, only : c_bool, c_int, c_int64_t
+    type(torch_tensor), intent(in) :: source_tensor      !! Source tensor to be moved
+    type(torch_tensor), intent(inout) :: target_tensor   !! Target tensor with the desired device and dtype
+    logical, optional, intent(in) :: non_blocking        !! Whether to perform asynchronous copy
+    logical, optional, intent(in) :: copy                !! Whether to force a copy if source_tensor already
+                                                         !! matches the desired conversion
```
jwallwork23, line 2467: Sorry, what does this mean?
> if source_tensor already matches the desired conversion

src/ftorch.F90
```diff
@@ -2457,6 +2457,69 @@ end subroutine torch_tensor_zero_c
...
+      subroutine torch_tensor_to_c(src_tensor, targ_tensor, non_blocking, copy) &
+          bind(c, name = 'torch_tensor_to')
+        use, intrinsic :: iso_c_binding, only : c_bool, c_ptr
+        implicit none
+        type(c_ptr), value, intent(in) :: src_tensor
+        type(c_ptr), value, intent(in) :: targ_tensor
+        logical(c_bool), value, intent(in) :: non_blocking
+        logical(c_bool), value, intent(in) :: copy
+      end subroutine torch_tensor_to_c
+    end interface
```
jwallwork23, line 2482: In #363, we are moving to make the variable names used in C interfaces consistent with those in the surrounding Fortran subroutine/function, albeit appended with `_c`. Please could you do the same here?

test/unit/test_tensor_manipulation_cuda.pf
```diff
@@ -0,0 +1,88 @@
...
+
+    do k = 1, 4
+       do j = 1, 3
+          do i = 1, 2
+             dp_data(i,j,k) = real(100*i + 10*j + k, dp)
+          end do
+       end do
+    end do
+
+    sp_data = real(dp_data, sp) ! source data cast to single-precision real for testing
```
jwallwork23, line 44: Why not just define the array in single precision to begin with?

jwallwork23: Also, why did you want/need to convert precisions anyway?

test/unit/test_tensor_manipulation_cuda.pf
```diff
@@ -0,0 +1,88 @@
...
+    test_array = 0.0_sp         ! Fortran array to extract target tensor data from
+    test_pass = .false.
```
jwallwork23, line 46: You don't need to initialise these.

test/unit/test_tensor_manipulation_cuda.pf
```diff
@@ -0,0 +1,88 @@
...
+contains
+
+    ! Unit test for the torch_tensor_to subroutine moving a tensor from the CPU (Float64) to a CUDA device (FLoat32)
+  @test
+  subroutine test_torch_tensor_to()
+    use, intrinsic :: iso_fortran_env, only: sp => real32, dp => real64
+
+    type(torch_tensor) :: cpu_tensor, gpu_tensor, test_tensor
+    real(dp), dimension(2,3,4) :: dp_data
+    real(sp), dimension(2,3,4) :: sp_data, test_array
```
jwallwork23, line 28: I wonder if `temp_array` might be a better name than `test_array`?

test/unit/test_tensor_manipulation_cuda.pf
```diff
@@ -0,0 +1,88 @@
...
+    end do
+
+    sp_data = real(dp_data, sp) ! source data cast to single-precision real for testing
+    test_array = 0.0_sp         ! Fortran array to extract target tensor data from
+    test_pass = .false.
+
+    call torch_tensor_from_array(cpu_tensor, dp_data, device_type_src)
+
+    call torch_tensor_empty(gpu_tensor,cpu_tensor%get_rank(),cpu_tensor%get_shape(), &
+                           dtype_tar,device_type_tar)
```
jwallwork23, line 51: Preferred formatting
```suggestion
    call torch_tensor_empty(gpu_tensor, cpu_tensor%get_rank(), cpu_tensor%get_shape(), &
                           dtype_tar, device_type_tar)
```

test/unit/test_tensor_manipulation_cuda.pf
```diff
@@ -0,0 +1,88 @@
...
+
+    ! Move to target device and dtype
+    call torch_tensor_to(cpu_tensor, gpu_tensor)
+
+    @assertTrue(gpu_tensor%get_device_type() == device_type_tar)
+    @assertTrue(gpu_tensor%get_shape() == cpu_tensor%get_shape())
+    @assertTrue(gpu_tensor%get_rank() == cpu_tensor%get_rank())
+
+    ! Temporary torch_tensor for testing
+    call torch_tensor_from_array(test_tensor, test_array,device_type_src)
```
jwallwork23, line 61: Preferred formatting
```suggestion
    call torch_tensor_from_array(test_tensor, test_array, device_type_src)
```

test/unit/test_tensor_manipulation_cuda.pf
```diff
@@ -0,0 +1,88 @@
...
+      call torch_tensor_delete(cpu_tensor)
+      call torch_tensor_delete(gpu_tensor)
+      call torch_tensor_delete(test_tensor)
+
+      stop 999
+    end if
+
+    call torch_tensor_delete(cpu_tensor)
+    call torch_tensor_delete(gpu_tensor)
+    call torch_tensor_delete(test_tensor)
```
jwallwork23, line 84: I think we can drop these calls? The finalizer should call `torch_tensor_delete` automatically.

jwallwork23 (changes requested):
Thanks for this @niccolozanotti - great work!

I have a few suggested edits.

src/ftorch.F90
niccolozanotti: We addressed that in one resolved comment above. From the [docs](https://pytorch.org/docs/stable/generated/torch.Tensor.to.html) I understand this means that if 
- `copy=.false.` If the tensor is already on the specified device and has the specified data type, no copy is made and the original tensor is returned. If a conversion is needed (different device or data type), a new tensor will be created with the converted values
- `copy=.true.`: A new tensor is always created and returned, even if no conversion is needed

test/unit/test_tensor_manipulation_cuda.pf
jatkinson1000: @niccolozanotti worth noting that if you build with tests you can run Fortitude on the generated Fortran to check formatting.
It requires a little sleuthing to get back to the correct pf lines, like with fypp, but I have used this when writing tests.

.gitignore
```diff
@@ -197,3 +197,6 @@ cython_debug/
 
 # Apple Desktop Services Store
 **.DS_Store
+
+# fortls LSP config file
+.fortls
```
jatkinson1000, line 202: Is the usual place to put this file in the root of a repository?
If so I'm not against adding this here.
Do we need to add other possible locations, or should it be stored with other LSP configs?

src/ctorch.cpp
```diff
@@ -351,6 +351,17 @@ void torch_tensor_zero(torch_tensor_t tensor) {
...
+                     bool non_blocking, bool copy) {
+  auto source_t = reinterpret_cast<torch::Tensor *>(source_tensor);
+  auto dest_t = reinterpret_cast<torch::Tensor *>(dest_tensor);
+
+  torch::Device device_type = dest_t->device();
+  at::ScalarType dtype = dest_t->scalar_type();
+
+  std::move(*dest_t) = source_t->to(device_type, dtype, non_blocking, copy);
+}
+
```
jatkinson1000, line 364: Consider adding any relevant checks following #347 to this code.

src/ftorch.F90
jatkinson1000, line 2467: Perhaps this clarifies:
```suggestion
                                                         !! matches the desired device, index, dtype etc.
```

test/unit/test_tensor_manipulation_cuda.pf
jatkinson1000: > Why not just define the array in single precision to begin with?

I think we are shortcutting assigning of the values for comparison to make sure that they match the result of the conversion by torch?
If this is the case perhaps renaming `sp_data` to `sp_data_expected` might make this clearer.

jatkinson1000 (commented):
Thanks @niccolozanotti this has been a lot of work!

Left a couple of additional comments for you.

src/ftorch.F90
jwallwork23: Hm okay. So it doesn't seem like people would pass `copy=.false.` very often then?

jwallwork23: Is there a use case where you would use it?

niccolozanotti: I reflected again on this. Whereas `non_blocking` is something we want the user to have access to in the FTorch API (see [here](https://pytorch.org/tutorials/intermediate/pinmem_nonblock.html)) I think we could just use the default value for `copy=false`.
That is because in our implementation we are not exposing `to` with the original behavior (in which case my understanding is that by default we would have ended up with a `target_tensor` being simply a pointer to the `source_tensor` in case of same `dtype` and `device`) but we are getting the output of `to` and reassign it anyway:
```cpp
std::move(*target_t) = source_t->to(device_type, dtype, non_blocking, copy);
```
So my understanding is that `copy=false` would just avoid unnecessary copies at the C++ level and we can therefore keep it as default and not include it in the API.
Does this make sense?

jwallwork23: Yes, that makes sense to me. Simpler from a maintenance and docs perspective if we don't expose arguments that are unlikely to be needed.

jwallwork23: Although it'll be worth explaining the behaviour somewhere in the docs.

test/unit/test_tensor_manipulation_cuda.pf
niccolozanotti: Ah, ok thanks. I ran fortitude with `fortitude check test/unit/test_tensor_manipulation_cuda.pf` and I thought I only left out the linting errors like including the whole `funit` module

src/ftorch.F90
niccolozanotti: Yes. I should probably also add this subroutine to the relevant [page](https://cambridge-iccs.github.io/FTorch/page/tensor.html).

test/unit/test_tensor_manipulation_cuda.pf
niccolozanotti: Yes, the intended goal is to have a double precision array from which to create the source tensor and a single precision one for output comparison with the values held by `target_tensor`. 
Perhaps I could just do 
```fortran
do k = 1, 4
       do j = 1, 3
          do i = 1, 2
             dp_data(i,j,k) = real(100*i + 10*j + k, dp)
             sp_data_expected(i,j,k) = real(100*i + 10*j + k, sp)
          end do
       end do
    end do
```
for more clarity

jatkinson1000: I think as it is is OK, and makes it clear that you want a sp version of the array to compare to.
Perhaps you could move the `sp_data = real(dp_data, sp) ! source data cast to single-precision real for testing` to be just before you perform the comparison to make it clearer? And add a comment to say this is what you are doing.

src/ftorch.F90
jatkinson1000: Also agree with this, and I think I remember discussing this in the context of FTorch operation when we first looked at the extra arguments.

Agree that clarification of how FTorch differes from PyTorch or LibTorch is needed somewhere.
For this and perhaps more generally, so maybe a 'How FTorch differs from Py/LibTorch' page?

jatkinson1000: > Yes. I should probably also add this subroutine to the relevant [page](https://cambridge-iccs.github.io/FTorch/page/tensor.html).

Also noticed that the formatting of bullets is off on the linked page

niccolozanotti: Addressed in 85625b1. Will now look into documenting.

niccolozanotti: Argument removed.

src/ctorch.cpp
niccolozanotti: Addressed in 85625b1.

test/unit/test_tensor_manipulation_cuda.pf
niccolozanotti: Addressed in 85625b1.

test/unit/test_tensor_manipulation_cuda.pf
niccolozanotti: Addressed in 85625b1.

src/ftorch.F90
niccolozanotti: Addressed in 85625b1.

.gitignore
niccolozanotti: Yes it turns out you can place it also elsewhere and tell fortls its path, so I dropped it in 8790dbc.

src/ftorch.F90
niccolozanotti: I think it's better to take care of this separately so I opened #367.

src/ctorch.cpp
```diff
@@ -351,6 +351,21 @@ void torch_tensor_zero(torch_tensor_t tensor) {
   t->zero_();
 }
 
+void torch_tensor_to(const torch_tensor_t source_tensor, torch_tensor_t target_tensor,
+                     bool non_blocking) {
+  auto source_t = reinterpret_cast<torch::Tensor *>(source_tensor);
+  auto target_t = reinterpret_cast<torch::Tensor *>(target_tensor);
```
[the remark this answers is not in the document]
niccolozanotti: `source_tensor` or `target_tensor` are used already in the arguments for `torch_tensor`s though. So I'll go with `torch_tens`.

test/unit/test_tensor_manipulation_cuda.pf
```diff
@@ -0,0 +1,78 @@
...
+  subroutine test_torch_tensor_to()
+    use, intrinsic :: iso_fortran_env, only: sp => real32, dp => real64
+
+    type(torch_tensor) :: cpu_tensor, gpu_tensor, test_tensor
+    real(dp), dimension(2,3,4) :: dp_data
+    real(sp), dimension(2,3,4) :: sp_data_expected, temp_array
+    integer, parameter :: device_type_src = torch_kCPU
+    integer, parameter :: device_type_tar = torch_kCUDA
+    integer, parameter :: dtype_src = torch_kFloat64
+    integer, parameter :: dtype_tar = torch_kFloat32
```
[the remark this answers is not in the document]
niccolozanotti: Addressed in 92b2ad1.

src/ftorch.F90
```diff
@@ -2457,6 +2457,60 @@ end subroutine torch_tensor_zero_c
...
+    target_rank = target_tensor%get_rank()
+
+    if (source_rank /= target_rank) then
+      write(*,*) "Error in torch_tensor_to :: Cannot move source_tensor to target_tensor because the ranks do not match."
+      write(*,*) "Source tensor rank:", source_rank, "Target tensor rank:", target_rank
+      stop 1
+    end if
+
+    source_shape => source_tensor%get_shape()
+    target_shape => target_tensor%get_shape()
```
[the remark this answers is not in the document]
niccolozanotti: Addressed in 92b2ad1.

src/ctorch.cpp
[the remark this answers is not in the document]
niccolozanotti: 92b2ad1

jwallwork23 (approved):

PR #363, opened 2025-04-23
Tidying of wrapped functions in ftorch.F90

### LINKED ISSUE #361: Standardise the argument ordering between ftorch and ctorch
As noted in https://github.com/Cambridge-ICCS/FTorch/pull/344#discussion_r2053604020 of #344 the order of arguments (input, output) is reversed between ftorch API and ctorch API.

We should look to standardise this by changing the ctorch to match the ftorch.

jatkinson1000:
Updates to a consistent argument order and ensures wrapped subroutine arguments do not shadow outer scope.

Should close #361

### DIFF
--- a/src/ftorch.F90
+++ b/src/ftorch.F90
@@ -240,20 +240,20 @@ contains
     logical(c_bool)                 :: requires_grad_value  !! Whether gradients need to be computed for the created tensor
 
     interface
-      function torch_empty_c(ndims, tensor_shape, dtype, device_type, &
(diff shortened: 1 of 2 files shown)


jatkinson1000:
I have been through the argument order for `ftorch.F90` and everything seems consistent, so I think #361 is redundant, and the argument order for the C++ functions just needs switching to match in #344 - I'll aim to re-review that and comment directly there.

As a result I think this can be merged simply as removing name shadowing in the older wrapped C functions of `ftorch.F90`.

PR #364, opened 2025-04-24
Do not carry over `requires_grad` with overloaded ops

jwallwork23:
I don't think we should be carrying over the `requires_grad` property with overloaded operations. This should only be used for tensors that we want to differentiate w.r.t.

### DIFF
--- a/src/ftorch.F90
+++ b/src/ftorch.F90
@@ -2478,8 +2478,7 @@ contains
 
     if (.not. c_associated(output%p)) then
       call torch_tensor_empty(output, input%get_rank(), input%get_shape(), input%get_dtype(), &
-                              input%get_device_type(), device_index=input%get_device_index(), &
-                              requires_grad=input%requires_grad())
+                              input%get_device_type(), device_index=input%get_device_index())
     else if (input%get_device_type() /= output%get_device_type()) then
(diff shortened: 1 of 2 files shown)


jatkinson1000 (approved):
Thanks @jwallwork23 I think you are correct.

A quick Qn on garbage collection etc. as you know this best: When we use, say, `torch_tensor_multiply` as create a third tensor that is the result of the multiplication of the previous two, but all three of these stay in memory as part of the operation chain, is that correct?

Any tensors we declare in Fortran will persist, but can you remind me at what point does the third tensor we created during the overloaded operation go out of scope?

jwallwork23:
> When we use, say, `torch_tensor_multiply` we create a third tensor that is the result of the multiplication of the previous two, but all three of these stay in memory as part of the operation chain, is that correct?

Yes, whenever we use the overloaded assignment operator (except for in isolation, e.g., `a = b`), intermediate tensors will be created for the rvalue.

> Any tensors we declare in Fortran will persist, but can you remind me at what point does the third tensor we created during the overloaded operation go out of scope?

I would've thought that the third tensor goes out of scope when the assignment is done but we should probably check this with some memory analysis. Could be a good task for the next hackathon?

Issue #370, opened 2025-04-30
Possible bug with `requires_grad` attribute being changed following operations


jatkinson1000:
I have observed a case where the `requires_grad` attribute for a tensor is flipped from `.false.` to `.true.` as a result of assigning it based on a calculation involving other tensors.

I have created a MWE to showcase this and will open a PR to demonstrate.

The key lines are that:

```fortran
    output_vec = input_vec * scaling_tensor
    grad_reqd = loss%requires_grad()
    write(*,*) "loss%requires_grad = ", grad_reqd

    loss = ((output_vec - target_vec) ** 2)
    grad_reqd = loss%requires_grad()
    write(*,*) "loss%requires_grad = ", grad_reqd
```
cause `requires_grad` for `loss` to go from `.false.` to `.true.`.

The only tensor in the entire example that has `requires_grad=.true.` set is `scaling tensor`.

This is an issue because it prevents values being assigned to `loss` in subsequent calculations.

jatkinson1000:
I have added a MWE to the testing workflow in #371 that showcases this.

PR #371, opened 2025-04-30
DO NOT MERGE create minimum working example for requires grad bug.

jatkinson1000:
Relates to issue #370 

This provides a MWE to showcase the bug in `examples/req_grad_bug` and runs it as part of the test suite.


jwallwork23:
This can be closed now that #373 has been merged.

PR #373, opened 2025-05-02
`requires_grad` *should* be carried through expressions

### LINKED ISSUE #370: Possible bug with `requires_grad` attribute being changed following operations

jwallwork23:
Closes #370.

It turns out I was wrong in #364. Some simple Python experimentation shows that we *should* be carrying `requires_grad` over through expressions.
```
>>> a = torch.Tensor([1])
>>> a.requires_grad = True
>>> a
tensor([1.], requires_grad=True)
>>> b = torch.Tensor([2])
>>> c = a + b
>>> c
tensor([3.], grad_fn=<AddBackward0>)
>>> c.requires_grad
True
```
This PR reverts the changes from #364 and drops the related check that was causing issues.

Further, we should actually be overwriting with `requires_grad=False` when appropriate, too:
```
>>> import torch
>>> a = torch.Tensor([1])
>>> a.requires_grad=True
>>> a
tensor([1.], requires_grad=True)
>>> b = torch.Tensor([1])
>>> b
tensor([1.])
>>> a = b + b
>>> a
tensor([2.])
```

### DIFF
--- a/test/unit/test_tensor_autograd.pf
+++ b/test/unit/test_tensor_autograd.pf
@@ -40,7 +40,7 @@ module test_tensor_autograd
   end type TestParametersType
 
   ! Typedef for a test case with a particular set of parameters
-  @testCase(constructor=test_case_ctor)
+  @testCase(constructor=test_case_constructor)
   type, extends (ParameterizedTestCase) :: TestCaseType
     type(TestParametersType) :: param
   end type TestCaseType
@@ -104,11 +104,11 @@ contains
   end function get_parameters_short
 
   ! Constructor for the test case type
-  function test_case_ctor(param)
-    type(TestCaseType) :: test_case_ctor
+  function test_case_constructor(param)
+    type(TestCaseType) :: test_case_constructor
     type(TestParametersType), intent(in) :: param
-    test_case_ctor%param = param
-  end function test_case_ctor
+    test_case_constructor%param = param
+  end function test_case_constructor
 
   ! Function for representing a parameter set as a string
   function toString(this) result(string)
--- a/test/unit/test_tensor_autograd.pf
+++ b/test/unit/test_tensor_autograd.pf
@@ -28,6 +28,9 @@ module test_tensor_autograd
   integer(c_int64_t), parameter, dimension(ndims) :: tensor_shape = [2, 3]
   integer, parameter :: dtype = torch_kFloat32
 
+  ! Scalar shape for reduction operators
+  integer(c_int64_t), parameter, dimension(1) :: scalar_shape = [1]
+
   @testParameter
   type, extends(AbstractTestParameter) :: TestParametersType
     logical :: switch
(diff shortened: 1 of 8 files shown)


jwallwork23:
Added a load of testing. There are some cases that seem to be inconsistent with what I'd expect from PyTorch. (See the `NOTE`s in the unit tests.)

test/unit/test_tensor_autograd.pf
```diff
@@ -28,13 +28,128 @@ module test_tensor_autograd
...
+  @testParameter
+  type, extends(AbstractTestParameter) :: TestParametersType
+    logical :: switch
+    integer :: op
+  contains
+    procedure :: toString
+  end type TestParametersType
+
+  ! Typedef for a test case with a particular set of parameters
+  @testCase(constructor=test_case_ctor)
```
jatkinson1000, line 43:
```suggestion
  @testCase(constructor=test_case_constructor)
```
I spent a while trying to work out what `ctor` was. I would suggest perhaps naming this explicitly (doesn't feel overly verbose) here and elsewhere where used.

test/unit/test_tensor_autograd.pf
```diff
@@ -28,13 +28,128 @@ module test_tensor_autograd
...
+  end function get_parameters_short
+
+  ! Constructor for the test case type
+  function test_case_ctor(param)
+    type(TestCaseType) :: test_case_ctor
+    type(TestParametersType), intent(in) :: param
+    test_case_ctor%param = param
+  end function test_case_ctor
+
+  ! Function for representing a parameter set as a string
```
jatkinson1000, line 113: I'm not wild about this large if-else, but I have no better suggestions at the moment and it's not production code.
I can see why the feedback to stdout is useful.

test/unit/test_tensor_autograd.pf
```diff
@@ -119,5 +235,201 @@ contains
...
+      tensor3 = tensor1 + tensor2
+    else if (this%param%op == 2) then
+      tensor3 = tensor1 - tensor2
+    else if (this%param%op == 3) then
+      tensor3 = tensor1 * tensor2
+    else if (this%param%op == 4) then
+      tensor3 = tensor1 / tensor2
+    else
+      write(*,*) "Error :: invalid operator code"
+    end if
```
jatkinson1000, line 303: Note to self, I might steal this approach in #320 as a way of running tests over all optimizer types.

jatkinson1000 (approved):
Thanks for looking at this @jwallwork23 

Code wise it looks good, I just had a couple of minor suggestions.

Only other thought is that I wonder if we should explicitly document the requires grad behaviour anywhere? Whilst it is true that it 'matches that expected from PyTorch' it caught both of us out... ;)

test/unit/test_tensor_autograd.pf
jwallwork23: Ditto

test/unit/test_tensor_autograd.pf
jwallwork23: It's a fairly standard abbreviation in object-oriented programming, but sure. Done in 3eccb29f164a4a03099c484c7f6c83f20e1ea58a.

PR #374, opened 2025-05-02
Docs on overloaded assignment operator

### LINKED ISSUE #305: Docs on overloaded assignment
The segfault was resolved by ensuring `use ftorch, only : assignment(=)` was present.

So that is good news. The bad news I guess is that this can be a tricksy bug/error to spot/resolve as there isn't much to go on from the FTorch output.
Not sure there is an easy fix to this however, but definitely something to be aware and make overly clear to users of when writing the docs @jwallwork23 .

_Originally posted by @jatkinson1000 in https://github.com/Cambridge-ICCS/FTorch/issues/272#issuecomment-2695377940_

jwallwork23:
Closes #305.

Brings much-needed clarity on the overloaded assignment operator.

This PR is somewhat of a companion of #373, in that I discovered some of this stuff while working on it.

### DIFF
--- a/pages/tensor.md
+++ b/pages/tensor.md
@@ -153,6 +153,6 @@ For an example of how to compute mathematical expressions involving Torch
 tensors, see the
 [autograd worked example](https://github.com/Cambridge-ICCS/FTorch/tree/main/examples/6_Autograd).
 
-[^1]Note: In most cases, these should be the same, so that the operator makes
+[^1]: Note: In most cases, these should be the same, so that the operator makes
 sense. In the case of the `requires_grad` property, the values might differ, and
 the result should be the logical `.and.` of the two values.
--- a/pages/tensor.md
+++ b/pages/tensor.md
@@ -20,6 +20,7 @@ Fortran.
 
 We provide several subroutines for constructing `torch_tensor` objects. These
 include:
+
 * `torch_tensor_empty`, which allocates memory for the `torch_tensor`, but does
   not set any values.
 * `torch_tensor_zeros`, which creates a `torch_tensor` whose values are
@@ -47,6 +48,7 @@ other operations are meaningless.
 
 We provide several subroutines for interrogating `torch_tensor` objects. These
 include:
+
 * `torch_tensor_get_rank`, which determines the rank (i.e., dimensionality) of
   the tensor.
 * `torch_tensor_get_shape`, which determines the shape (i.e., extent in each
@@ -124,7 +126,7 @@ c = a + b
 the addition is evaluated first. It is implemented as a Fortran function and its
 return value is an *intermediate* tensor. The setup is such that this is created
 using `torch_tensor_empty` under the hood (inheriting all the properties of the
-tensors being added*). Following this, the intermediate tensor is assigned
+tensors being added[^1]). Following this, the intermediate tensor is assigned
 to `c`. Finally, the finalizer for `torch_tensor` is called for the intermediate
 tensor because it goes out of scope.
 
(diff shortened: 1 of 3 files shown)


pages/tensor.md
```diff
@@ -100,6 +100,45 @@ use ftorch, only: assignment(=), operator(+), operator(-), operator(*), &
...
+
+For overloaded operators the situation is more complex. Consider the overloaded
+addition operator (the same applies for the rest). When we execute the line
+```fortran
+c = a + b
+```
+the addition is evaluated first. It is implemented as a Fortran function and its
+return value is an *intermediate* tensor. The setup is such that this is created
+using `torch_tensor_empty` under the hood (inheriting all the properties of the
+tensors being added*). Following this, the intermediate tensor is assigned
```
niccolozanotti, line 127:
```suggestion
tensors being added[^1]). Following this, the intermediate tensor is assigned
```

pages/tensor.md
```diff
@@ -100,6 +100,45 @@ use ftorch, only: assignment(=), operator(+), operator(-), operator(*), &
...
+a `torch_tensor`, an intermediate `torch_tensor` will be created, assigned, and
+destroyed because the call will have the form
+```fortran
+a = func()
+```
+
+*Note: In most cases, these should be the same, so that the operator makes
+sense. In the case of the `requires_grad` property, the values might differ, and
+the result should be the logical `.and.` of the two values.
+
```
niccolozanotti, line 141: You could place these lines at the end of the file with 
```
[^1]: Note: In most cases, these should be the same, so that the operator makes
sense. In the case of the `requires_grad` property, the values might differ, and
the result should be the logical `.and.` of the two values.
```

pages/tensor.md
niccolozanotti, line 1: I also realized that the bullet points in the "Constructors" and "Tensor Interrogation" sections are not properly rendered so we should also fix those by adding a blank line at Lines 23 and 50 😃

niccolozanotti (changes requested):
The content looks good and very much needed @jwallwork23 ! Just left a couple of small comments.

Up to you if using the footnote rather than the `*Note` and the end of the section (maybe that's the approach you've decided to take elsewhere)

jwallwork23:
Thanks very much for the review, @niccolozanotti. Addressed your suggestions in fa17047.

pages/tensor.md
jwallwork23: Done in fa17047.

pages/tensor.md
jwallwork23: Done in fa17047.

pages/tensor.md
jwallwork23: Done in fa17047.

niccolozanotti: Thanks. In order to have the foot note properly rendered
https://github.com/Cambridge-ICCS/FTorch/blob/fa17047b4f92d87f284c0961c67beb17885d723d/pages/tensor.md?plain=1#L156
should be
```
[^1]: Note: ...

File: pages/tensor.md (hunk 2 above)
Comment (niccolozanotti) [reply pages/tensor.md]: Thanks. In order to have the foot note properly rendered
https://github.com/Cambridge-ICCS/FTorch/blob/fa17047b4f92d87f284c0961c67beb17885d723d/pages/tensor.md?plain=1#L156
should be
```
[^1]: Note: ...

jwallwork23: Ah, my bad. Fixed in a9bd785.

niccolozanotti (approved):
Looks good to me.

Issue #365, opened 2025-04-24
Error when calling `torch_tensor_from_array` on GPU


TomMelt:
A new error has been introduced in commit 864b0ebaada. When calling `torch_tensor_from_array` on GPU e.g.,
```fortran
call torch_tensor_from_array(input_tensor(1), input_data_2d, layout_2d, torch_kCUDA)
```
We get the following error:
```
[ERROR]: The specified pointer resides on host memory and is not registered with any CUDA device.
```

It appears it is linked to the following lines in `src/ctorch.cpp`:
https://github.com/Cambridge-ICCS/FTorch/blob/df8cfd634de6f2792a59531c97bb5c72ea3deed7/src/ctorch.cpp#L275-L279

It will most likely also affect `torch_empty`, `torch_zeros` and `torch_ones` as well.

jatkinson1000:
Thanks for raising this @TomMelt, and even more thanks for tracking down the commit and lines!!

Do you have any idea as to what the cause/fix might be?
I see that that commit started using `tensorOptions` rather than a chained call to `.to()`.
I have not been able to look too deeply, but came across this in the C++ docs:
https://github.com/pytorch/pytorch/blob/main/c10/core/TensorOptions.h#L100-L127 which may or may not be relevant.

I guess a quick fix might be to revert back to using `.to()`(?), though it would be good to understand why the device from the options is not working.

niccolozanotti:
I think the problems lies in that `torch::from_blob(data, vshape, vstrides, options)` expects data to *already* reside on the `device` passed as part of `options`. (See [here](https://docs.pytorch.org/cppdocs/api/function_namespacetorch_1ac009244049812a3efdf4605d19c5e79b.html)).

Thus we cannot avoid calling `to(get_libtorch_device(device_type, device_index))`

PR #377, opened 2025-05-21
Fix bug in implementation of wrapper function `torch_from_blob` (#365)

niccolozanotti:
This would patch #365. 

As discussed in #365 this PR would fix the bug that sneaked in at 864b0eb causing any call to `torch_tensor_from_array` on a gpu device to throw
```
[ERROR]: The specified pointer resides on host memory and is not
registered with any CUDA device.
```

It rolls back to previous approach (6deab9a) where the device is not explicitly passed to the `options` but the tensor is rather moved `to` the target device.

### DIFF
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -272,6 +272,9 @@ torch_tensor_t torch_from_blob(void *data, int ndim, const int64_t *shape,
     // This doesn't throw if shape and dimensions are incompatible
     c10::IntArrayRef vshape(shape, ndim);
     c10::IntArrayRef vstrides(strides, ndim);
+    // NOTE: Do not pass device TensorOptions since this would cause torch::from_blob
+    //       to expect data to reside on the host device_type, which is not case if
+    //       device_type is a GPU device (see #365)
     auto options = torch::TensorOptions()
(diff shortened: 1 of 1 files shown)


jatkinson1000 (changes requested):
Hi @niccolozanotti thanks for looking in to this, and I am happy with the implementation as discussed yesterday in-person.

However, I suggest a short comment in-code clarifying that device is not part of `TensorOptions` for a reason to stop this creeping back in at a future point.

My other question would be - when we are creating a tensor on the same device as which the data resides, what happens when calling `to`?
Does this invoke a copy?
If you can clarify this then I think we are OK to merge.

niccolozanotti:
Thank you @jatkinson1000.

As for the behavior of `to` in the case of an already matching conversion, it's a simular discussion to the one we had in #360. 

The [docs](https://docs.pytorch.org/docs/stable/generated/torch.Tensor.to.html) point out that if `copy=False` - which is our case since we don't pass that explicitely - then no copy is made if both `device` and `dtype` happen to be the same (`dtype` is not passed in our case so we only care about the device).

jatkinson1000 (approved):
OK, thanks @niccolozanotti, I think I'm happy with this now.

PR #383, opened 2025-06-01
Trivial example text tweaks

jwallwork23:
Spotted a few things when copying over READMEs for the workshop.


jwallwork23:
> Surprised example 8 worked with a missing operator, or maybe it didn't...

It wasn't missing one; it had one it didn't actually need.

PR #385, opened 2025-06-03
feat: add HIP support to cmake using CUDA backend

TomMelt:
This fixes #128 

AMD does have PyTorch built with explicit `HIP` support. But this does not currently work.

<details><summary>Error Message</summary>
<p>

```
[ERROR]: Could not run 'aten::empty_strided' with arguments from the 'HIP' backend. This could be because the operator doesn't exist for this backend, or was omitted during the selective/custom build process (if using custom build). If you are a Facebook employee using Py
Torch on mobile, please visit https://fburl.com/ptmfixes for possible resolutions. 'aten::empty_strided' is only available for these backends: [CPU, CUDA, Meta, QuantizedCPU, QuantizedCUDA, BackendSelect, Python, FuncTorchDynamicLayerBackMode, Functionalize, Named, Conjug
ate, Negative, ZeroTensor, ADInplaceOrView, AutogradOther, AutogradCPU, AutogradCUDA, AutogradHIP, AutogradXLA, AutogradMPS, AutogradIPU, AutogradXPU, AutogradHPU, AutogradVE, AutogradLazy, AutogradMTIA, AutogradPrivateUse1, AutogradPrivateUse2, AutogradPrivateUse3, Autog
radMeta, AutogradNestedTensor, Tracer, AutocastCPU, AutocastMTIA, AutocastXPU, AutocastMPS, AutocastCUDA, FuncTorchBatched, BatchedNestedTensor, FuncTorchVmapMode, Batched, VmapMode, FuncTorchGradWrapper, PythonTLSSnapshot, FuncTorchDynamicLayerFrontMode, PreDispatch, Pyt
honDispatcher].
```

</p>
</details> 

Instead PyTorch recommends using the `CUDA` backend (see [here](https://docs.pytorch.org/docs/stable/notes/hip.html))

> PyTorch for HIP intentionally reuses the existing [torch.cuda](https://docs.pytorch.org/docs/stable/cuda.html#module-torch.cuda) interfaces. This helps to accelerate the porting of existing PyTorch code and models because very few code changes are necessary, if any.

This PR add minimal change to `CMakeLists.txt` to build with the `CUDA` backend, even if the user picks `HIP`

PR #386, opened 2025-06-03
Add `torch_kHIP` and `hip` device type for `pt2ts`

jwallwork23:
Merges into #385.

Some extras to add a `torch_kHIP` enum (aliased to `torch_kCUDA`) and to accept a `hip` device type option for the `pt2ts.py` scripts.


TomMelt (approved):
Thanks @jwallwork23 this makes it more consistent

PR #388, opened 2025-06-03
Update docs with information about HIP.

### LINKED ISSUE #387: Docs for HIP GPU Backend
#385 made a great step in adding support for AMD/HIP backend.

However, there were no documentation updates as part of this PR.

These should be added for users:

- Updates to repository README
- Updates to [online GPU Support pages](https://cambridge-iccs.github.io/FTorch/page/gpu.html)

Plus anywhere else as may seem relevant.

jatkinson1000:
Closes #387 

- [x] Updates README to reference HIP
- [x] Updates GPU support page to reference HIP


jatkinson1000:
One area that it occurs to me that might be useful is to specify to users which binary is required for which backend (CUDA vs Rocm), and also how to get them in the case of MPS and XPU (I think a pytorch install is the only way to do this at present?).

Let me know your thoughts and I can add this if we want.

jwallwork23 (changes requested):
Thanks for this @jatkinson1000. The changes look good but the suggestions you made seem worthwhile.

> One thing it occurs to me that might be useful is to specify to users which binary is required for which backend (CUDA vs Rocm), and also how to get them in the case of MPS and XPU (I think a pytorch install is the only way to do this at present?).

Yes, this would be a valuable addition to the GPU docs page.

> Also whether there should be any explanation of HIP mapping to CUDA in LibTorch?
Though I personally feel that the in-code docs are sufficient and adding anything to the FTorch docs might only confuse users.

If we're mentioning the binaries (above) then we should mention this too, otherwise people might wonder why we specify CUDA PyTorch backend for HIP. It doesn't need to be a long explanation, just a mention that we alias HIP to CUDA and a link to the page in the PyTorch docs on this.

README.md
```diff
@@ -234,10 +234,10 @@ you wish to target (using the `GPU_DEVICE` CMake option as detailed in the
 1. When saving a TorchScript model, ensure that it is on the appropriate GPU
    device type. The `pt2ts.py` script has a command line argument
    `--device_type`, which currently accepts four different device types: `cpu`
```
TomMelt:
```suggestion
   `--device_type`, which currently accepts 5 different device types: `cpu`
```

jwallwork23 (approved):
Great! Thanks @jatkinson1000.

TomMelt (approved):
🚀 LGTM

PR #392, opened 2025-07-01
Optimisation demo: checks and plotting

jwallwork23:
Merges into #320.

This PR adds a few nice things for the optimisation example:
* Check both the Python and Fortran implementations converge to the expected array [1,2,3,4].
* Track the convergence of the loss function for both implementations and write these to files.
* Add a `ctest` check that the plotting script can be run and produces a PNG file.

### DIFF
--- a/examples/n_Optimizers/optimizers.f90
+++ b/examples/n_Optimizers/optimizers.f90
@@ -16,6 +16,9 @@ program foptimizer
                     torch_tensor_mean
   use ftorch_optim, only: torch_optim, torch_optim_SGD
 
+  ! Import our tools module for testing utils
+  use ftorch_test_utils, only : assert_allclose
+
   implicit none
 
   ! Set working precision for reals
@@ -24,7 +27,7 @@ program foptimizer
   ! Set up Fortran data structures
   integer, parameter :: ndims = 1
   integer, parameter :: n=4
-  real(wp), dimension(n), target :: input_data, output_data, target_data
+  real(wp), dimension(n), target :: input_data, output_data, target_data, scaling_data
   real(wp), dimension(1), target :: loss_data
   integer :: scalar_layout(1) = [1]
   integer :: tensor_layout(ndims) = [1]
@@ -46,8 +49,8 @@ program foptimizer
   call torch_tensor_from_array(target_vec, target_data, tensor_layout, torch_kCPU)
 
   ! Initialise Scaling tensor as ones as in Python example
-  call torch_tensor_ones(scaling_tensor, ndims, tensor_shape, &
-                         torch_kFloat32, torch_kCPU, requires_grad=.true.)
+  call torch_tensor_from_array(scaling_tensor, scaling_data, tensor_layout, torch_kCPU, &
+                               requires_grad=.true.)
 
   ! Initialise an optimizer and apply it to scaling_tensor
   ! TODO optimizer expects an array of tensors, should be a cleaner consistent way to formalise this.
@@ -104,6 +107,12 @@ program foptimizer
     call torch_delete(output_vec)
   end do
   close(unit=10)
+
+  ! Check scaling tensor converges to the expected value
+  if (.not. assert_allclose(scaling_data, target_data, test_name="optimizers", rtol=1e-3)) then
+    write(*,*) "Error :: value of scaling_data does not match expected value"
+    stop 999
+  end if
 
   write(*,*) "Training complete."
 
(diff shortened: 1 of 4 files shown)


jwallwork23:
losses

For reference, this is what `losses.png` looks like for me. PyTorch and FTorch have exactly the same behaviour, as we might hope!

jwallwork23:
For reference, this is the plot I get. PyTorch and FTorch have exactly the same behaviour, as we might hope!

losses

jwallwork23:
> Hey @jwallwork23 could you rebase on `optim` to make this cleaner and pick up latest changes, then we can hopefully merge.
> 
> I'm happy with the png if you are - though suggest making the label-size of the `X` larger, or plotting them at the top level, so it is clear that both are on there.

Thanks @jatkinson1000. Reordered the plot symbols so that the cross is on top of the circle. The rebase picked up some slight differences and I propose some minor rewordings of the NOTE comments, which you might want to check you're happy with.

.gitignore
```diff
@@ -196,4 +196,7 @@ cython_debug/
 *.pt
 
 # Apple Desktop Services Store
-**.DS_Store
\ No newline at end of file
+**.DS_Store
+
+# Example outputs
+examples/n_Optimizers/losses.png
```
jatkinson1000: Is it worth just ignoring `*.png` files and explicitly adding when we need given this path will be changing soon?

examples/n_Optimizers/expected_losses.png
jatkinson1000, line 1: Your plotting changes don't seem to be reflected in this file (unless it's a caching problem...)

examples/n_Optimizers/optimizers.f90
```diff
@@ -45,32 +48,40 @@ program foptimizer
...
+  scaling_data(:) = 1.0_wp
+  call torch_tensor_from_array(scaling_tensor, scaling_data, tensor_layout, torch_kCPU, &
+                               requires_grad=.true.)
 
   ! Initialise an optimizer and apply it to scaling_tensor
   ! NOTE: The optimizer expects an array of tensors.
   call torch_optim_SGD(optimizer, [scaling_tensor], learning_rate=1D0)
 
+  ! Create an empty tensor for the gradient of the scaling
+  call torch_tensor_empty(scaling_grad, ndims, tensor_shape, torch_kFloat32, torch_kCPU)
```
jatkinson1000: Would suggest leaving this next to the creation of scaling tensor (lines 55-56) instead of moving here, unless you had a reason?

examples/n_Optimizers/optimizers.f90
```diff
@@ -45,32 +48,40 @@ program foptimizer
...
   call torch_tensor_from_array(target_vec, target_data, tensor_layout, torch_kCPU)
 
-  ! Initialise Scaling tensor as ones as in Python example and a tensor for its gradient
-  call torch_tensor_ones(scaling_tensor, ndims, tensor_shape, &
-                         torch_kFloat32, torch_kCPU, requires_grad=.true.)
-  call torch_tensor_empty(scaling_grad, ndims, tensor_shape, torch_kFloat32, torch_kCPU)
+  ! Initialise Scaling tensor as ones as in Python example
+  scaling_data(:) = 1.0_wp
+  call torch_tensor_from_array(scaling_tensor, scaling_data, tensor_layout, torch_kCPU, &
+                               requires_grad=.true.)
```
jatkinson1000, line 54: Thanks for spotting this needs to be a layout!

jwallwork23: Ah sorry, must've picked the wrong change in the rebase. Fixed in 5ab663aca6ddd3e01f3fdadd70e8810a036a29a5.

examples/n_Optimizers/optimizers.f90
joewallwork: Ah sorry, must've picked the wrong change in the rebase. Fixed in 5ab663aca6ddd3e01f3fdadd70e8810a036a29a5.

.gitignore
joewallwork: Done in 5ab663aca6ddd3e01f3fdadd70e8810a036a29a5.

examples/n_Optimizers/expected_losses.png
joewallwork: My bad - forgot to update it. Done in 5ab663aca6ddd3e01f3fdadd70e8810a036a29a5.

PR #395, opened 2025-07-02
Fig bug that made Fortran source is missing from the API docs

### LINKED ISSUE #394: Fortran source is missing from the API docs
Looking at https://cambridge-iccs.github.io/FTorch/lists/files.html I don't see ftorch.F90

jwallwork23:
Fixes #394.

The issue was that we hadn't told Ford about the HIP macro but I've included other updates to refer to the correct `.F90` extension for the main module.

The fix works locally for me.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -10,23 +10,22 @@ set -eo pipefail
 
 # ===================================================================
 
-if git rev-parse --verify HEAD >/dev/null 2>&1
-then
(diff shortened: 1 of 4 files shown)


jatkinson1000 (approved):
Brilliant, thanks for the rapid fix @jwallwork23 

As discussed, we ought to try and improve the CI to catch any docs build failures.

PR #397, opened 2025-07-03
Optim: docs

jwallwork23:
Merges into #320.

This PR updates the online training docs for the `optim` branch.

## Checklist

- [x] Add schematic for running training in Fortran like we have for the offline training docs.


jatkinson1000 (commented):
Hi @jwallwork23 I think this needs a rebase as I believe I am now happy with the process for conducting training in a loop with appropriate tensor creation (see latest commits on `optim` branch).

Based on what I found I think there needs to be some clearer guidance on when and where to create tensors as part of a training process (again see notes on recent commits). Not sure if you want to do that here or would prefer to merge this and let me write it on #320 

I also wonder if we want to restructure documentation - is it worth having separate pages for things like optimizers and loss functions? It might also be nice to have someexamples, like lib/PyTorch do, though we'd want to be sure that these stayed in sync with he source.

jwallwork23:
> Hi @jwallwork23 I think this needs a rebase as I believe I am now happy with the process for conducting training in a loop with appropriate tensor creation (see latest commits on `optim` branch).

Done.

> Based on what I found I think there needs to be some clearer guidance on when and where to create tensors as part of a training process (again see notes on recent commits). Not sure if you want to do that here or would prefer to merge this and let me write it on #320 

I think I'll let you write it if that's okay?
 
> I also wonder if we want to restructure documentation - is it worth having separate pages for things like optimizers and loss functions? It might also be nice to have someexamples, like lib/PyTorch do, though we'd want to be sure that these stayed in sync with he source.

Yeah that'd be good. Although perhaps this restructure might be better done as part of a wider docs overhaul after merging #320? I think I've mentioned before that the user guide section of the webpage has a lot of useful info, but is fairly unstructured and could do with some reworking.

pages/online.md
```diff
@@ -2,17 +2,61 @@ title: Online training
...
+
+#### 2. pt2ts
+
+The scripting section comes earlier in the online workflow. Having written a
+model to a file with `.pt` extension, use the `pt2ts.py` utility Python script
+to convert it to TorchScript format. A template `pt2ts.py` script can be found
+in the [`utils`](https://github.com/Cambridge-ICCS/FTorch/tree/main/utils)
+subdirectory. See the
+[README](https://github.com/Cambridge-ICCS/FTorch/blob/main/utils/README.md)
+there for more details on how to use the script.
```
jatkinson1000, line 38: As an internal note, have we tried this, as we had some discussion about whether it was possible to train `jit`ted models or not, and whether models would need to be designed in Fortran. Happy to leave this and have it updated at a later date.

pages/online.md
```diff
@@ -2,17 +2,61 @@ title: Online training
...
+
+The training code should be set up such that the TorchScript model is both read
+in and written out.
+
+#### 4. Fortran model with inference
+
+In order to run inference with the trained ML model, you will need to create
+another modified version of your Fortran model that loads the TorchScript model
+and uses FTorch syntax to set up appropriate `torch_tensor` and `torch_model`
+objects and call the `torch_tensor_forward` subroutine to run the inference. For
```
jatkinson1000: `torch_model_forward`?
```suggestion
objects and call the `torch_model_forward` subroutine to run the inference. For
```

pages/online.md
```diff
@@ -2,17 +2,61 @@ title: Online training
...
+
+In the online case, data generation and training are done in the same step,
+purely in Fortran. Code modifications are made so that we can run the Fortran
+model to generate training data and immediately use this training data (whilst
+in memory) to train the ML model. There is not necessarily an optimization loop
+in this case - one option is to take a pre-trained model and to continue
+improving it using data generated online.
+
+The training code should be set up such that the TorchScript model is both read
+in and written out.
```
jatkinson1000: Not immediately clear to me what this means?
At the start and end of the simulation?
And is this a hypothetical that is waiting for us to implement `jit`ting and saving of a model from Fortran?

pages/online.md
```diff
@@ -123,10 +167,26 @@ Note that `torch_tensor_get_gradient` must be called after every call to
...
+Fortran as well as in Python. To make use of optimizers in FTorch, you need the
+`torch_optim` derived type. This derived type has two member subroutines as
+follows:
+* `torch_optim%zero_grad`, which zeroes all tensors associated with the
+  optimizer. This should be called at the beginning of every step of the
+  optimization loop.
+* `torch_optim%step`, which takes an iteration of the optimization method.
 
-*Not yet implemented.*
+The [optimizer worked example](https://github.com/Cambridge-ICCS/FTorch/tree/main/examples/n_Optimizers)
```
jatkinson1000, line 187: I'll try and remember to update this...

pages/online.md
joewallwork: No because we haven't set up training in Fortran yet.

joewallwork: As in training of the ML model. They haven't been exposed yet, right? I guess maybe should add a clarifying note.

jatkinson1000: Cool, that's what I thought. As I say, happy to leave and clarify at a later date. I'll leave up to you whether you want to retain a `not yet implemented/tested`.

pages/online.md
joewallwork: You're right, this needs to be clearer. I was thinking of having a pre-trained model that's been converted to TorchScript, which gets picked up by the Fortran code, tweaked, and written back to the same file. Yeah I guess it's somewhat hypothetical at this stage.

joewallwork: Okay hopefully e4dfb382b97e954505a794fa8680902057cdc33a makes it clearer.

jatkinson1000 (approved):
Yes, I'm happy with this now, thanks @jwallwork23 

If you want to merge now I don;t think it should conflict with my local changes to unit tests, so can stash and pull locally.

PR #398, opened 2025-07-03
Fix unused-variable warning

Mikolaj-A-Kowalski:
A micro change that removes local variable that was indeed unused.


jatkinson1000 (approved):
Great stuff, thanks for contributing @Mikolaj-A-Kowalski

PR #400, opened 2025-07-04
Add detach functionality for tensors and use in optimizers example.

jatkinson1000:


jatkinson1000:
Closing this as somewhat stale and no longer a solution to the problem it was trying to solve.

If we decide we want the `detach` functionality we can return to reimplement based on the relevant commits here.

PR #401, opened 2025-07-07
Move AOPP tutorial to presentation archive

jwallwork23:
Moves the event from upcoming events to the presentation archive. Reorders past trainings in reverse order, consistently with the past talks.


jatkinson1000 (approved):
Yep, sure

PR #403, opened 2025-07-10
Update FPTLIB-TODO to bare TODO

jatkinson1000:
Done throughout pt2ts files and associated READMEs in main utilities and exercises.

This matches a change made in the workshop https://github.com/Cambridge-ICCS/FTorch-workshop/pull/40


jatkinson1000:
> Thanks. Perhaps we should mention that these `TODO`s are to be completed by the user in the docstring at the top of the file? Otherwise people might just think we have various plans for improvements of the script.

That was originally the reason for `FTORCH-TODO` IIRC. I suspect a comment could be missed, so would be in favour of reverting to `FTORCH-TODO` to clarify? Thoughts?

jwallwork23 (approved):

PR #405, opened 2025-07-14
Make FORD fail on warnings

Mikolaj-A-Kowalski:
Turns out by default FORD runs in a debug mode :upside_down_face:. 
So it prints a message to a console on some exceptions. 

Switching it off should make it fail. It is still draft since I want to test if the warning due to missing `graphviz` will also trip the build.

### DIFF
--- a/FTorch.md
+++ b/FTorch.md
@@ -15,6 +15,7 @@ extra_filetypes: c   //
                  h //
                  py  #
 preprocess: true
+dbg: false
 macro: UNIX
        GPU_DEVICE_NONE=0
        GPU_DEVICE_CUDA=1
@@ -117,7 +118,7 @@ distributed under an [MIT License which can be viewed here](page/LICENSE.html).
 Projects using FTorch
 ---------------------
 
-The following projects make use of FTorch.  
+The following projects make use of FTorch.<br>
 If you use our library in your work please let us know.
 
 * [DataWave CAM-GW](https://github.com/DataWaveProject/CAM/) -
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -27,6 +27,10 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
+      - name: Install Graphviz (Required by FORD)
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y graphviz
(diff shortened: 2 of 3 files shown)


Mikolaj-A-Kowalski:
Should be ready now. 

I have added `graphviz` to the doc CI jobs. An alternative would be to disable graphs explicitly in FORD (but personally I don't see why not include them).

FTorch.md
```diff
@@ -117,7 +118,7 @@ distributed under an [MIT License which can be viewed here](page/LICENSE.html).
 Projects using FTorch
 ---------------------
 
-The following projects make use of FTorch.  
+The following projects make use of FTorch.
```
jatkinson1000: I believe this `  ` is here to enforce a line break in the markdown, so suggest putting back.
Or you can see if a `<br>` or `\` works, though I recall issues with the parser FORD uses.
```suggestion
The following projects make use of FTorch.  
```

FTorch.md
```diff
@@ -15,6 +15,7 @@ extra_filetypes: c   //
                  h //
                  py  #
 preprocess: true
+dbg: false
```
jatkinson1000, line 18: Interesting...

This is great, but please add a note to the "docs" docs to note how users can revert to debug mode when building and developing locally.
These should be at `pages/developer.md`

jatkinson1000 (changes requested):
Thanks for tackling this @Mikolaj-A-Kowalski 2 small requests prior to merging.

FTorch.md
Mikolaj-A-Kowalski: I have fallen victim to automated whitespace trimming on save :grimacing:  To be honest I was not aware of this feature of Markdown!

I switch to using `<br>`. The FORD seems to be happy with it on my side, it is used above in the 'Presentations' section.

FTorch.md
Mikolaj-A-Kowalski: I have just pushed an update with an extra paragraph about it.

FTorch.md
jatkinson1000: Cool, we should ideally revert to using this throughout the docs, but not in this PR.

FTorch.md
jatkinson1000: Thanks.

jatkinson1000 (approved):
Thanks @Mikolaj-A-Kowalski I'm happy for this to be merged now.

PR #407, opened 2025-07-21
Prevent using system Python despite venv check

Mikolaj-A-Kowalski:
Previously a system Python could be used despite the protection if one: 
1. Tried to run the `cmake` without venv
2. Activated venv after seeing the error message
3. Re-run the cmake without cleaning cache

Now check for venv is done before any search for Python so nothing will be cached.

I have also used this opportunity to change the `Python_FIND_VIRTUALENV` to `ONLY` which is I believe the behaviour we want (never pick-up the system interpreter in any case?)


CMakeLists.txt
```diff
@@ -157,19 +157,21 @@ install(FILES "${CMAKE_BINARY_DIR}/modules/ftorch_test_utils.mod"
...
-  find_package(
-    Python
-    COMPONENTS Interpreter
-    REQUIRED)
-
   # Check if Python is in a virtual environment by checking the VIRTUAL_ENV
   # environment variable exists
+  # Must be done before `find_package` in to prevent system Python getting
+  # cached by CMake
   if(NOT DEFINED ENV{VIRTUAL_ENV})
```
Mikolaj-A-Kowalski, line 164: We should probably make it work with `conda` environment as well? 
But for that we will need to increase minimum CMake version 3.17 since this is when `FindPython` `FIND_VIRTUALENV` got support for [`conda`](https://cmake.org/cmake/help/latest/module/FindPython.html#hints).

Mikolaj-A-Kowalski:
I was thinking that perhaps the cleanest and safest check for the virtual environment would be to set the 
[`PIP_REQUIRE_VEN`](https://pip.pypa.io/en/stable/cli/pip/#cmdoption-require-virtualenv) in the examples `CMakeList.txt`s?

CMakeLists.txt
joewallwork: I'd be fine with bumping the minimum version to 3.17. We could also plausibly increase the maximum version given how many more releases there have been since 3.31 https://cmake.org/cmake/help/latest/release/index.html

jatkinson1000: I think that's fine, but can we ideally have a separate commit and definitely a Changelog entry for it please.

Looking at the breaking changes for 4.0 we should be unaffected.

PR #408, opened 2025-07-21
Mention `graphviz` dependency of FORD in the docs

niccolozanotti:
This would complement the improvements done in #405.

### DIFF
--- a/pages/developer.md
+++ b/pages/developer.md
@@ -223,6 +223,10 @@ FORD is pip installable:
 ```
 pip install ford
 ```
+FORD uses [graphviz](https://graphviz.org) to generate dependency graphs from the Fortran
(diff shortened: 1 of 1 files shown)


jatkinson1000 (approved):
I'm happy with this @niccolozanotti but can you just check that the rendering of the footnote is done correctly before you merge please - I have noticed in the past that the FORD Markdown engine does not support a number of common features of the language (e.g. `\` linebreaks).

niccolozanotti:
Yes @jatkinson1000 , it does render correctly. We have introduced some footnotes in the docs recently, see e.g. [here](https://cambridge-iccs.github.io/FTorch/page/tensor.html#fnref:1) :)

PR #409, opened 2025-07-22
DO NOT MERGE - Add Github `GPU-runner` test workflow

niccolozanotti:
Opening this now to have the workflow registered so a member of the repo can trigger it.

cc @jatkinson1000 

---

This adds a minimal example workflow to test the new GPU runner and print out relevant information.

Towards #369.

PR #410, opened 2025-07-22
Add CI for test suite on GPU runner | Add `MULTI_GPU` cmake option

### LINKED ISSUE #369: Setup CI workflow for CUDA tests
This is now a high priority because a bug snuck in without us noticing, as reported in #365.

niccolozanotti:
After testing the [GPU runner](https://docs.github.com/en/enterprise-cloud@latest/actions/reference/larger-runners-reference#specifications-for-gpu-larger-runners) in #409 I am now looking into this.

A few questions worth discussing came up, especially related to the triggers we want to have for the testing workflow due to its running [cost](https://docs.github.com/en/billing/managing-billing-for-your-products/about-billing-for-github-actions#per-minute-rates-for-gpu-powered-larger-runners).

Following the discussion with @jatkinson1000 it seems sensible to have
* a manual trigger
* a recurrent `schedule`d trigger (1 per month?)
* a trigger on push, i.e. PR merge, only when certain `path`s are changed, which I currently handpicked [here](https://github.com/Cambridge-ICCS/FTorch/blob/369_CI_workflow_CUDA_test/.github/workflows/test_suite_ubuntu_cuda.yml#L14-L30)

Another point worth discussing is that we have set up the testing for both `fortran2008` and `2018` via the `job.strategy.matrix`. Despite being non-ideal for testing purposes we could just choose one of the two language standards for what concerns this gpu testing to save a bit and have the other CPU-only tested (I wouldn't expect GPU specific problems to arise in one language standard over the other).

Closes #369.


niccolozanotti:
Update from the [latest run](https://github.com/Cambridge-ICCS/FTorch/actions/runs/16451952556/job/46499221721):
* all the unit tests are passing as expected
* integration tests pass with the exception of the `MultiGPU` (as expected since we just have 1 GPU device

.github/workflows/test_suite_ubuntu_cuda.yml
```diff
@@ -0,0 +1,130 @@
...
+# Controls when the workflow will run
+on:
+
+  schedule:
+    # Run monthly (see https://crontab.guru/#00_00_1_*_*)
+    - cron: '00 00 1 * *'
+
+  # Triggers the workflow on pushes to the "main" branch, i.e., PR merges
+  push:
+    branches: [ "main" ]
```
jatkinson1000: If it is run on all merges to main then there should be no need for a cron-job.
Though it's perhaps not the worst to leave in given current low cost.

.github/workflows/test_suite_ubuntu_cuda.yml
```diff
@@ -0,0 +1,130 @@
...
+# Workflow run - one or more jobs that can run sequentially or in parallel
+jobs:
+  test-suite-ubuntu-cuda:
+    # The type of runner that the job will run on
+    runs-on: GPU-runner
+    timeout-minutes: 10
+    strategy:
+      fail-fast: true
+      matrix:
+        std: ["f2008"]
```
jatkinson1000: Only one item so remove matrix and explicitly declare below?
Though would be open to leaving like this for matching to other workflows.

.github/workflows/test_suite_ubuntu_cuda.yml
```diff
@@ -0,0 +1,130 @@
...
+      - name: Install cmake and NVIDIA dev toolkit
+        run: |
+          sudo apt update
+          sudo apt install -y cmake nvidia-cuda-toolkit
+
+      # MPI is required by pFUnit
+      - name: Install an MPI distribution
+        run: |
+          sudo apt install -y openmpi-bin openmpi-common libopenmpi-dev
+
```
jatkinson1000: Based on work elsewhere I don't believe MPI _is_ required which would cut down the time required.

As far as I know we only use `funit` in the tests.

Obviously check this though!!

niccolozanotti:
Thanks for the comments @jatkinson1000. 

> Based on these options I think perhaps having a `TEST_MULTIPLE_GPU` flag for CMake that is passed at build time and used (along with `GPU_DEVICE`) to trigger whether the multi-GPU example is built in the `examples/CMakeLists.txt` is the best approach.

This is what I thought as the most desirable solution as well. This is worth a separate PR.

> It might be possible to disable CPU tests with some clever logic, but at the moment I'm not sure it's worth pursuing this
Yeah, requires some thought. 

One thing we could do right off the bat would be to just run the `f2018` standard in the CPU testing for the cases in which the GPU workflow is run.  
The GPU run just adds the testing pool so it should catch any issue in the `f2008` anyway.

.github/workflows/test_suite_ubuntu_cuda.yml
niccolozanotti: This is true for unit tests. However [Example 7_MPI](https://github.com/Cambridge-ICCS/FTorch/tree/main/examples/7_MPI) requires it so we should keep it.

.github/workflows/test_suite_ubuntu_cuda.yml
niccolozanotti: I just kept it as it was relatively easy to going back. But now that you have agreed on testing only `f2008` it's cleaner without it

.github/workflows/test_suite_ubuntu_cuda.yml
niccolozanotti: Yeah that's true. I mean, FTorch is - hopefully forever 😄  - a quite active project. So we should be good with the manual and the push trigger

.github/workflows/test_suite_ubuntu_cuda.yml
jatkinson1000: Ahhhh, yes, OK.
The workflow states that is it needed for pFUnit so we should maybe amend this at some point.

.github/workflows/test_suite_ubuntu_cuda.yml
niccolozanotti: Removed in https://github.com/Cambridge-ICCS/FTorch/pull/410/commits/c0626bfdfe9722e28d88dce24f70011da4775d0d.

.github/workflows/test_suite_ubuntu_cuda.yml
niccolozanotti: Yes, indeed. I modified that in [4f859db](https://github.com/Cambridge-ICCS/FTorch/pull/410/commits/4f859dbb9daca208cfe710d269404664411b03b8). Building pFUnit without MPI support maybe skips some tests that save us time

niccolozanotti: Actually it's about 6-7 seconds difference from the run that just happened.

CMakeLists.txt
```diff
@@ -170,6 +181,10 @@ if(CMAKE_BUILD_TESTS)
             "No Python virtual environment detected. Please activate one.")
   endif()
 
+  if(MULTI_GPU)
+    message(STATUS "Building tests with Multi GPU support")
+  endif()
+
```
jatkinson1000: I wonder if a "Building with GPU support" would also be useful feedback to the user?

CMakeLists.txt
```diff
@@ -57,6 +57,17 @@ else()
...
+# to determine whether to run Multi GPU testing
+if(NOT "${GPU_DEVICE}" STREQUAL "NONE")
+  option(MULTI_GPU "Whether to enable Multi GPU testing (OFF [default], ON)"
+         OFF)
+else()
+  if(MULTI_GPU)
+    message(SEND_ERROR "Multi GPU turned ON but no GPU device specified")
+  endif()
+endif()
+
```
jatkinson1000: I think I would be inclined to have it set `ON` by default.
This would match the current behaviour retaining backwards compatibility.

In use the reason to specify this option is to turn multiGPU `OFF` for testing, so I think it should be a positive action to turn it off, default to on.

niccolozanotti: Yes, I was a bit in doubt on this. Better to change the default behavior as little as possible

niccolozanotti:
> Sorry, started reviewing before I saw your comment @niccolozanotti so feel free to apply these suggestions on #413
> 
> You're suggesting merging 413 to `main`, right?

No worries, thanks for the review. At this point I feel a bit torn about where to go on. I was intending to merge #413 here but maybe it makes sense to direct that to main

.gitignore
```diff
@@ -182,6 +182,19 @@ dmypy.json
...
+CMakeFiles
+CMakeScripts
+Testing
+cmake_install.cmake
+install_manifest.txt
+compile_commands.json
+CTestTestfile.cmake
+_deps
+CMakeUserPresets.json
+
```
jatkinson1000, line 197: Surprised this wasn't already there tbh!

CMakeLists.txt
```diff
@@ -36,27 +36,36 @@ option(GPU_DEVICE
...
-  # > because very few code changes are necessary, if any."
-  # Therefore we use the CUDA backend when HIP is selected. This is technically
-  # equivalent to specifying -DGPU_DEVICE=CUDA.
-  set(GPU_DEVICE_CODE ${GPU_DEVICE_HIP})
-elseif("${GPU_DEVICE}" STREQUAL "XPU")
-  set(GPU_DEVICE_CODE ${GPU_DEVICE_XPU})
-elseif("${GPU_DEVICE}" STREQUAL "MPS")
-  set(GPU_DEVICE_CODE ${GPU_DEVICE_MPS})
+
+if(NOT "${GPU_DEVICE}" STREQUAL "NONE")
```
jatkinson1000: I'd make this:
```suggestion
if("${GPU_DEVICE}" STREQUAL "NONE")
  message(STATUS "CPU-only build")
  set(GPU_DEVICE_CODE ${GPU_DEVICE_NONE})
elseif("${GPU_DEVICE}" STREQUAL "CUDA")
  message(STATUS "Building with CUDA support")
  ...
```

which reads clearer IMO and avoids an unnecessary  negation.

README.md
```diff
@@ -173,6 +173,7 @@ To build and install the library:
...
     | `CMAKE_BUILD_TESTS`                                                                               | `TRUE` / `FALSE`                        | Specifies whether to compile FTorch's [test suite](https://cambridge-iccs.github.io/FTorch/page/testing.html) as part of the build.                                        |
     | `GPU_DEVICE`                                                                                      | `NONE` / `CUDA` / `HIP` / `XPU` / `MPS` | Specifies the target GPU backend architecture (if any) <sup>3</sup>                                                                                                        |
+    | `MULTI_GPU`                                                                                      | `ON` / `OFF` |  Specifies whether to build the tests that involve multiple GPU devices (`ON` by default).                                                                                                        |
```
jatkinson1000:
```suggestion
    | `MULTI_GPU`                                                                                      | `ON` / `OFF` |  Specifies whether to build the tests that involve multiple GPU devices (`ON` by default if `CMAKE_BUILD_TESTS` and `GPU_DEVICE` are set.).                                                                                                        |
```

pages/testing.md
```diff
@@ -15,6 +15,7 @@ is set to `TRUE` for the build,  i.e., `-DCMAKE_BUILD_TESTS=True`. If you want
 to run the unit tests, you will also need to install
 [pFUnit](https://github.com/Goddard-Fortran-Ecosystem/pFUnit) and pass its
 install directory to the `CMAKE_PREFIX_PATH` when building FTorch.
+If a `GPU_DEVICE` is specified but only one is available, i.e. when in a single GPU setting, you will need to also set the `MULTI_GPU` option to `OFF`.
```
jatkinson1000, line 18: Explain why:
```suggestion
If a `GPU_DEVICE` is specified but only one is available, i.e. when in a single GPU setting, you will need to also set the `MULTI_GPU` option to `OFF`.
This will skip the 'multiple GPU' integration test which requires multiple devices to be available.
```

jatkinson1000 (changes requested):
Hi @niccolozanotti 

Subject to my suggestions below I'm happy with this, thanks for the great work getting this long-standing task completed!

Let me know when you've taken a look and I'll approve for merging.

.gitignore
niccolozanotti: My thoughts exactly

CMakeLists.txt
```diff
@@ -36,22 +36,28 @@ option(GPU_DEVICE
...
-  # > This helps to accelerate the porting of existing PyTorch code and models
-  # > because very few code changes are necessary, if any."
-  # Therefore we use the CUDA backend when HIP is selected. This is technically
-  # equivalent to specifying -DGPU_DEVICE=CUDA.
+  # (https://docs.pytorch.org/docs/stable/notes/hip.html) > "PyTorch for HIP
+  # intentionally reuses the existing torch.cuda interfaces. > This helps to
+  # accelerate the porting of existing PyTorch code and models > because very
+  # few code changes are necessary, if any." Therefore we use the CUDA backend
+  # when HIP is selected. This is technically equivalent to specifying
+  # -DGPU_DEVICE=CUDA.
```
jatkinson1000: Something funky happened to the formatting here:

```suggestion
  # (https://docs.pytorch.org/docs/stable/notes/hip.html) > "PyTorch for HIP
  # intentionally reuses the existing torch.cuda interfaces.
  # > This helps to accelerate the porting of existing PyTorch code and models
  # > because very few code changes are necessary, if any." 
  # Therefore we use the CUDA backend when HIP is selected. This is technically
  # equivalent to specifying -DGPU_DEVICE=CUDA.
```

CMakeLists.txt
```diff
@@ -170,6 +176,14 @@ if(CMAKE_BUILD_TESTS)
             "No Python virtual environment detected. Please activate one.")
   endif()
 
+  if(NOT "${GPU_DEVICE}" STREQUAL "NONE")
+    # If a GPU device is passed, specify if more than one is available. Set to
+    # OFF if only one is available. Used to determine whether to run Multi GPU
+    # testing
+    option(MULTI_GPU "Whether to enable Multi GPU testing (ON [default], OFF)"
+           OFF)
```
jatkinson1000: Should this not be set to `ON` as the default?
```suggestion
           ON)
```

niccolozanotti:
@jatkinson1000 your suggestions are in :)

What about this?
> > It might be possible to disable CPU tests with some clever logic, but at the moment I'm not sure it's worth pursuing this
> > Yeah, requires some thought.
> 
> One thing we could do right off the bat would be to just run the `f2018` standard in the CPU testing for the cases in which the GPU workflow is run. The GPU run just adds the testing pool so it should catch any issue in the `f2008` anyway.

examples/CMakeLists.txt
```diff
@@ -7,7 +7,8 @@ if(CMAKE_BUILD_TESTS)
   # NOTE: Example 5_Looping is built here but not run as part of the integration
   #       test suite. This is because it demonstrates 'good' versus 'bad'
   #       practice, as opposed to functionality.
-  if(NOT "${GPU_DEVICE}" STREQUAL "NONE")
+  if(NOT "${GPU_DEVICE}" STREQUAL "NONE" AND MULTI_GPU)
+    message(STATUS "Build Multi GPU integration tests")
     add_subdirectory(6_MultiGPU)
   endif()
```
jatkinson1000: General practice is to warn when tests are skipped rather than when they are included:
```suggestion
  if(NOT "${GPU_DEVICE}" STREQUAL "NONE" AND MULTI_GPU)
    add_subdirectory(6_MultiGPU)
  else()
      message(STATUS "Skipping Multi GPU integration tests")
  endif()
```

jatkinson1000 (changes requested):
We're on the edge of glory @niccolozanotti 

Some tiny queries/requests before this can go in.

jatkinson1000:
> What about this?
> 
> > > It might be possible to disable CPU tests with some clever logic, but at the moment I'm not sure it's worth pursuing this
> > > Yeah, requires some thought.

Nah, it's cheap enough that I think we can ignore this.

CMakeLists.txt
jatkinson1000: You'll then need to amend the workflow to pass `-DMULTI_GPU=OFF`

niccolozanotti: Addressed in [7c387bb](https://github.com/Cambridge-ICCS/FTorch/pull/410/commits/7c387bb875149e4321eb45352b9631ee149d127d) + [8e19465](https://github.com/Cambridge-ICCS/FTorch/pull/410/commits/8e194650aa998d05ab823779ff859a699390874e)

examples/CMakeLists.txt
niccolozanotti: Addressed in [e7f1bf8](https://github.com/Cambridge-ICCS/FTorch/pull/410/commits/e7f1bf84b4e6e6a936fbda81200fb398c1de57ef)

pages/cmake.md
```diff
@@ -53,18 +53,17 @@ cmake --install .
...
+| [`CMAKE_PREFIX_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html) | `</path/to/libTorch/>` | Location of Torch installation<sup>2</sup> |
+| [`CMAKE_INSTALL_PREFIX`](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html) | `</path/to/install/lib/at/>` | Location at which the library files should be installed. By default this is `/usr/local` |
+| [`CMAKE_BUILD_TYPE`](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) | `Release` / `Debug` | Specifies build type. The default is `Debug`, use `Release` for production code |
+| `CMAKE_BUILD_TESTS` | `TRUE` / `FALSE` | Specifies whether to compile FTorch's [test suite](testing.html) as part of the build. |
+| `GPU_DEVICE` | `NONE` / `CUDA` / `HIP` / `XPU` / `MPS` | Specifies the target GPU architecture (if any)<sup>3</sup> |
+| `MULTI_GPU` | `ON` / `OFF` | Specifies whether to build the tests that involve multiple GPU devices (`ON` by default if `CMAKE_BUILD_TESTS` and `GPU_DEVICE` are set). |
```
jatkinson1000: Again, something funky has gone on with formatting here.
Perhaps an auto-linting in your editor?

Please revert.

CMakeLists.txt
niccolozanotti: Ah yes it's my cmake LSP not picking up the formatting options specific to this project 🫠 . Fixed in [3066847](https://github.com/Cambridge-ICCS/FTorch/pull/410/commits/3066847b4adafeb6fcbe32cec3e4ac332076c129)

pages/cmake.md
niccolozanotti: You guessed right. My LSP is not picking up the right in-project config and its giving me headaches!

niccolozanotti: Fixed in [73fe688](https://github.com/Cambridge-ICCS/FTorch/pull/410/commits/73fe688cbd6c2bcbcccbcfaf818b987cc7ffc007)

.github/workflows/test_suite_ubuntu_cuda.yml
```diff
@@ -0,0 +1,125 @@
...
+      # Unit tests
+      - 'test/unit/*.pf'
+      # Integration tests
+      - 'examples/**/*.py'
+      - 'examples/**/*.f90'
+      # Build system
+      - '**CMakeLists.txt'
+      - '**requirements.txt'
+
+  # TODO: remove, just here for testing!!
```
niccolozanotti: When this is good to go I'll get rid of this

jatkinson1000 (approved):
Happy with this now @niccolozanotti 😁 

Please remove the TODO and then I'm happy for you to merge.

Based on commit number I suggest squashing as the line changes are pretty concise and isolated.

Then please have a nap/walk!!

PR #411, opened 2025-07-22
Fix cmake bug in building CUDA unit tests

niccolozanotti:
A small bug in the cmake introduced after 56e30c3.

I came across this while working on the CI for CUDA testing #410, which confirms the need for having GPU-aware CI :)


jatkinson1000 (approved):
Brilliant, thanks for catching this @niccolozanotti 
Happy for this to be merged, or for you to add it as part of #410 

Further evidence of the need for #410!

PR #413, opened 2025-07-23
Add cmake optional variable `MULTI_GPU` to switch `ON`/`OFF` 

niccolozanotti:
The need for this was prompted by #410 since the runner used to do the GPU testing offers just 1 GPU.

**Note**: This would change the build behavior as building MultiGPU integration tests now explicitely requires setting `MULTI_GPU=ON`
- [ ] Consider whether to change default behavior, i.e. OFF by default
- [ ] Add changes made by this PR to the docs

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,17 @@ else()
   message(SEND_ERROR "GPU_DEVICE '${GPU_DEVICE}' not recognised")
 endif()
 
+# If a GPU device is passed, specify if more than one is available This is used
+# to determine whether to run Multi GPU testing
+if(NOT "${GPU_DEVICE}" STREQUAL "NONE")
+  option(MULTI_GPU "Whether to enable Multi GPU testing (OFF [default], ON)"
+         OFF)
+else()
+  if(MULTI_GPU)
+    message(SEND_ERROR "Multi GPU turned ON but no GPU device specified")
+  endif()
+endif()
+
 # Other GPU specific setup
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -170,6 +181,10 @@ if(CMAKE_BUILD_TESTS)
             "No Python virtual environment detected. Please activate one.")
   endif()
 
+  if(MULTI_GPU)
+    message(STATUS "Building tests with Multi GPU support")
+  endif()
+
```
jatkinson1000, line 187: I wonder if a "Building with GPU support" would also be useful feedback to the user?

CMakeLists.txt
```diff
@@ -57,6 +57,17 @@ else()
...
+# If a GPU device is passed, specify if more than one is available This is used
+# to determine whether to run Multi GPU testing
+if(NOT "${GPU_DEVICE}" STREQUAL "NONE")
+  option(MULTI_GPU "Whether to enable Multi GPU testing (OFF [default], ON)"
+         OFF)
+else()
+  if(MULTI_GPU)
+    message(SEND_ERROR "Multi GPU turned ON but no GPU device specified")
+  endif()
+endif()
```
jatkinson1000, line 69: I think I would be inclined to have it set ON by default.
This would match the current behaviour retaining backwards compatibility.

In use the reason to specify this option is to turn multiGPU OFF for testing, so I think it should be a positive action to turn it off, default to on.

Issue #412, opened 2025-07-22
Confused about transposes / strides


peterukk:
Hi,

After reading ["When to transpose data" ](https://cambridge-iccs.github.io/FTorch/page/transposing.html) I'm afraid I'm still a bit confused. It seems the "normal" way to use FTorch is to have your Fortran and PyTorch arrays be similar, and have FTorch get around the column/row major differences of the languages by working out the appropriate strides in torch_tensor_from_array. But wouldn't this imply a performance penalty or a transpose somewhere else? For example, SIMD vectorization is generally much faster with unit strides. I imagine the PyTorch model wants to work with contiguous arrays for various optimizations e.g. vectorization. If the Fortran and PyTorch code are set up with similar arrays, there must be a transpose or non-unit-stride access happening **somewhere**. 

I've previously used the PyTorch-Fortran library and with that I constructed my Fortran input arrays with the transposed layout w.r.t. PyTorch, for instance my_fortran_array(nx, nbatch) where nx is the contiguous dimension in memory and my understanding is that the library takes a pointer and will simply re-interpret the array as pytorch_array(nbatch, nx) where again nx is the contiguous dimension in physical memory, as it should be.  How can I replicate this behaviour with FTorch and avoid any under-the-hood transposes / non contiguous memory access?. Do I do what I did before but pass layout=[2,1] to torch_tensor_from_array?  A bit brain-dead today, sorry if I missed something obvious!

peterukk:
Thanks Jack that was very helpful! I hope to benchmark the performance difference between the two approaches. It could be that the non-sequential reading of the input arrays, and non-sequential writing of output at the end, is very rarely a bottleneck in practice as [modern ML libraries such as PyTorch are liberal with making new memory allocations, assuming cost is dominated by matrix-matrix multiplies (a good assumption for anything but truly small neural networks)](https://julialang.org/blog/2022/04/simple-chains/).  So I guess there would be only one non-sequential read at the beginning of the model, but intermediate arrays allocated by PyTorch would be contiguous in memory.

peterukk:
I'm trying to understand the stride calculation in [torch_tensor_from_blob](https://cambridge-iccs.github.io/FTorch/proc/torch_tensor_from_blob.html). Let's say nx=16, nbatch=100 using the previous example where nx should be memory-contiguous dimension, so tensor_shape = [16,100] for the Fortran array. 
layout = [1,2] --> strides =  1   16 
[...]

PR #414, opened 2025-07-25
[WIP] Permute shape in torch_tensor_from_blob

jatkinson1000:
Exploring the functionality of layout following query raised in #412 

Aim is to generate row-major tensors correctly with layout `[n, n-1, ..., 2, 1]`

TODO:

- [ ] Make proposed change and check tests pass
- [ ] Write tests to check for correct row-major tensors in Torch generated from Fortran transpose with `layout = [2, 1]`
- [ ] Add explanation here of working to explore this issue
- [ ] Check that this correctly performs what is described in the [docs](https://cambridge-iccs.github.io/FTorch/page/transposing.html) (section 3).
- [ ] Improve documentation with clear examples
- [ ] Add changelog
- [ ] Make minor version release


jatkinson1000:
Take the data in memory as:
```
[ 1.0, 2.0, 3.0, 4.0, 5.0, 6.0 ]
```
This comes from a (column-major) Fortran array with shape `[2, 3]`:
```
[ 1.0, 3.0, 5.0, 
  2.0, 4.0, 6.0 ]
```
---
Calling `torch_tensor_to_array()` with `layout = [1, 2]` leaves this alone in memory to create an identical Torch tensor of shape `[2, 3]` accessed in strides `[1, 2]`:
```
[ 1.0, 3.0, 5.0, 
  2.0, 4.0, 6.0 ]
```
---
Calling `torch_tensor_to_array()` with `layout = [2, 1]` leaves this alone in memory to create a Torch tensor preserving shape `[2, 3]` accessed in strides `[3, 1]`:
```
[ 1.0, 2.0, 3.0, 
  4.0, 5.0, 6.0 ]
```
This is 'row-major' access to the data, but preserving the shape.
---
A more realistic scenario is that users want to obtain 'row-major' access to the data for efficiency in Torch, but doing so with a 'transpose' of the original Fortran data for the purposes of maintaining the 'scientific meaning' of the data.

This can be achieved with the new `permute_shape = .true.` option.\
Calling `torch_tensor_to_array()` with `layout = [2, 1]` and `permute_shape = .true.` leaves the data alone in memory to create a Torch tensor, but permutes the shape to be `[3, 2]` then giving strided access in `[2, 1]`:
```
[ 1.0, 2.0, 
  3.0, 4.0,
  5.0, 6.0 ]
```
i.e. the transpose of the Fortran array accessed in a 'row-major' way.
---

Suppose a user wants to access the Fortran array using the same shape and indexing from Torch, but with row-major layout in memory for internal (to Torch) efficiency.
They can do this by transposing the array in Fortran:
```
[ 1.0, 2.0, 
  3.0, 4.0,
  5.0, 6.0 ]
```
to appear in Memory as:
```
[ 1.0, 3.0, 5.0, 2.0, 4.0, 6.0 ]
```
and then calling `torch_tensor_to_array()` with `layout = [2, 1]` and `permute_shape = .true.` to leave the data alone in memory, but permute shape to `[2, 3]` with strided access of `[3, 1]`:
```
[ 1.0, 3.0, 5.0, 
  2.0, 4.0, 6.0 ]
```
Exactly as would be expected had we declared this tensor from first principles in Torch.

PR #416, opened 2025-07-30
Expose access to tensor stride

Mikolaj-A-Kowalski:
I believe it will be useful feature to be able to access the strides of a tensor from Fortran. I feel like it should help with debugging/understanding what the `_from_array` tensor constructors are doing.  

In addition I have removed the `UNIX` preprocessor variable and associated `ifdef-else`. I *think* it is possible to avoid them by using appropriate typedefs. I separated it into its own commit so it would be easier to drop if I am missing something here ;-)  It is certainly something to discuss if we want integrated.  

I have marked it as draft until we decide if we want to integrate these features.

### TODO
- [x] Update  `CHANGELOG.md` (?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ For specific details see the [FTorch online documentation](https://cambridge-icc
 
 ### Added
 
+- Expose tensor strides via `get_stride` method [#416](https://github.com/Cambridge-ICCS/FTorch/pull/416)
+- Remove `UNIX` preprocessor variable that selected the  right C-integer type for 64bit int. Use `int64_t` instead [#416](https://github.com/Cambridge-ICCS/FTorch/pull/416)
 - A new cmake option `MULTI_GPU` to control the build of multi GPU integration tests in [#410](https://github.com/Cambridge-ICCS/FTorch/pull/410)
 - Support for AMD GPU backends (HIP) provided in
   [#385](https://github.com/Cambridge-ICCS/FTorch/pull/385) and


test/unit/unittest_tensor_interrogation.pf
```diff
@@ -156,6 +156,54 @@ contains
...
+    integer(kind=c_int64_t), parameter :: stride(ndims) = [1]
+
+    ! Create a tensor with uninitialised values and check torch_tensor_get_shape can correctly
+    ! identify its shape
+    call torch_tensor_empty(tensor, ndims, tensor_shape, dtype, device_type)
+    @assertEqual(stride, tensor%get_stride())
+
+  end subroutine test_get_stride_1D
+
+  ! Unit test for the torch_tensor_get_shape method of a 1D tensor
```
jatkinson1000: Copy-paste typo
```suggestion
  ! Unit test for the torch_tensor_get_shape method of a 2D tensor
```

test/unit/unittest_tensor_interrogation.pf
```diff
@@ -156,6 +156,54 @@ contains
...
+    integer(kind=c_int64_t), parameter :: stride(ndims) = [3, 1]
+
+    ! Create a tensor with uninitialised values and check torch_tensor_get_shape can correctly
+    ! identify its shape
+    call torch_tensor_empty(tensor, ndims, tensor_shape, dtype, device_type)
+    @assertEqual(stride, tensor%get_stride())
+
+  end subroutine test_get_stride_2D
+
+  ! Unit test for the torch_tensor_get_shape method of a 1D tensor
```
jatkinson1000: Copy-paste typo
```suggestion
  ! Unit test for the torch_tensor_get_shape method of a 3D tensor
```

test/unit/unittest_tensor_interrogation.pf
jatkinson1000, line 1: As a bit of a nit I'd prefer `stride` to be `expected_stride` just be be very explicit (and match most other tests).

I'd also suggest checking the strides of a tensor generated with inverse layout (`[2, 1]`) to ensure this is picked up correctly.

src/ctorch.h
```diff
@@ -25,6 +25,14 @@ typedef void *torch_int_t;
 // Opaque pointer type alias for float scalars
 typedef void *torch_float_t;
 
+// Type that represents size of the tensors
+//
+// Torch is using internally int64_t (30/07/2025)
+// We can as well to avoid portability problems
+// (i.e. 64bit integer beeing long long on Windows)
+//
+typedef int64_t torch_size_t;
```
jatkinson1000, line 35: Not sure if it needs a change, but something to consider - could this type be renamed now that it is being used for more than just sizes?

jatkinson1000 (commented):
Thanks @Mikolaj-A-Kowalski can confirm it works on Mac!
It's great to remove some jank, so thanks for updating this!
How did you find that they were using `int64_t`?

Would be good to test on Windows, but that can become part of reinstating the Windows CI I think.

Couple of comments/queries added inline.
Could also consider adding to the "tensor interrogation" example - I also want to make an example around these aspects as part of #415 where this will be very useful.

The other thing is I think you'll need to update `FTorch.md` to remove the `UNIX` macro.

test/unit/unittest_tensor_interrogation.pf
Mikolaj-A-Kowalski: :sweat_smile: Thanks!

test/unit/unittest_tensor_interrogation.pf
Mikolaj-A-Kowalski: This is a fair point! 100% agree. Will rename `stride` -> `expected_stride`. 

As for the check on the stride upon construction I will maybe add them to the `constructors_destructors` suite?

Just have a feeling that in the new tests here, we just want to test that the binding works. 
But as you pointed out it is a good idea to use the new binding to add extra checks that the construction from array works as expected.

src/ctorch.h
Mikolaj-A-Kowalski: I have went with this name by analogy with `std::size_t`, which is generally used for all indexing- and size-related computations for the  standard containers. So I think that the name is ok in that sense. 

But I think the comment may be misleading. Will change to make the above more explicit.

Mikolaj-A-Kowalski:
> How did you find that they were using int64_t?

I didn't go too deep, but just looked at the return types of the [`stride`](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/core/TensorBase.h#L181) (and other) methods for the A10 BaseTensor. For all the size/stride-related functionality they use  `int64_t` or the `c10::SymInt` which claims to have ["API equivalent to int64_t"](https://github.com/pytorch/pytorch/blob/main/c10/core/SymInt.h#L21).  

> Couple of comments/queries added inline.
Could also consider adding to the "tensor interrogation" example - I also want to make an example around these aspects as part of https://github.com/Cambridge-ICCS/FTorch/pull/415 where this will be very useful.

> The other thing is I think you'll need to update FTorch.md to remove the UNIX macro.

:ok_hand:

test/unit/unittest_tensor_interrogation.pf
Mikolaj-A-Kowalski: Extra checks in the 'constructor' tests have been added.

Mikolaj-A-Kowalski:
@jatkinson1000 I think I have addressed all of your comments and I think this one should be ready for more through review and to be integrated. 

When adding the extra example I have decided to go with minimal modification and just modify `Example_1` with few lines to spit out shape and strides of the tensor. I felt that this might be the most appropriate place for that. The alternative I was thinking of was to create a 2nd executable in the `Example_1` which would focus on interrogations of the tensor.  This is what we might do if the number of things to showcase grows (e.g. as a result of #415).

examples/1_Tensor/tensor_manipulation.f90
jatkinson1000, line 1: Perhaps a note in the comments to say what the 'expected' result of these should be would be useful?

src/ctorch.h
```diff
@@ -25,6 +25,15 @@ typedef void *torch_int_t;
 // Opaque pointer type alias for float scalars
 typedef void *torch_float_t;
 
+// Type that represent size, strides and indexing on tensors
```
jatkinson1000:
```suggestion
// Type that represents size, strides and indexing on tensors
```

src/ctorch.h
```diff
@@ -25,6 +25,15 @@ typedef void *torch_int_t;
...
+// (like std::size_t for standard containers)
+//
+// Torch is using internally int64_t [i.e. signed 64bit integer] (30/07/2025)
```
jatkinson1000: What does this date signify?

test/unit/unittest_tensor_interrogation.pf
```diff
@@ -156,6 +156,54 @@ contains
...
+  @test(testparameters={get_parameters_short()})
+  subroutine test_get_stride_1D(this)
+    class(TestCaseType), intent(inout) :: this
+    type(torch_tensor) :: tensor
+    integer, parameter :: ndims = 1
+    integer(kind=c_int64_t), parameter :: tensor_shape(ndims) = [6]
+    integer(kind=c_int64_t), parameter :: expected_stride(ndims) = [1]
+
+    ! Create a tensor with uninitialised values and check torch_tensor_get_shape can correctly
+    ! identify its shape
```
jatkinson1000:
```suggestion
    ! Create a tensor with uninitialised values and check torch_tensor_get_stride can correctly
    ! identify its stride pattern
```

test/unit/unittest_tensor_interrogation.pf
```diff
@@ -156,6 +156,54 @@ contains
...
+  @test(testparameters={get_parameters_short()})
+  subroutine test_get_stride_2D(this)
+    class(TestCaseType), intent(inout) :: this
+    type(torch_tensor) :: tensor
+    integer, parameter :: ndims = 2
+    integer(kind=c_int64_t), parameter :: tensor_shape(ndims) = [2,3]
+    integer(kind=c_int64_t), parameter :: expected_stride(ndims) = [3, 1]
+
+    ! Create a tensor with uninitialised values and check torch_tensor_get_shape can correctly
+    ! identify its shape
```
jatkinson1000:
```suggestion
    ! Create a tensor with uninitialised values and check torch_tensor_get_stride can correctly
    ! identify its stride pattern
```

test/unit/unittest_tensor_interrogation.pf
```diff
@@ -156,6 +156,54 @@ contains
...
+  @test(testparameters={get_parameters_short()})
+  subroutine test_get_stride_3D(this)
+    class(TestCaseType), intent(inout) :: this
+    type(torch_tensor) :: tensor
+    integer, parameter :: ndims = 3
+    integer(kind=c_int64_t), parameter :: tensor_shape(ndims) = [1,2,3]
+    integer(kind=c_int64_t), parameter :: expected_stride(ndims) = [6, 3, 1]
+
+    ! Create a tensor with uninitialised values and check torch_tensor_get_shape can correctly
+    ! identify its shape
```
jatkinson1000:
```suggestion
    ! Create a tensor with uninitialised values and check torch_tensor_get_stride can correctly
    ! identify its stride pattern
```

test/unit/unittest_tensor_interrogation.pf
```diff
@@ -156,6 +156,54 @@ contains
...
+    integer(kind=c_int64_t), parameter :: expected_stride(ndims) = [3, 1]
+
+    ! Create a tensor with uninitialised values and check torch_tensor_get_shape can correctly
+    ! identify its shape
+    call torch_tensor_empty(tensor, ndims, tensor_shape, dtype, device_type)
+    @assertEqual(expected_stride, tensor%get_stride())
+
+  end subroutine test_get_stride_2D
+
+  ! Unit test for the torch_tensor_get_shape method of a 3D tensor
```
jatkinson1000:
```suggestion
  ! Unit test for the torch_tensor_get_stride method of a 3D tensor
```

test/unit/unittest_tensor_interrogation.pf
```diff
@@ -156,6 +156,54 @@ contains
...
+    integer(kind=c_int64_t), parameter :: expected_stride(ndims) = [1]
+
+    ! Create a tensor with uninitialised values and check torch_tensor_get_shape can correctly
+    ! identify its shape
+    call torch_tensor_empty(tensor, ndims, tensor_shape, dtype, device_type)
+    @assertEqual(expected_stride, tensor%get_stride())
+
+  end subroutine test_get_stride_1D
+
+  ! Unit test for the torch_tensor_get_shape method of a 2D tensor
```
jatkinson1000:
```suggestion
  ! Unit test for the torch_tensor_get_stride method of a 2D tensor
```

test/unit/unittest_tensor_interrogation.pf
```diff
@@ -156,6 +156,54 @@ contains
 
   end subroutine test_get_shape_3D
 
+  ! Unit test for the torch_tensor_get_shape method of a 1D tensor
```
jatkinson1000:
```suggestion
  ! Unit test for the torch_tensor_get_stride method of a 1D tensor
```

test/unit/unittest_tensor_constructors_destructors.pf
jatkinson1000, line 1: I don't object, but just wondered your reasoning for adding a stride and shape check here?

jatkinson1000 (changes requested):
Hi @Mikolaj-A-Kowalski Thanks for this.

I'm broadly happy with this PR now, just a couple of queries/suggestions.

Also a bunch of sneaky copy-paste typos that I'm happy for you to collapse into a single commit.

test/unit/unittest_tensor_constructors_destructors.pf
Mikolaj-A-Kowalski: Basically I though that, now since we can, we can make sure that the strides for the `from_array` constructors are as we expect them to be. Extra benefit is that if ones looks into test for learn more about the interface, they will know what strides the `layout` will produce. 

For the rest of the constructors I just added the check for consistency.

src/ctorch.h
Mikolaj-A-Kowalski: Wanted to make a record when I checked in case it will stop being valid. 

But one can just `git blame` if one needs. Will remove.

examples/1_Tensor/tensor_manipulation.f90
Mikolaj-A-Kowalski: I have added a trailing comment to indicate that (for the `shape` as well for consistency).

Mikolaj-A-Kowalski:
I think that all the comments should be addressed.

Sorry for the delay.

examples/1_Tensor/tensor_manipulation.f90
jatkinson1000: Thanks

PR #418, opened 2025-08-04
Use pointers in array tensor constructor functions

### LINKED ISSUE #417: Protect `torch_tensor_from_array` from abuse
I think there are two potentially dangerous things that can be done with the `from_array` tensor constructors. Perhaps we should try to think about some protections or at least we should probably discuss these failure modes more explicitly in the documentation?

### Passing non-contiguous arrays is allowed 

Currently there is nothing that protects against doing the following (e.g. in the SimpleNet example)
```Fortran
   ...
   real(wp), dimension(10), target :: in_data
   real(wp), dimension(10), target :: out_data
   real(wp), dimension(10), target :: expected
   ...

   call torch_tensor_from_array(in_tensors(1), in_data(1:10:2), torch_kCPU)
   call torch_tensor_from_array(out_tensors(1), out_data(1:10:2), torch_kCPU)
   ...
```
which (unless I am forgetting something) is valid Fortran and something a user may do at some point. The problem is that the result is not what is expected since the `torch_tensor_from_array` implicitly assumes that the input array is contiguous. As the result in this example the `in_data(1:5)`  will be used instead of the odd indices. 

Perhaps we should add `is_contiguous` check inside `torch_tensor_from_array` and throw an error if it isn't? 

### Passing a temporary array

Since the `data_in` dummy argument is declared `intent(in)` the following is allowed: 
```Fortran
   call torch_tensor_from_array(in_tensors(1), in_data*2, torch_kCPU)
``` 
as in we can provide result of an expression as an argument, which I think will make a tensor point to a temporary. This problem is technically mentioned in the docs (temporary array has no `target` attribute which the docs say is required), but I feel like this is easy to miss (e.g. some of the usage like that sneaked into the unit tests) and we should think of some automatic  protection against it. 

I think there are two legal ways of doing that: 
1. (preferred) Make the `data_in` an `intent(in)` pointer. I believe in that case a `target` actual argument should [still bind](https://www.ibm.com/docs/en/xffbg/121.141?topic=association-pointers-as-dummy-arguments). Also maybe adding `contiguous` attribute will help to move the check from the 1st problem from runtime to compile-time (need to check) [EDIT: Seems to work on the first glance. The only problem is that the error message is not very informative. Just: `"There is no specific subroutine..." :-/   
2. We can make the `data_in` `intent(inout)` and just be careful not to modify it by accident. It should prevent accidental usage with temporary arrays (or at least reduce a chance of that happening)

Last but not least I am not sure how it will all work for backwards compatibility :-/

Mikolaj-A-Kowalski:
Closes #417

Changing the type of the dummy argument in the `tensor_from_array` constructors to `pointer, contiguous` should generate a compiler error if:
 - one provides non-contiguous array as on actual argument
 - one provides a temporary array as an actual argument

Providing `target` array as an actual argument still works since it is allowed as an actual argument to `pointer, intent(in)` dummy argument since F2008.

Todo: 
- [ ] Improve documentation to help with unclear compiler message
- [ ] Include the runtime check for the non-contiguous arguments anyway, since compiler (should?) allow a code which selects slice of the array dynamically since it cannot know if it will or not be contiguous (?) [EDIT: Actually it seem it does not allow a code with vector indexing [like that](https://godbolt.org/z/K3hrn9c59)]
- [ ] Update `Changelog.md` (?)

### DIFF
--- a/examples/8_Autograd/autograd.f90
+++ b/examples/8_Autograd/autograd.f90
@@ -18,7 +18,9 @@ program autograd
 
   integer, parameter :: ndims = 1
   integer, parameter :: n = 2
+  real(wp), dimension(n), target :: in_data1, in_data2
   real(wp), dimension(n), target :: out_data1, out_data2, out_data3
+  real(wp), dimension(1), target :: multiplier_value, divisor_value
   real(wp), dimension(n) :: expected
 
   ! Flag for testing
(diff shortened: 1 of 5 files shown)


joewallwork:
This all seems like a good idea. Tbh I wasn't actually aware of some of these issues. We've targeted it for the v1.1 release.

PR #422, opened 2025-08-26
fix a minor bug

zhenkunl:
fix: prevent gradient calculation in inference mode

Add torch.no_grad() context manager to the model inference step
to avoid unnecessary computation and memory usage.

### DIFF
--- a/examples/3_ResNet/resnet_infer_python.py
+++ b/examples/3_ResNet/resnet_infer_python.py
@@ -96,6 +96,7 @@ if __name__ == "__main__":
 
     batch_size_to_run = 1
 
-    result = deploy(saved_model_file, device_to_run, batch_size_to_run)
+    with torch.no_grad():
+        result = deploy(saved_model_file, device_to_run, batch_size_to_run)
     print_top_results(result)
     check_results(result)


examples/3_ResNet/resnet_infer_python.py
```diff
@@ -96,6 +96,7 @@ def check_results(output: torch.Tensor) -> None:
 
     batch_size_to_run = 1
 
-    result = deploy(saved_model_file, device_to_run, batch_size_to_run)
+    with torch.no_grad():
```
jatkinson1000, line 99: I'd suggest using `inference mode` as preferable to `no_grad`

```suggestion
    with torch.inference_mode():
```

jatkinson1000 (changes requested):
Thanks for spotting this this @zhenkunl 

I have suggested changing for `inference_mode()` rather than `no_grad()`, though realise this would also need updating to match in the other exercises that currently use `no_grad()`.

If you can accept the suggestion I'm happy to merge this and update the others.
Alternatively, if you want to make the updates to the other exercises here we would be grateful and happily merge.

PR #423, opened 2025-08-26
Permute view

### LINKED ISSUE #157: Implement batching
When we run `examples/1_SimpleNet/simplenet.py`, the final thing that's executed is effectively
```
a = [0.0, 1.0, 2.0, 3.0, 4.0]
model(torch.Tensor(a))
```

This would also work with batching e.g., 
```
a = [0.0, 1.0, 2.0, 3.0, 4.0]
model(torch.Tensor([a, a]))
```

We should enable calling batching in this way on the FTorch side, too.

Mikolaj-A-Kowalski:
Add a binding for the `permute` function so one can create permuted views of a tensor. 

I believe it is a desired feature to resolve #157 since to use batching, on the Fortran side one would like to organise the data such that the batch dimension is the rightmost one (to avoid scattering data for a single datapoint(?) in a batch across memory). However, the `pytorch` assumes the leftmost dimension is the batch dimension. Hence some sort of permutation is required to make the two conventions agree. 

One could probably do the required permutation on the tensor constructor using the changes in #415, but I had a feeling that having an explicit option to make new, permuted views after construction will not hurt. 

The side effect is that I had to add another 'assignment'  function to mimic what would happen on assignment in Python (I think!) or when `libtorch` does assignment to an lvalue i.e. make shallow copy of the tensor.

Will address the todo below after we agree that we are sure we want these features ;-) 

Todo: 
- [ ] Update changelog 
- [ ] Add additional tests to assert `requires_grad` is propagated correctly
- [ ] Add user-facing (web) documentation and example in `Example 1`

### DIFF
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -376,6 +376,22 @@ void torch_tensor_to(const torch_tensor_t source_tensor, torch_tensor_t target_t
 // --- Operator overloads acting on tensors
 // =====================================================================================
 
+void torch_tensor_shallow_copy(torch_tensor_t *to, const torch_tensor_t from) {
+  auto to_tensor = reinterpret_cast<torch::Tensor *>(*to);
+  auto from_tensor = reinterpret_cast<torch::Tensor *>(from);
+
+  validate_tensor(from_tensor, "from");
+  if (!to_tensor) {
+    to_tensor = new torch::Tensor();
+  }
+
+  // Copy Tensors
+  *to_tensor = *from_tensor;
+
+  // Update lhs pointer value
+  *to = to_tensor;
+}
+
 void torch_tensor_assign(torch_tensor_t output, const torch_tensor_t input) {
   auto out = reinterpret_cast<torch::Tensor *>(output);
   auto in = reinterpret_cast<torch::Tensor *const>(input);
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -8,6 +8,8 @@
 
 #include "ctorch.h"
 
+#include <sstream>
+
 #ifndef GPU_DEVICE
 #define GPU_DEVICE GPU_DEVICE_NONE
 #endif
@@ -30,6 +32,19 @@ void ctorch_error(const std::string &message,
 void ctorch_warn(const std::string &message) {
   std::cerr << "[WARNING]: " << message << std::endl;
 }
+
+void assert_not_nullptr(const void *ptr, const char *variable_name) {
+  if (!ptr) {
+    std::stringstream ss;
+    ss << "Pointer " << variable_name << "cannot be null\n";
+    ctorch_error(ss.str());
+  }
+}
+
+/**
+ * Raise `ctorch_error` if the expression evaluates to a NULL ptr
+ */
+#define CTORCH_ASSERT_NOT_NULLPTR(ptr) assert_not_nullptr(ptr, #ptr)
 
 // =============================================================================
 // --- Constant expressions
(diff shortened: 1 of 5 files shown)


src/ftorch.F90
```diff
@@ -2482,10 +2482,92 @@ end subroutine torch_tensor_to_c
...
   ! ============================================================================
   ! --- Overloaded operators acting on tensors
   ! ============================================================================
 
+  !> Makes a shallow copy of a tensor
+  !>
+  !> Mimics more Python-like assigment behaviour where `b = a` makes `b` point
+  !> to the same underlying data as `a`.
+  subroutine torch_tensor_shallow_copy(to, from)
+    type(torch_tensor), intent(out) :: to !! target
```
Mikolaj-A-Kowalski, line 2555:
```suggestion
    type(torch_tensor), intent(inout) :: to !! target
```
It requires some extra thinking about edge cases but it should be all right to reuse the allocated tensor on the lhs... I think

src/ctorch.cpp
```diff
@@ -372,10 +387,50 @@ void torch_tensor_to(const torch_tensor_t source_tensor, torch_tensor_t target_t
...
+  CTORCH_ASSERT_NOT_NULLPTR(dims);
+
+  // `dim()` method of the tensor returns a signed int64, but ArrayRef requires
+  // unsigned std::size_t. Since dimensions of a tensor are never negative the cast
+  // is safe... unless we are on a system where 64bit is not used for size_t.
+  static_assert(sizeof(int64_t) == sizeof(std::size_t),
+                "CTorch has an assumption that `size_t` is 64bit to safely cast "
+                "between `int64` and `size_t`. This should be the case for most modern "
+                "64bit systems. If you are running into this problem when "
+                "compiling FTorch please open an Issue so we can fix this!");
```
jatkinson1000, line 407: Thank you for adding verbose error handling.

src/ctorch.cpp
```diff
@@ -372,10 +387,50 @@ void torch_tensor_to(const torch_tensor_t source_tensor, torch_tensor_t target_t
...
 
+// =============================================================================
+// --- Tensor view functions
+// =============================================================================
+
+torch_tensor_t torch_tensor_permute(const torch_tensor_t self, const int64_t dims[]) {
+  auto t = reinterpret_cast<const torch::Tensor *>(self);
+  validate_tensor(t, "self");
+
+  CTORCH_ASSERT_NOT_NULLPTR(dims);
```
jatkinson1000, line 398: Side note, I wonder if there are any other error messages that could be cleaned up with macros.

src/ftorch.fypp
```diff
@@ -745,10 +745,92 @@ contains
...
+
+    out%p = torch_tensor_permute_c(tensor%p, dims_c)
+
+  end function torch_tensor_permute
+
   ! ============================================================================
   ! --- Overloaded operators acting on tensors
   ! ============================================================================
 
+  !> Makes a shallow copy of a tensor
```
jatkinson1000, line 813: This makes sense to me as an operator overload if it was `=`, but not sure if it should be here as not actually 'overloading' anything afaict.
Will tag @jwallwork23 for his thoughts as more in touch with this code and I believe tried something similar previously.

I think I'd also like to see usage of this added to example 1.

jatkinson1000 (commented):
Hi @Mikolaj-A-Kowalski 

I'm generally happy with including and wrapping these functions.
Left some comments. Please add any necessary documentation/docstrings, usage to example 1, and any other 'paperwork' and then re-request a review.

src/ftorch.fypp
Mikolaj-A-Kowalski: If I recall correctly I put it here to be close to the `assign` since it is in a way another type of an assignment.

src/ftorch.fypp
```diff
@@ -745,10 +745,92 @@ contains
...
+  !>
+  !> Mimics more Python-like assigment behaviour where `b = a` makes `b` point
```
joewallwork, line 815:
```suggestion
  !> Mimics more Python-like assignment behaviour where `b = a` makes `b` point
```

src/ftorch.fypp
joewallwork: Yeah I'd probably put it in a separate section because it doesn't get an `interface` like the other overloaded operators. Perhaps permute and shallow copy could have a "tensor conversion" section?

joewallwork:
Pushing this back to v1.2 as not urgent.

jatkinson1000:
This became superseded by #604 to introduce permutation to tensors, so I am now closing.

PR #425, opened 2025-09-04
Update to use inference_mode in preference to no_grad

jatkinson1000:
It is preferable to use `inference_mode` as a context instead of `no_grad` where possible to disable additional overheads.
See e.g. [here](https://stackoverflow.com/questions/74191070/why-are-there-two-different-flags-to-disable-gradient-computation-in-pytorch/74197846#74197846) or [here](https://discuss.pytorch.org/t/pytorch-torch-no-grad-vs-torch-inference-mode/134099/2?u=timgianitsos)

This PR makes these changes in the examples and other related areas.

### DIFF
--- a/pages/troubleshooting.md
+++ b/pages/troubleshooting.md
@@ -148,7 +148,7 @@ aren't using the bare `use ftorch` import then you should ensure you specify
 `use ftorch, only: assignment(=)` (as well as any other module members you
 require). See the [tensor documentation](doc/page/tensor.html) for more details.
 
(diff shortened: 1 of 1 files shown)


jatkinson1000:
> Perhaps it'd be worth mentioning `inference_mode` there?

So, I did exactly the same and thought about this.
I can't remember exactly why I decided not to mention it here in the end, so will add it in now.

jwallwork23 (approved):
Thanks!

PR #429, opened 2025-09-08
Enable building with fpm

joewallwork:
Closes #77.

This PR adds fpm support.
https://fpm.fortran-lang.org/

I am currently building with
```
fpm build --cxx-flag "-cpp -std=c++17 -I${Torch_DIR}/include -I${Torch_DIR}/include/torch/csrc/api/include -L${Torch_DIR}/lib" --verbose
```

I get the following error but I think this might go away soon if I recall correctly (@TomMelt @jatkinson1000)?
```
././src/ctorch.cpp: In function ‘const long long int* torch_tensor_get_sizes(torch_tensor_t)’:
././src/ctorch.cpp:317:25: error: invalid conversion from ‘const long int*’ to ‘const long long int*’ [-fpermissive]
  317 |   return t->sizes().data();
      |          ~~~~~~~~~~~~~~~^~
      |                         |
      |                         const long int*
[ 25%]                     ctorch.cpp  done.

././src/ctorch.cpp: In function ‘const long long int* torch_tensor_get_sizes(torch_tensor_t)’:
././src/ctorch.cpp:317:25: error: invalid conversion from ‘const long int*’ to ‘const long long int*’ [-fpermissive]
  317 |   return t->sizes().data();
      |          ~~~~~~~~~~~~~~~^~
      |                         |
      |                         const long int*
<ERROR> Compilation failed for object " src_ctorch.cpp.o "
<ERROR> stopping due to failed compilation
STOP 1
```

## Checklist

- [x] Fix `long` issue
- [x] Update metadata in `fpm.toml`
- [ ] Avoid duplicated preprocessor macro settings?
- [ ] Update docs
- [x] Test `fpm` build in CI
- [ ] Add separate `fpm` CI workflow (or possibly use matrix)
- [ ] Try out `fpm test`
- [ ] Add linting for `.toml` files?

### DIFF
--- a/fpm.toml
+++ b/fpm.toml
@@ -15,7 +15,8 @@ include-dir = [
 
 [build]
 auto-executables = true
-auto-tests = true
+# TODO: pFUnit currently doesn't provide an fpm.toml
+auto-tests = false
 auto-examples = true
 module-naming = false
 


jatkinson1000:
> I get the following error but I think this might go away soon if I recall correctly (@TomMelt @jatkinson1000)?

This would be resolved as part of #416 

Though I'm surprised its showing up unless you are building on an obscure system or compiler?

jwallwork23:
Ah, testing doesn't currently work because it's all done through CMake. Will need to add a `main.F90` script for the unit tests. Not sure that the integration tests (examples) can be done through fpm given it involves Python scripts etc.

Issue #430, opened 2025-09-24
Examples will cause fatal error if user has `CONDA_PREFIX` defined


jfdev001:
If a user has a conda environment and attempts build examples in `examples/`, then a `FATAL_ERROR` will get triggered due to the `DEFINED $ENV{CONDA_PREFIX}` logic below. Is this intended behavior?

https://github.com/Cambridge-ICCS/FTorch/blob/162b6e165538ecd9649a67c8331a82ec1eb48f39/examples/2_SimpleNet/CMakeLists.txt#L21-L24

This logic appears in the following files in the `examples/`

<img width="1221" height="217" alt="Image" src="https://github.com/user-attachments/assets/2a0e5e4a-a99e-4899-9421-2a8286f2f6b2" />

I can open a PR to change the corresponding parts of the CMakeLists.txt in the `examples/` such that this error will not be thrown, but I wanted to open an issue first. The change would simply be 

```python
if(NOT DEFINED ENV{VIRTUAL_ENV} AND NOT DEFINED ENV{CONDA_PREFIX})
  message(FATAL_ERROR "Please activate your virtualenv or conda environment")
endif()

```

jatkinson1000:
Another thing that would be nice, I guess, would be to add some conda CI.
Though I accept this is a bigger task and would require a separate PR

PR #431, opened 2025-09-24
Prevent fatal error from being thrown during cmake examples

jfdev001:
Fatal error was caused by faulty logic in environment variables related to Python installation. This logic is corrected.

Resolves https://github.com/Cambridge-ICCS/FTorch/issues/430


jatkinson1000 (approved):
Thanks @jfdev001 for fixing this and the contribution to the project!

Since it's a small tweak I think we can proceed without notes in the documentation or changelog.
Any further advice you may have on conda or elsewhere would be welcomed in the project.

Issue #432, opened 2025-09-24
`find_package(FTorch)` should be `find_package(FTorch REQUIRED)`


jfdev001:
It does not make sense to have all the instances of `find_package(FTorch)` not use the `REQUIRED` keyword as an argument to `find_package`. Stopping the cmake build process early will save users time, especially given that, for example, in the `examples/**/CMakeLists.txt` the python installation occurs *after* calling `find_package(FTorch)`. This means time would be "wasted" installing the python packages and then the user would be confronted with this error during linking:

<img width="821" height="289" alt="Image" src="https://github.com/user-attachments/assets/758c2dff-533a-4e76-80ca-5544cba9b517" />

The places in which `find_package(FTorch)` occurs appear below:

<img width="502" height="323" alt="Image" src="https://github.com/user-attachments/assets/57da1680-4e34-4e09-a090-5cb336f2b4ca" />

I can open a PR to replace these with `find_package(FTorch REQUIRED)`, lmk if this makes sense!

jatkinson1000:
Hi again @jfdev001!

This has definitely come up before in #134 

We discovered that due to building tests at the same time as building FTorch the fact that FTorch did not 'exist' when CMake sought it out during configuration lead to failure. At the time we did not want to separate out the build of the library and tests, so decided not to proceed with this.

However, that was quite a while ago, and the CMake of the project has perhaps changed since then, so if you can find a way around to now use `REQUIRED` that we'd welcome a PR.

PR #434, opened 2025-09-26
Jfrazier/432 find package ftorch required

jfdev001:
Resolves https://github.com/Cambridge-ICCS/FTorch/issues/432

Perhaps it is not the most elegant change; however, in the `examples/*/CMakeLists.txt` and in the project level `CMakeLists.txt` I have wrapped the call to `find_package(FTorch)` in an if statement that requires the package if not building tests and is otherwise quiet if building tests. This ensures that unit and integration tests run as expected without throwing an error if FTorch were to be required, but users who just wish to compile one of the examples will by itself will be forced to specify `FTorch_DIR` or update the `CMAKE_PREFIX_PATH` in order to build an example. 

I considered trying to use generator expressions, but since `find_package` occurs during the configure step, this is not possible.


jfdev001 (commented):
https://github.com/Cambridge-ICCS/FTorch/pull/434/commits/5492dfa431b32600cbc0fa60acbb98baf1a6af40 addresses formatting while https://github.com/Cambridge-ICCS/FTorch/pull/434/commits/4f72e067d0b704a0a3a7043c291afee3d53e7129 addresses changelog

jatkinson1000 (changes requested):
Thanks @jfdev001 

This is nearly there, but it looks like https://github.com/Cambridge-ICCS/FTorch/pull/434/commits/5492dfa431b32600cbc0fa60acbb98baf1a6af40 has touched a lot of unrelated code muddying the commit.

Please could you rebase and make sure that that commit only amends the specific lines that you changed that were failing (indentation depth of `find_package`)

Once that's done we should be good to go.

I will also take a look at `cmake-format` since it seems to have forcibly changed a lot more than it should when applied.

jatkinson1000 (approved):
Amazing, thanks for the perseverence @jfdev001 and addressing a long-standing abandoned issue.

This all looks good to me and we're grateful for your contribution!

Issue #140, opened 2024-06-24
Add intel ifx and ifort build to CI


TomMelt:

ElliottKasoar:
You may be able to use an existing actions e.g. https://github.com/fortran-lang/setup-fortran, which seems to support ifx and ifort (see also: [discussion](https://fortran-lang.discourse.group/t/fortran-compiler-setup-for-github-actions/5671/10))

jfdev001:
# Findings

What follows are comments that apply only to the Intel MPI implementation since it implements the MPI-3.1. specification (see [Intel MPI Library (version: 2021.14): Get Started Guide for Linux OS](https://www.intel.com/content/www/us/en/docs/mpi-library/get-started-guide-linux/2021-14/overview.html)).

The module file `mpi.mod` (note: default install location with root privileges is `/opt/intel/oneapi/mpi/latest/include/mpi`) defines explicit interfaces only for a *subset* of all MPI routines. MPI routines (e.g., `MPI_Gather`, `MPI_Send`) with a _choice buffer_ (see [MPI-3.1 standard: Fortran Support Through the `mpi` Module](https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node410.htm#Node410)) do *not* have explicit interfaces because the first MPI versions used implicit interfaces. To maintain backwards compatibility, explicit interfaces are *only* available through the `mpi_f08` module since the Fortran 2008 standard introduced `TYPE(*)` and `DIMENSION(..)` (see [MPI-4.1 standard: MPI for Different Fortran Standard Versions](https://www.mpi-forum.org/docs/mpi-4.1/mpi41-report/node470.htm#Node470)) that facilitate the use of choice buffers.

What this means practically is that while you *can* use routines like `mpi_gather`, you *cannot* include them in the `only: [routines...]` argument list. So, no compile time error checking is available for such routines. 

# Conclusions and Further Action

Including a CI job for Intel MPI is trivial but requires one of the following mutually exclusive changes:
1. Any MPI integration tests should use the module `mpi_f08`, not `mpi`. The use of `mpi_f08` is [strongly recommended by the MPI-3.1 specification](https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node409.htm) anyways.
[...]

PR #271, opened 2025-01-30
Spack configuration

jwallwork23:
Closes #122.

PR #438, opened 2025-10-05
CI for Intel (ifx) and Intel Classic (ifort) each with OpenMPI

jfdev001:
Co-authored-by: Jack Atkinson <jwa34@cam.ac.uk>

# Summary

Building off work in https://github.com/Cambridge-ICCS/FTorch/pull/164 (and therefore solve https://github.com/Cambridge-ICCS/FTorch/issues/140), this PR attempts a more atomic contribution to the CI for compiling and testing FTorch with only the intel and intel classic compilers---in contrast, the original PR opened a year ago makes macos, nvfortran, and intel CI contributions but was failing tests at the time. My contribution follows the structure of pFUnit's CI file
[pFUnit/.github/workflows/main.yml](https://github.com/Goddard-Fortran-Ecosystem/pFUnit/blob/1a915dd87d1ed17500eb4e5a07c68e5f398377f1/.github/workflows/main.yml), that is I split the GNU testing and the Intel testing jobs in order to more clearly isolate the different environments. I use OpenMPI compiled from source in order to successfully run the integration tests.

# Using OpenMPI instead of oneAPI MPI

The MPI implementation that ships with Intel oneAPI is **not** used since the below error occurs when using that:

``` [ 94%] Building Fortran object
examples/7_MPI/CMakeFiles/mpi_infer_fortran.dir/mpi_infer_fortran.f90.o
/home/runner/work/FTorch/FTorch/examples/7_MPI/mpi_infer_fortran.f90(15):
error #6580: Name in only-list does not exist or is not accessible.
[MPI_GATHER] mpi_gather, mpi_init -------------------^ compilation
aborted for
/home/runner/work/FTorch/FTorch/examples/7_MPI/mpi_infer_fortran.f90
(code 1) gmake[2]: ***
[examples/7_MPI/CMakeFiles/mpi_infer_fortran.dir/build.make:78:
examples/7_MPI/CMakeFiles/mpi_infer_fortran.dir/mpi_infer_fortran.f90.o]
Error 1 gmake[1]: *** [CMakeFiles/Makefile2:1717:
examples/7_MPI/CMakeFiles/mpi_infer_fortran.dir/all] Error 2 gmake: ***
[Makefile:146: all] Error 2 ```

That is, `mpi_gather` is not recognized when compiling with the Intel
oneAPI MPI Fortran compiler. I verified this behavior by checking to see
which MPI `FindMPI` gets:

``` -- Found MPI_C: /opt/intel/oneapi/mpi/2021.16/lib/libmpifort.so (found version "4.1") -- Found MPI_CXX:
/opt/intel/oneapi/mpi/2021.16/lib/libmpicxx.so (found version "4.1") -- Found MPI_Fortran: /opt/intel/oneapi/mpi/2021.16/lib/libmpifort.so (found version "4.1") -- Found MPI: TRUE (found version "4.1")
```

The above error can be found in the logs for [github job/51782067191](https://github.com/jfdev001/FTorch/actions/runs/18189775089/job/51782067191).

**Instead, I compile OpenMPI from source** using either `ifx` or `ifort`. The C/C++ compilers for OpenMPI are `icx` and `icpx`, respectively. I do not use the Intel classic C/C++ compilers (`icc` and `icpc`, respectively) since this appears to cause the below error:

``` Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS
OpenMP_CXX_LIB_NAMES)
```

The above error can be found in the logs for [github
job/519939443997](https://github.com/jfdev001/FTorch/actions/runs/18263208690/job/51993943997).

I use OpenMPI v4.1.2 because this is just what happens to be on the
HPC systems that I use the most (DKRZ's Levante), though a different
version could certainly be used.

# Open questions---possibly new issues/new PRs

* ~Compiling OpenMPI from source takes about 7 minutes. I have tried to
  cache the resulting binary using the [actions/cache](https://github.com/actions/cache) action but it does not
seem to speed up this step. I'm probably doing something wrong, any
ideas what? I can work it out if the maintainers are generally happy
with the direction of my PR.~ Caching appears to be working... see [job/51998637228](https://github.com/jfdev001/FTorch/actions/runs/18265247725/job/51998637228)
* I only use the Intel C/C++ compilers (i.e., `icx` and `icpx` v2023.2)
  for OpenMPI to circumvent an OpenMP issue that arises when using the
Intel Classic C/C++ compilers (i.e., `icc` and `icpc`). Does this seem
like a reasonable approach?
* Unlike the original PR, I only call the [fortran-lang/setup-fortran](https://github.com/fortran-lang/setup-fortran)
action with the `intel-classic` compiler option since this also gets an
older version (v2023.2) of the newer Intel Fortran and C/C++ compilers
(i.e., `ifx`, `icx`, `icpx`). Should I instead try and use a more recent
version of `ifx` (e.g., v2024.1)?
* The [fortran-lang/setup-fortran](https://github.com/fortran-lang/setup-fortran)
action builds oneAPI MPI distribution by default. This is not used by
the current iteration of the CI, and is therefore a waste of time.
Perhaps we should manually install the Intel Compilers we need (see
again [pFUnit/.github/workflows/main.yml#L147](https://github.com/Goddard-Fortran-Ecosystem/pFUnit/blob/1a915dd87d1ed17500eb4e5a07c68e5f398377f1/.github/workflows/main.yml#L147)?
Otherwise, we could follow the approach of [dealii/.github/workflows/linux.yml#L251](https://github.com/dealii/dealii/blob/3b7dfc0ff88c7b9f3aa7112f3956e4ade892209b/.github/workflows/linux.yml#L251)
and have an action that pulls **some** of what we need? Note that the
action [rscohn2/setup-oneapi](https://github.com/rscohn2/setup-oneapi)
does not yet seem to implement an easy setup of classic compilers, but
rather is tailored for the modern Intel compilers only.


jatkinson1000 (changes requested):
Thanks @jfdev001 this seems like a reasonable approach to me and better to get halfway there than not at all.

I'm not wild about the repeated code, but it seems this is the way that makes it work.
I considered separate files, but after some reflection agree that what you suggest is the best approach. However,

- [ ] Please could you add to the header of the workflow that it contains multiple jobs, and any changes to one should always be reproduced in the others.

> I only use the Intel C/C++ compilers (i.e., icx and icpx v2023.2)
for OpenMPI to circumvent an OpenMP issue that arises when using the
Intel Classic C/C++ compilers (i.e., icc and icpc). Does this seem
like a reasonable approach?

I think so. It should catch all except the most unusual bugs, and if there is an incompatibility somewhere on a system I'd hope people are more widely aware already.

> Unlike the original PR, I only call the [fortran-lang/setup-fortran](https://github.com/fortran-lang/setup-fortran)
action with the intel-classic compiler option since this also gets an
older version (v2023.2) of the newer Intel Fortran and C/C++ compilers
(i.e., ifx, icx, icpx). Should I instead try and use a more recent
version of ifx (e.g., v2024.1)?

More recent would be nice I suppose, but we're not doing anything so esoteric here that it's a concern, and I'm also conscious of not burning unnecessary CPU cycles.

My question here is whether this is running just the new intel compilers or also the classic? It seems the classic are fetched as part of the setup, but the job is only written for the newer 'x' compilers?

>The [fortran-lang/setup-fortran](https://github.com/fortran-lang/setup-fortran)
action builds oneAPI MPI distribution by default. This is not used by
the current iteration of the CI, and is therefore a waste of time.
Perhaps we should manually install the Intel Compilers we need (see
again [pFUnit/.github/workflows/main.yml#L147](https://github.com/Goddard-Fortran-Ecosystem/pFUnit/blob/1a915dd87d1ed17500eb4e5a07c68e5f398377f1/.github/workflows/main.yml#L147)?
Otherwise, we could follow the approach of [dealii/.github/workflows/linux.yml#L251](https://github.com/dealii/dealii/blob/3b7dfc0ff88c7b9f3aa7112f3956e4ade892209b/.github/workflows/linux.yml#L251)
and have an action that pulls some of what we need? Note that the
action [rscohn2/setup-oneapi](https://github.com/rscohn2/setup-oneapi)
does not yet seem to implement an easy setup of classic compilers, but
rather is tailored for the modern Intel compilers only.

Hmm. I like the use of Fortran-lang, as we may well extend to other compilers in future, but that's a valid point. Do you have any idea how long the timing compares between the two?
I'm inclined to say let's use this as it is.
Looking at the 12min runtime we could consider making this non-automatic, run on merges to main and by explicitly calling before merge as I proposed for the GPU runner in #410. This does rely on someone remembering to do this, however!
I'll pull in @jwallwork23 for a review at this point as he is particularly clued up on efficient use of CI.

Other notes:

- [ ] There are a number of `TODO`s floating that I'd want clearing up before merge 😉 
- [ ] I note that the job name does not make it overly clear what is running (e.g. https://github.com/Cambridge-ICCS/FTorch/actions/runs/18265325920/job/52016830096?pr=438 with long job name) is it possible to remove the `CC` and `CXX` paths from the description and perhaps rename the `toolchain` to `intel-ifx` and `intel-ifort` to reflect the fact that `icx` and `icpx` are used for both builds?
    I think this could be done by moving the `CC` and `CXX` definitions to be under the toolchain. Repetition, I know, but would make jobname cleaner and intent explicit.
- [ ] Can you have a quick scan through the online docs (likely developer page) to see if anything needs updating to reflect this change?

Finally, thank you for including co-authorship! 🫶 
[GitHub's advice](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) is to also append to the commit message:
```

> Co-authored-by: jatkinson1000 <jwa34@cam.ac.uk>
```

PS: I made issue #140 a [hacktoberfest](https://hacktoberfest.com/) one if you are participating in that at all.

jfdev001:
@jatkinson1000 

I will work on addressing your comments and will naturally re-request review when I'm done.

I guess the co-authorship needs to go at the end of the commit message! I put it it at the beginning, my mistake :)) 

image

Easy fix of course!

It is purely coincidental that I am contributing to FTorch this month, completely forgot about hacktoberfest. Cool that it coincides, though.

.github/workflows/test_suite_ubuntu.yml
```diff
@@ -33,14 +34,23 @@ concurrency:
...
+          - {compiler: gcc, version: 13}
+          - {compiler: gcc, version: 12}
+          - {compiler: gcc, version: 11}
+          - {compiler: gcc, version: 10}
+          - {compiler: gcc, version: 9}
+        std: ["f2008"]
+        include:
+          # Run 2018 Fortran standards on latest gcc only
+          - toolchain: {compiler: gcc, version: 13}
+            std: "f2018"
```
joewallwork: It would be excessive to run tests with these 5 different combinations every time a commit is pushed. It'd be fine to stick to running with both Fortran standards at v13 but the f2008 v9-v12 runs should only happen upon merging into `main`. You should be able to configure this in the same workflow using conditional logic.

.github/workflows/test_suite_ubuntu.yml
```diff
@@ -110,3 +134,157 @@ jobs:
...
+
+      - name: Check Intel compiler paths 
+        run: |
+          echo $(ls /opt/intel/oneapi/compiler/)
+
+          echo "ifx: $(which ifx)"
+          echo "icx: $(which icx)"
+          echo "icpx: $(which icpx)"
+
+          echo "ifort: $(which ifort)"
```
joewallwork: Is this debugging code? It seems like something that should be removed before merging.

jwallwork23 (changes requested):
Thanks for your work on this. It seems promising but I have some concerns from a green software engineering point of view.

The Intel workflow seems to include quite a lot of work, which I feel it would be excessive to run every time someone pushes to the repo. For the purposes of this PR, I recommend changing the triggers so that the Intel job only runs upon merging into `main`. (Similarly for GCC v9-v12.)

It was already taking a long time for us to build pFUnit in the CI but building OpenMPI under the Intel compiler adds to this. Both of these could benefit from a packaging solution (as proposed in #271 and #429), but that would be for a future PR.

.github/workflows/test_suite_ubuntu.yml
jatkinson1000: In fairness to @jfdev001, I think that is what I had done in #164 that formed a basis of this PR, but I agree.

.github/workflows/test_suite_ubuntu.yml
```diff
@@ -33,14 +34,57 @@ concurrency:
...
+        matrix: ${{ steps.set-matrix.outputs.matrix }}
+      env:
+        EVENT_NAME: ${{ github.event_name }}
+        BASE_REF: ${{ github.base_ref || '' }}
+        REF: ${{ github.ref }}
+      steps:
+        - id: set-matrix
+          run: |
+            if [[ "$EVENT_NAME" == "pull_request" && "$BASE_REF" == "main" 
+                  || "$EVENT_NAME" == 'push' && "$REF" == 'refs/heads/main' ]]
```
joewallwork: Would it not be sufficient to just check if the event name is `push`? Because the workflow triggers would then imply that the branch is `main`.

I don't follow why the other `pull_request` event is included? Surely most PRs will have base reference of `main` so this part of the workflow will be triggered frequently?

.github/workflows/test_suite_ubuntu.yml
```diff
@@ -110,3 +168,144 @@ jobs:
...
+      - name: Check OpenMPI installation
+        run: |
+          echo $(which mpicc)
+          mpicc --version
+
+          echo $(which mpif90)
+          mpif90 --version
+
+          echo $(which mpiexec)
+          mpiexec --version
```
joewallwork: This looks like debugging code, please remove it.

jwallwork23 (changes requested):
Thanks for these changes. I have a couple more requests.

Also thank you for offering to take a look at the Spack issue - that'd be much appreciated! I most of the way there but was stuck with a linking issue for pFUnit.

.github/workflows/test_suite_ubuntu.yml
jfdev001: > Would it not be sufficient to just check if the event name is push? Because the workflow triggers would then imply that the branch is `main`.

You're right, I'll make this change.

> I don't follow why the other `pull_request` event is included? Surely most PRs will have base reference of main so this part of the workflow will be triggered frequently?

You're also right here. I can just do something like

```
if [[ "$EVENT_NAME" == 'push' ]]
```

I mistook pull_request for a merge event--which is not relevant anyways because pushing to main is itself a merge event---at least I infer this based on:

```
on:
  # Triggers the workflow on pushes to the "main" branch, i.e., PR merges
  push:
    branches: [ "main" ]
```

So, I thought to get the desired behavior I'd have to include the pull request in the conditional expression. But this is clearly not the case, and I see this now! Thanks!

.github/workflows/test_suite_ubuntu.yml
```diff
@@ -33,14 +34,54 @@ concurrency:
...
+                ]
+              }'
+            else
+              # Only GCC 13 for everything else
+              MATRIX_JSON='
+              {
+                "toolchain": [
+                  {"compiler": "gcc", "version": 13}
+                ],
+                "std": ["f2008, f2018"]
```
joewallwork: I think perhaps this is what you need?
```suggestion
                "std": ["f2008", "f2018"]
```

jwallwork23 (approved):
Beyond the test failure and the patch suggestion I just provided, I'm happy with how this is set up now and can see that only the specified jobs get triggered in the PR. Thanks!

Good to go (assuming tests pass).

jatkinson1000 (commented):
Hi @jfdev001 looking good

Yes, please update changelog, and I agree with suggestion to squash.

It seems there is one sticking point in the action regarding [template-injection](https://docs.zizmor.sh/audits/#template-injection). It seems odd as you haven't changed much of this particular step, but my guess is that it relates to the dynamic matrix that is now in use...??

So close now! 🚀

jatkinson1000 (approved):
This is all looking good for me now @jfdev001 🚀 
Just resolved a merge conflict introduced by #442 but will now squash and merge once CI passes.

Thanks again for contributing!

PR #441, opened 2025-10-07
Use Zizmor action

jwallwork23:
It turns out there's a Zizmor GitHub Action that we could be using instead of the current manual approach: https://docs.zizmor.sh/integrations/.

With this action, we get Security info (see comment below), which I also address in this PR. Namely:
* Avoid giving anything write permissions that doesn't need it.
* Pin community actions to a specific hash.

Also includes some relabelling for consistency.

### DIFF
--- a/.github/workflows/JOSS_paper_pdf.yml
+++ b/.github/workflows/JOSS_paper_pdf.yml
@@ -14,10 +14,17 @@ on:
       - '.github/workflows/JOSS_paper_pdf.yml'
       - 'paper/*'
 
+# No write permission by default
+permissions: {}
+
 jobs:
-  paper:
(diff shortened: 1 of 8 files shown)


jatkinson1000 (approved):
Hey @jwallwork23 
I mean, sure.
Surprised there are a few things that hadn't been previously picked up.

jwallwork23:
> Surprised there are a few things that hadn't been previously picked up.

I think it might be a version difference thing because we pinned a version for `zizmor` in `requirements-dev.txt`. This prompted me to drop it as no longer needed in https://github.com/Cambridge-ICCS/FTorch/pull/441/commits/c8b96fac59f48056fae434851af0ace647d4dbb1.

PR #442, opened 2025-10-07
Adopt Green CI good practices

jwallwork23:
Merges into #441 

FTorch is already doing pretty well in terms of Green CI usage, although there are two recommendations from https://github.com/Cambridge-ICCS/green-ci that are missing for most workflows:
1. Cancel running jobs when the workflow is re-triggered.
2. Set a time limit lower than the default 6 hours to avoid wasted resources when jobs stall.

### DIFF
--- a/.github/workflows/JOSS_paper_pdf.yml
+++ b/.github/workflows/JOSS_paper_pdf.yml
@@ -28,6 +28,9 @@ jobs:
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
 
+    # Terminate the job if it runs for more than 5 minutes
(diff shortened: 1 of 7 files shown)


jatkinson1000 (approved):
@jwallwork23 
Yes, fine, though do you want to rename static_analysis.yml to linting.yml?

jatkinson1000:
Yep, not a sticking point, I'll leave it to you.

jwallwork23:
[Pushed b21c533 to fix the CI error with publishing the website (see https://github.com/Cambridge-ICCS/FTorch/actions/runs/18317824558/job/52162985315)]

PR #443, opened 2025-10-07
Fix website deployment CI job

joewallwork:
Apologies, my bad. This PR fixes the error reported in https://github.com/Cambridge-ICCS/FTorch/actions/runs/18318178055/job/52164034473 that was introduced in #441 and #442.


jatkinson1000 (approved):
Thanks for the patch @jwallwork23

PR #445, opened 2025-10-15
Bump CUDA version used in GPU testing CI job

### LINKED ISSUE #444: PyTorch have removed CUDA 11 binary from install
CUDA workflow failing on latest merge to main.

This seems to be because PyTorch have removed the CUDA 11 wheel.
We should upgrade to 12.

jatkinson1000:
To close #444 

Bumps CUDA version requested as wheel in GPU CI job.

### DIFF
--- a/.github/workflows/test_suite_ubuntu_cuda.yml
+++ b/.github/workflows/test_suite_ubuntu_cuda.yml
@@ -24,6 +24,10 @@ on:
       # Build system
       - '**CMakeLists.txt'
       - '**requirements.txt'
+  
+  pull_request:
+    paths:
+      - '.github/workflows/test_suite_ubuntu_cuda.yml'
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:


.github/workflows/test_suite_ubuntu_cuda.yml
```diff
@@ -58,15 +62,15 @@ jobs:
...
+          python-version: '3.13'  # Use 3.13 as no PyTorch wheels for 3.14 yet.
 
       - name: Install PyTorch
         run: |
           python -m pip install --upgrade pip
           python -m venv ftorch
           . ftorch/bin/activate
-          # CUDA v11.8 chosen as the GPU-runner has v12.4 installed (backward compatible)
-          pip install torch torchvision --index-url https://download.pytorch.org/whl/cu118
+          # CUDA v12.9 chosen as latest - check what the GPU-runner has installed (backward compatible)
```
TomMelt: Does this mean, we have select v12.9 because the GPU instance has installed 12.9 drivers? If so could we re-word it to something like:

```suggestion
          # CUDA v12.9 chosen to match version installed by GPU-runner
```

Or did we just pick it because (as you mention) it's backwards compatible so it doesnt really matter. In which case I dont think we need the comment at all :shrug:

jatkinson1000: The original comment from @niccolozanotti suggested the runners may not have latest CUDA.
I decided to start with 12.9 to keep us as far ahead of the curve, planning to iterate backwards, but it seems 12.9 works 🤷

jatkinson1000: Always a challenge when modifying someone else's comments.
I've made it much more explicit in https://github.com/Cambridge-ICCS/FTorch/pull/445/commits/ad7da411acb0d712ff0ad2a17eccfd69b73112e1 :)

TomMelt: From the [logs](https://github.com/Cambridge-ICCS/FTorch/actions/runs/18523265051/job/52787981269?pr=445) it looks like version 12.0 is installed. Albeit I find CUDA versions confusing between the libs, the drivers and compute capability etc.

```
-- Found CUDA: /usr (found version "12.0") 
-- Found CUDAToolkit: /usr/include (found version "12.0.140") 
```

I guess tests will start failing if something goes wrong

jatkinson1000:
CI previously passed, cancelled here as long job and will re-run on main following merge.

.github/workflows/test_suite_ubuntu_cuda.yml
niccolozanotti: > The original comment from @niccolozanotti suggested the runners may not have latest CUDA. I decided to start with 12.9 to keep us as far ahead of the curve, planning to iterate backwards, but it seems 12.9 works 🤷

@jatkinson1000 Just to add to this, at the time of writing that, i.e. ~3 months ago, I recall the logs reporting the driver version to be `12.4`. I did some research online and I took that choice as it seemed the "safest" out of familiarity with handling CUDA driver versions. 
This makes more sense now :)

PR #437, opened 2025-10-02
link issue in example (fixes #435)

jfdev001:
See https://github.com/Cambridge-ICCS/FTorch/issues/435#issuecomment-3351700096

Issue #436, opened 2025-09-30
CI for building standalone examples


jfdev001:
> It's yours @jfdev001 and thanks for the details and looking into this.
> 
> Indeed, it was something I thought about last night that we should perhaps have a separate 'standalone examples' CI job. 

 _Originally posted by @jatkinson1000 in [#435](https://github.com/Cambridge-ICCS/FTorch/issues/435#issuecomment-3350147699)_

The examples that get run as part of integration testing are built using `libftorch.so` that is generated by the *build* phase (i.e., `cmake --build .`) but this does not guarantee functionality of the `libftorch.so` that gets generated by the *install* phase (i.e., `cmake --install .`). Since the user will link against the `libftorch.so` that is generated by the install phase, it makes sense to provide CI that checks to make sure "standalone examples" run using the installed library target. See https://github.com/Cambridge-ICCS/FTorch/issues/435#issuecomment-3351700096 for more details. 

Related also to https://github.com/Cambridge-ICCS/FTorch/pull/304

jfdev001:
@joewallwork I left one comment just now in your PR, but otherwise yes, it solves the issue :)

I'll close this issue once your PR gets merged

joewallwork:
Oh I see it now.

PR #448, opened 2025-10-23
Support building FTorch as static library

jfdev001:
# Summary
FTorch is hard-coded to build as a shared library currently. This PR provides the standard approach (see [cmake docs: BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html)) to building it as a static library. Moreover, following the standard approaches of [fortran-lang/stdlib](https://github.com/fortran-lang/stdlib), [pFUnit](https://github.com/Goddard-Fortran-Ecosystem/pFUnit), and [fftpack](https://github.com/fortran-lang/fftpack) I provide a separate cmake configuration file that forces downstream targets linking to an installed FTorch static library to find the correct Torch dependency. While PR https://github.com/Cambridge-ICCS/FTorch/pull/437 ensures that the FTorch *shared* library finds the correct Torch dependency, no such guarantee was previously available for an FTorch *static* library without the cmake configuration file added as part of the current PR.

# Justification for supporting static library build

While there is existing debate about the merits of static libraries as a whole, at least on HPC systems it certainly has clear benefits since errors due to shared libraries can be mitigated if all the libraries of a binary are built statically. While of course not a rigorous proof, consider the following two climate/weather models that rely on static libraries:

(1) The first specific case for reliance on static external libs is the [ICON](https://gitlab.dkrz.de/icon/icon-model/-/tree/289490d6c166f27c280df246c6df2da9446c8c9b/) model. By default, all external libs are built statically. You can verify this with the following if you wish:

```shell
# Get the icon model
git clone --depth 1 --branch release-2025.04-public https://gitlab.dkrz.de/icon/icon-model.git
cd icon-model

# Get the external libraries
git submodule update --init 

# list the names of directories in externals and use that name to search for 
# corresponding static libraries in the configure.ac 
# e.g., externals/comin will search for comin.*\.a and if found prints externals/comin
# output:
# externals/cdi/
# externals/comin/
# externals/ecrad/
# externals/fortran-support/
# externals/iconmath/
# externals/mtime/
# externals/ppm/
# externals/rte-rrtmgp/
# externals/sct/
# externals/tixi/
# externals/yac/
# externals/yaxt/
ls externals/ | xargs -I {} grep configure.ac -e "{}/.*\.a" | grep -o -e "externals\/[a-zA-Z]*-*[a-zA-Z]*/" | sort -u 

# note: there are 4 remaining external libs not addressed by the above grep commands
# * probtest -- a python lib, so static lib not applicable 
# * hd -- this is an application, not a library...  see externals/hd/code/Makefile
# * jsbach -- not actually external lib, object files generated get linked directly 
#                    as part of ICON build system, see icon.mk.in
# * art -- not actually external lib, object files generated get linked directly 
#              as part of ICON build system, see icon.mk.in
```

(2) A second specific case for which I am admittedly less familiar, is the [WRF](https://github.com/wrf-model/WRF) model which also relies heavily on static libs---both internal and external. You can verify this with the following if you wish:

```shell
# Get recent version of WRF
git clone --depth 1 --branch release-v4.7.1 https://github.com/wrf-model/WRF

# Search for static libs in externals built with cmake 
grep -liE "(add_library)*.*static.*" $(find external -name "CMakeLists.txt") | grep -o "external/[a-zA-Z]*_*[a-zA-Z]*" | sort -u
# output:
# external/atm_ocn
# external/esmf_time
# external/fftpack
# external/io_adios
# external/io_esmf
# external/io_grib
# external/io_int
# external/io_netcdf
# external/io_netcdfpar
# external/io_pio
# external/io_pnetcdf
# external/RSL_LITE

# You can also see in general static libraries are built in many other places in WRF
grep -E "(add_library)*.*STATIC.*" $(find . -name "CMakeLists.txt" -not -path "./external/*")
```

# Small note on why `INTERFACE stdc++` was added 

This was added to `target_link_libraries` since otherwise the below error gets thrown by downstream targets

```
/usr/bin/ld: /home/jf01/dev/FTorch-contrib/build/build-pytorch-cpu-build-shared-libs-OFF/libftorch.a(ctorch.cpp.o): undefined reference to symbol '_ZNSt6localeD1Ev@@GLIBCXX_3.4'
/usr/bin/ld: /lib/x86_64-linux-gnu/libstdc++.so.6: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
gmake[2]: *** [CMakeFiles/tensor_manipulation.dir/build.make:100: tensor_manipulation] Error 1 
gmake[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/tensor_manipulation.dir/all] Error 2        
gmake: *** [Makefile:91: all] Error 2
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,7 +176,7 @@ if(CMAKE_BUILD_TESTS)
 
   # Check if Python is in a virtual environment by checking the VIRTUAL_ENV
   # environment variable exists
-  if(NOT DEFINED ENV{VIRTUAL_ENV})
+  if(NOT DEFINED ENV{VIRTUAL_ENV} AND NOT DEFINED ENV{CONDA_PREFIX})
     message(FATAL_ERROR
             "No Python virtual environment detected. Please activate one.")
   endif()
--- a/.github/workflows/test_suite_ubuntu.yml
+++ b/.github/workflows/test_suite_ubuntu.yml
@@ -86,8 +86,8 @@ jobs:
       fail-fast: false
       matrix: ${{ fromJSON(needs.setup-gnu-matrix.outputs.matrix) }}
 
-    # Terminate the job if it runs for more than 10 minutes
-    timeout-minutes: 10
+    # Terminate the job if it runs for more than 15 minutes
+    timeout-minutes: 15
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
@@ -127,13 +127,13 @@ jobs:
 
       - name: Install pFUnit
         run: |
-          export FC=/usr/bin/gfortran
-          export MPIF90=/usr/bin/mpif90
           # TODO: Avoid version pinning (needed because version appears in install path)
           git clone -b v4.12.0 https://github.com/Goddard-Fortran-Ecosystem/pFUnit.git
           mkdir pFUnit/build
           cd pFUnit/build
-          cmake ..
+          cmake ..\
+            -DCMAKE_Fortran_COMPILER=gfortran\
+            -DMPI_Fortran_COMPILER=mpif90
           make -j 4 install
 
       - name: Build FTorch
@@ -152,9 +152,9 @@ jobs:
           cmake .. \
             -DPython_EXECUTABLE="$(which python)" \
             -DCMAKE_BUILD_TYPE=Release \
-            -DCMAKE_Fortran_COMPILER=${FC} \
-            -DCMAKE_C_COMPILER=${CC} \
-            -DCMAKE_CXX_COMPILER=${CXX} \
+            -DCMAKE_Fortran_COMPILER=gfortran  \
+            -DCMAKE_C_COMPILER=gcc  \
+            -DCMAKE_CXX_COMPILER=g++  \
             -DCMAKE_INSTALL_PREFIX=${BUILD_DIR} \
             -DCMAKE_BUILD_TESTS=TRUE \
             -DCMAKE_PREFIX_PATH="${PFUNIT_DIR};${Torch_DIR}" \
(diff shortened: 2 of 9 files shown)


CMakeLists.txt
```diff
@@ -8,11 +8,13 @@ project(
...
   LANGUAGES C CXX Fortran)
 
+option(BUILD_SHARED_LIBS "Build using shared libraries" ON)
+
 if(WIN32)
   # if building on windows we need to make sure the symbols are exported
   set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS TRUE)
   # required to build the .dll on windows
-  set(BUILD_SHARED_LIBS TRUE)
+  set(BUILD_SHARED_LIBS ON)
```
jatkinson1000: Why does this need setting here if it has now been set in the option definition above?

CMakeLists.txt
```diff
@@ -126,7 +127,9 @@ set_target_properties(
   Fortran_MODULE_DIRECTORY "${CMAKE_BINARY_DIR}/modules")
 # cmake-format: on
 # Link TorchScript
-target_link_libraries(${LIB_NAME} PRIVATE ${TORCH_LIBRARIES})
+# NOTE: Linking stdc++ for downstream targets fixes ctorch link error
+target_link_libraries(${LIB_NAME}
+  PRIVATE ${TORCH_LIBRARIES} INTERFACE stdc++)
```
jatkinson1000, line 132: I'd like to check this on mac.
Mac can be weird when it comes to their stdlibs

CMakeLists.txt
jfdev001: You're right, it seems like it's redundant to keep line 17 in. I've removed it *and* line 16 now!

CMakeLists.txt
jfdev001: I think this is another instance where standalone CI would've caught this (see also https://github.com/Cambridge-ICCS/FTorch/issues/436). This issue stdc++ linker error only comes up when using the installed version of FTorch in a downstream target. I only tested locally on Ubuntu 24.04, so might be different on macos of course like you suggested.

It may not, however, be necessary to incorporate CI for FTorch built/installed as a *static* lib since even major libraries that are known for extensive testing e.g., [dealii](git@github.com:dealii/dealii.git) don't bother to test for this sort of thing, rather they only test their build as a *shared* lib.

jfdev001:
@jatkinson1000 

> However, to help guide these users I would like to see a section in the online documentation with a basic discussion of static vs shared and making it clear what FTorch does so people can understand what they are getting and how to use it appropriately. Add wherever it seems sensible for now, as I know I really need to go and sort out what have become spaghetti docs.

Is it possible we can migrate this to a different issue/PR for now? We might have to go through several iterations of documentation revision to clearly explain the static/shared lib differences as well as how FTorch fits into this. I could come back to doing this documentation PR next weekend.

In the interim, I could update the `README.md` section by adding that we support the option `BUILD_SHARED_LIBS` so that users are at least aware of it! **Thoughts?** https://github.com/Cambridge-ICCS/FTorch/blob/01bae6e548a5a5eccf2e46f78a9e1cc63f61db3f/README.md?plain=1#L158-L170

CMakeLists.txt
jatkinson1000: I think also line 16 was a comment referring to 17?

jfdev001: presumably yes! Though i am not familiar with that Win32 conditional block, and have not done any testing on windows (looks like the windows CI is broken anyways right now, https://github.com/Cambridge-ICCS/FTorch/issues/300)

jatkinson1000: It is 🥲 🪟

jatkinson1000:
Thanks.

My understanding is that that if the user wants to build a static version of their application (ICON, WRF, etc.) then they _must_ have built FTorch statically and link against that. To be explicit, they cannot build/link a static application against `libftorch.so`.

(Building their project as a dynamic library against a statically built FTorch I can see might work, though I guess they would need to statically link FTorch and have it included in their final `.so`.)

jfdev001:
> My understanding is that that if the user wants to build a static version of their application (ICON, WRF, etc.) then they _must_ have built FTorch statically and link against that. To be explicit, they cannot build/link a static application against `libftorch.so`.

Ahhh, I see what you mean! You're understanding is certainly correct, sorry for the confusion on my part. So if a user is compiling their application with e.g., `gcc` and they pass the `-static` option during compilation, then per [man GCC](https://man7.org/linux/man-pages/man1/gcc.1.html), they would be unable to link their application against shared libs (i.e., it would be impossible to link against libftorch.so under these conditions).

PR #449, opened 2025-10-24
Intel CI now uses Intel oneAPI MPI instead of OpenMPI built with Intel compilers (fixes #140)

### LINKED ISSUE #140: Add intel ifx and ifort build to CI

jfdev001:
Resolves https://github.com/Cambridge-ICCS/FTorch/issues/140

Improvement of original fix for issue https://github.com/Cambridge-ICCS/FTorch/issues/140 that was introduced in PR https://github.com/Cambridge-ICCS/FTorch/pull/438

Note, this is a nice improvement in build time in the intel CI github workflow:

Assuming uncached, Intel oneAPI MPI setup now only needs ~5 minutes to build compared with the
original ~10 minutes needed for the OpenMPI built with Intel compilers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,8 +165,8 @@ To build and install the library:
     | [`CMAKE_PREFIX_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html)        | `</path/to/LibTorch/>`                  | Location of Torch installation<sup>2</sup>                                                                                                                                 |
     | [`CMAKE_INSTALL_PREFIX`](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html)  | `</path/to/install/lib/at/>`            | Location at which the library files should be installed. By default this is `/usr/local`                                                                                   |
     | [`CMAKE_BUILD_TYPE`](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html)          | `Release` / `Debug`                     | Specifies build type. The default is `Debug`, use `Release` for production code                                                                                            |
-    | `CMAKE_BUILD_TESTS`                                                                               | `TRUE` / `FALSE`                        | Specifies whether to compile FTorch's [test suite](https://cambridge-iccs.github.io/FTorch/page/testing.html) as part of the build.<sup>4</sup>                            |
-    | `GPU_DEVICE`                                                                                      | `NONE` / `CUDA` / `HIP` / `XPU` / `MPS` | Specifies the target GPU backend architecture (if any) <sup>3</sup>                                                                                                        |
+    | `CMAKE_BUILD_TESTS`                                                                               | `TRUE` / `FALSE`                        | Specifies whether to compile FTorch's [test suite](https://cambridge-iccs.github.io/FTorch/page/testing.html) as part of the build.<sup>3</sup>                            |
+    | `GPU_DEVICE`                                                                                      | `NONE` / `CUDA` / `HIP` / `XPU` / `MPS` | Specifies the target GPU backend architecture (if any) <sup>4</sup>                                                                                                        |
     | `MULTI_GPU`                                                                                      | `ON` / `OFF` |  Specifies whether to build the tests that involve multiple GPU devices (`ON` by default if `CMAKE_BUILD_TESTS` and `GPU_DEVICE` are set).                                                                                                        |
 
     <sup>1</sup> _On Windows this may need to be the full path to the compiler if CMake cannot locate it by default._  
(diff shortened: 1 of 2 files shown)


.github/workflows/test_suite_ubuntu.yml
```diff
@@ -214,55 +214,34 @@ jobs:
...
+            impi@2021.14.0
 
-      - name: Add OpenMPI to env 
+      - name: Setup Intel oneAPI environment
         run: |
-          echo "PATH=$HOME/openmpi-${{ matrix.toolchain }}/bin:${PATH}" >> $GITHUB_ENV 
-          echo "LD_LIBRARY_PATH=$HOME/openmpi-${{ matrix.toolchain }}/lib:${LD_LIBRARY_PATH}" >> $GITHUB_ENV 
+          source /opt/intel/oneapi/setvars.sh
+          printenv >> $GITHUB_ENV
+          printenv | grep intel
```
joewallwork: This looks like a debug print?

jfdev001: `printenv >> $GITHUB_ENV` is necessary in order to update the github environment with the appropriate LD_LIBRARY_PATH, PATH, etc.

` printenv | grep intel` was extra, i have now removed it!

.github/workflows/test_suite_ubuntu.yml
```diff
@@ -214,55 +214,33 @@ jobs:
...
-          export MPIF90=mpif90
-          export OPENMPI_DIR=$HOME/openmpi-${{ matrix.toolchain }}
           # TODO: Avoid version pinning (needed because version appears in install path)
           git clone -b v4.12.0 https://github.com/Goddard-Fortran-Ecosystem/pFUnit.git
           mkdir pFUnit/build
           cd pFUnit/build
-          cmake .. -DCMAKE_PREFIX_PATH=${OPENMPI_DIR}
+          cmake .. \
+            -DCMAKE_C_COMPILER=${{ matrix.CC }} \
+            -DCMAKE_CXX_COMPILER=${{ matrix.CXX }} \
```
jatkinson1000: Could these be explicitly specified - to match the changes made to the gnu workflow - and remove them from the matrix leaving just the fortran compiler (which is all we are varying).
```suggestion
            -DCMAKE_C_COMPILER=icx \
            -DCMAKE_CXX_COMPILER=icpx \
```

The same could be said for the standard.

I can also see the argument of leaving them in the matrix so that the full configuration can be seen in one place, however.

.github/workflows/test_suite_ubuntu.yml
```diff
@@ -273,19 +251,16 @@ jobs:
...
-            -DCMAKE_Fortran_COMPILER=${FC} \
-            -DCMAKE_C_COMPILER=${CC} \
-            -DCMAKE_CXX_COMPILER=${CXX} \
+            -DCMAKE_C_COMPILER=${{ matrix.CC }} \
+            -DCMAKE_CXX_COMPILER=${{ matrix.CXX }} \
+            -DCMAKE_Fortran_COMPILER=${{ matrix.FC }} \
+            -DMPI_Fortran_COMPILER=${{ matrix.MPIFC }} \
             -DCMAKE_INSTALL_PREFIX=${BUILD_DIR} \
             -DCMAKE_BUILD_TESTS=TRUE \
-            -DCMAKE_PREFIX_PATH="${OPENMPI_DIR};${PFUNIT_DIR};${Torch_DIR}" \
```
jatkinson1000, line 288: I presume the new action adds OneAPI-MPI to the relevant paths which is why we can remove this?
Asking for my own interest.

README.md
```diff
@@ -181,6 +181,7 @@ To build and install the library:
     <sup>3</sup> _This is often overridden by PyTorch. When installing with pip, the `index-url` flag can be used to ensure a CPU-only or GPU-enabled version is installed, e.g.
           `pip install torch --index-url https://download.pytorch.org/whl/cpu`.
           URLs for alternative versions can be found [here](https://pytorch.org/get-started/locally/)._
+    <sup>4</sup>_To run the tests, your system's MPI must support `use mpi_f08`. Note that OpenMPI < v2.0 and MPICH < v3.1 do not support that module._
```
jatkinson1000: Ah, yes, I forgot this...
As discussed on the issue I think we can presume that most people will be using something that will support this.

Please also check the docs, I think It also needs updating in the `cmake`/build docs: https://cambridge-iccs.github.io/FTorch/page/cmake.html

jatkinson1000 (changes requested):
Thanks @jfdev001 

Generally happy, but some online docs need updating, and a query about not placing the intel compilers that are fixed in the job matrix

.github/workflows/test_suite_ubuntu.yml
jfdev001: Yup, the oneapi MPI action updates the relevant paths and then CMakes builtin FindMPI.cmake locates the correct MPI!

.github/workflows/test_suite_ubuntu.yml
jfdev001: Agreed and updated accordingly!

README.md
jfdev001: Hi @jatkinson1000 I checked the docs at https://cambridge-iccs.github.io/FTorch/page/cmake.html, it looks like ifx/ifort are already listed as compiler options, so I am not quite sure what you would like me to update in this section :) we can discuss at today's (2025-10-27) meeting as needed. As far as I can tell, the docs look good

jatkinson1000: I think the `<sup>4...` needs adding to the page as you have done in the README?

jatkinson1000: Though I see you've since removed it.
There is a whitespace change in the README now, so if you can `restore` to remove that change then we're probably good to go.

jfdev001: hi, i re-added the note to the readme, reordered the superscripts since i realized they were out of order (i.e., 4 came before 3), and added the note to the pages as well.

I agree btw that most people will be using an MPI implementation compatible with mpi_f08, so it's probably overkill to leave in the note about MPI compatibility, but at least with the note we cover our bases

jatkinson1000 (approved):
Yep, happy now @jfdev001 

Will squash and merge once tests pass.

Issue #454, opened 2025-10-27
Extend timeout on test suite


jatkinson1000:
A couple of times now I have had PRs fail the test suite for hitting the 10 min limit of #442 with 2-3 tests left to execute.

I suggest lengthening to 12.5 mins to allow completion when there is a slightly sluggish runner.

PR #455, opened 2025-10-27
Extend test suite timeout and allow running of intel on PRs if required.

### LINKED ISSUE #453: Enable running intel CI on pull requests, no just merges to main
For efficiency we set up the Intel CI so that ity only runs on merges to main, similar to CUDA.

However, having just reviewed #449 which explicitly deals with modifying the intel CI it would have been useful to be able to execute it there as a final check before merging.

jatkinson1000:
Should close #453 and #454

### DIFF
--- a/.github/workflows/test_suite_ubuntu.yml
+++ b/.github/workflows/test_suite_ubuntu.yml
@@ -175,8 +175,11 @@ jobs:
           ctest --verbose --tests-regex example
 
   Intel:
-    # Trigger intel build only if push to main 
-    if: github.event_name == 'push'
+    # Trigger intel build only if push to main, or called manually via dispatch.
+    if: |
+      (github.ref == 'refs/heads/main' && github.event_name == 'push')
+      || github.event_name == 'workflow_dispatch'
+      || (github.event_name == 'pull_request' && contains(github.event.pull_request.changed_files, '.github/workflows/test_suite_ubuntu.yml'))
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
     strategy:


.github/workflows/test_suite_ubuntu.yml
```diff
@@ -86,8 +86,8 @@ jobs:
       fail-fast: false
       matrix: ${{ fromJSON(needs.setup-gnu-matrix.outputs.matrix) }}
 
-    # Terminate the job if it runs for more than 10 minutes
-    timeout-minutes: 10
+    # Terminate the job if it runs for more than 12.5 minutes
+    timeout-minutes: 12.5
```
joewallwork: Why not just 15?

.github/workflows/test_suite_ubuntu.yml
```diff
@@ -175,8 +175,11 @@ jobs:
           ctest --verbose --tests-regex example
 
   Intel:
-    # Trigger intel build only if push to main 
-    if: github.event_name == 'push'
+    # Trigger intel build only if push to main, or called manually via dispatch.
+    if: |
+      (github.ref == 'refs/heads/main' && github.event_name == 'push')
+      || github.event_name == 'workflow_dispatch'
+      || (github.event_name == 'pull_request' && contains(github.event.pull_request.changed_files, '.github/workflows/test_suite_ubuntu.yml'))
```
joewallwork, line 182: We could add repo rules to remind us to run these tests, although I'm not necessarily recommending to do so. The advantage would be that maintainers would get a reminder to run such tests. That is:
1. Specify that the test suite workflows must pass before merging into `main`.
2. In order to merge the PR, the maintainer needs to trigger any missing workflows with manual dispatch.

The downside is then that for trivial or non-code PRs maintainers would need to either run such tests unnecessarily or bypass the tests pass rule.

joewallwork: I don't think you need the files changed logic because it's already covered by the triggers at the top.
```suggestion
    # Trigger intel build only if push to main, or called manually via dispatch.
    if: |
      (github.ref == 'refs/heads/main' && github.event_name == 'push')
      || github.event_name == 'workflow_dispatch'
      || github.event_name == 'pull_request'
```

.github/workflows/test_suite_ubuntu.yml
jatkinson1000: Done - it also turns out it has to be an integer, and I think that's why zizmor was failing.

.github/workflows/test_suite_ubuntu.yml
jatkinson1000: Agree, but that will the automatically trigger the intel section on changes to any of the listed files at the top (?).

My proposal here is that intel tests will still be skipped in general for code changes (as per the original intent) but will be run if this workflow has been changed (which it was not on #449).

joewallwork: Ah okay good point. Yeah this is fine then.

jatkinson1000: Though I also note that it hasn't been triggered here... 🤔

PR #460, opened 2025-10-31
Example 5: fix parallel .mod race by separate Fortran module dirs per target

zka26:
# Summary
Parallel builds of Example 5 create a race: good and bad both define ml_mod.mod into the same folder and in parallel, one target overwrites the other's .mod, which results in error.
## Reproduction
```vbnet
$ cmake --build . --parallel 16
[ 16%] Building Fortran object CMakeFiles/example5_simplenet_infer_fortran_good.dir/good/fortran_ml_mod.f90.o
[ 33%] Building Fortran object CMakeFiles/example5_simplenet_infer_fortran_bad.dir/bad/fortran_ml_mod.f90.o
f951: Fatal Error: Cannot rename module file ‘ml_mod.mod0’ to ‘ml_mod.mod’: No such file or directory
compilation terminated.
gmake[2]: *** [CMakeFiles/example5_simplenet_infer_fortran_bad.dir/build.make:88: CMakeFiles/example5_simplenet_infer_fortran_bad.dir/bad/fortran_ml_mod.f90.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:85: CMakeFiles/example5_simplenet_infer_fortran_bad.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
Error copying Fortran module "ml_mod.mod".  Tried "ML_MOD.mod" and "ml_mod.mod".
gmake[2]: *** [CMakeFiles/example5_simplenet_infer_fortran_good.dir/depend.make:8: CMakeFiles/example5_simplenet_infer_fortran_good.dir/ml_mod.mod.stamp] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:111: CMakeFiles/example5_simplenet_infer_fortran_good.dir/all] Error 2
gmake: *** [Makefile:91: all] Error 2
```

## Fixes #306 
Give each target its own Fortran module output directory and add it to the include path to avoid collisions.

## After
Both example5_simplenet_infer_fortran_good and example5_simplenet_infer_fortran_bad build reliably and parallel.
```vbnet
$ cmake --build . --parallel 16
[ 33%] Building Fortran object CMakeFiles/example5_simplenet_infer_fortran_bad.dir/bad/fortran_ml_mod.f90.o
[ 33%] Building Fortran object CMakeFiles/example5_simplenet_infer_fortran_good.dir/good/fortran_ml_mod.f90.o
[...]

PR #461, opened 2025-10-31
Pin Python version in test suite.

jatkinson1000:
In #445 I had to pin the version of Python used in test CI to 3.13 instead of 3.x as there were no PyTorch wheels yet available for 3.14.

It seems this has just appeared in #460 and can be seen [here where I run previously passing tests on main](https://github.com/Cambridge-ICCS/FTorch/actions/runs/18982919100).

This PR pins it, though we should be careful to unpin in future once wheels released.

Slightly worryingly it failed in `jit.script`, though I hope that is coincidental.

It also needs fixing for the cpp-linter action to get a working wheel of pygit

### DIFF
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Install Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.x'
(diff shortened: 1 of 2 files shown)


jwallwork23 (approved):
Sure, makes sense.

PR #464, opened 2025-11-22
pkg-config support

jfdev001:
While I was reading the `pages/hpc.md`, I noticed that the description for using FTorch with a make project or with other build systems could be improved by providing users with a package config file (i.e., `FTorch.pc`) so that they can compile against FTorch more easily by querying compilation flags and library flags using `pkg-config` 

This PR adds the following:

* A package config template file that could be used in conjunction with `pkg-config` (a ubiquitous linux utility) for querying relevant flags for compiling against FTorch.
* Setting new variables `CMAKE_Fortran_MODULE_DIRECTORY` and `CMAKE_INSTALL_MODULEDIR` .
* CMake project description based on the FTorch GitHub repo description.
* Update changelog
* update hpc docs to reflect using pkg-config
* update CI to double check pkg-config file 

I thought to add the package config file because I saw [fortran-lang/stdlib](https://github.com/fortran-lang/stdlib/blob/03736faab793ee52fd88e79f6d21d8bda1d294ed/config/template.pc) and [fortran-lang/fftpack](https://github.com/fortran-lang/fftpack/blob/8476c39e1b25d7b5a3a2fac980d76fc82a8c1f37/config/template.pc) do this, and we should naturally try and support users of HPC codes who may not be able to use CMake entirely for their codes.

~I think I should also update `pages/hpc.md` file to include the use of `pkg-config` to query package information, **but I wanted to open the PR first to see what you guys think, so please let me know :)**~

Another thought: maybe it's more appropriate to put `template.pc` into a different directory called say `config` and also move the `cmake/FTorchConfig.cmake.in` into that directory (this is also the approach `stdlib` takes)... so something like 

```
config/
    template.pc
    cmake/
        FTorchConfig.cmake.in  
```

but this is an organizational question that's more up to you guys I would say.

In case you're unfamiliar, `pkg-config` has output like the below:

```shell
 pkg-config --libs /path/to/FTorch/installed/libs/pkgconfig/ftorch.pc
# output: -L/path/to/FTorch/installed/libs/installed/lib -lftorch

pkg-config --cflags /path/to/FTorch/installed/libs/pkgconfig/ftorch.pc
# output: -I/path/to/FTorch/installed/include -I/path/to/FTorch/installed/include/ftorch
```

Or if you added the directory containing `ftorch.pc` to the `PKG_CONFIG_PATH`, you can simply call

```shell
pkg-config --libs ftorch
pkg-config --cflags ftorch
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@ set(PACKAGE_VERSION 1.0.0)
 project(
   ${PROJECT_NAME}
   VERSION ${PACKAGE_VERSION}
+  DESCRIPTION "A library for directly calling PyTorch ML models from Fortran"
   LANGUAGES C CXX Fortran)
 
 option(BUILD_SHARED_LIBS "Build using shared libraries" ON)
@@ -106,6 +107,11 @@ file(RELATIVE_PATH relDir ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_INSTALL_BINDIR}
 get_filename_component(TORCH_LIBRARY_DIR ${TORCH_LIBRARY} DIRECTORY)
 set(CMAKE_INSTALL_RPATH "$ORIGIN/${relDir};${TORCH_LIBRARY_DIR}")
 
+# Initialize Fortran module build output directory for all subsequent targets
+if(NOT DEFINED CMAKE_Fortran_MODULE_DIRECTORY)
+  set(CMAKE_Fortran_MODULE_DIRECTORY "${CMAKE_BINARY_DIR}/modules")
+endif()
+
 # Library with C and Fortran bindings
 add_library(${LIB_NAME} src/ctorch.cpp src/ftorch.F90 src/ftorch_test_utils.f90)
 
@@ -122,9 +128,7 @@ target_compile_definitions(
 # Add an alias FTorch::ftorch for the library
 add_library(${PROJECT_NAME}::${LIB_NAME} ALIAS ${LIB_NAME})
 # cmake-format: off
-set_target_properties(
-  ${LIB_NAME} PROPERTIES PUBLIC_HEADER "src/ctorch.h"
-  Fortran_MODULE_DIRECTORY "${CMAKE_BINARY_DIR}/modules")
+set_target_properties(${LIB_NAME} PROPERTIES PUBLIC_HEADER "src/ctorch.h")
 # cmake-format: on
 # Link TorchScript
 # NOTE: Linking stdc++ for downstream targets fixes ctorch link error
@@ -132,7 +136,7 @@ target_link_libraries(${LIB_NAME}
   PRIVATE ${TORCH_LIBRARIES} INTERFACE stdc++)
 # Include the Fortran mod files in the library
 target_include_directories(
-  ${LIB_NAME} PUBLIC $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/modules>
+  ${LIB_NAME} PUBLIC $<BUILD_INTERFACE:${CMAKE_Fortran_MODULE_DIRECTORY}>
   # $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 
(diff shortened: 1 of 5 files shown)


jfdev001:
Hi @jatkinson1000 

> Hi @jfdev001 another thought - is it possible to some sort of 'testing', even if it's just checking that pkg-config can be run and produce roughly expected output?

Absolutely! I can add to the ubuntu CI a check that calls pkg-config. I'll do this now

pages/hpc.md
```diff
@@ -63,9 +63,12 @@ with CMake to enforce this.
...
 ### Building Projects and Linking to FTorch
 
-Whilst we describe how to link to FTorch using CMake to build a project on our main
-page, many HPC models do not use CMake and rely on `make` or more elaborate build
-systems.
+Whilst we describe how to link to FTorch using CMake to build a project on our
+main page, many HPC models do not use CMake and rely on `make` or more
+elaborate build systems. This section assumes that you have successfully
+configured, built, *and* installed FTorch using CMake commands like `cmake [args...]`, 
+`cmake --build [args...]`, *and* `cmake --install [args...]`, respectively.
```
jatkinson1000: I'd be in favour of keeping concise, unless there was a particular argument to add this?
```suggestion
elaborate build systems. This section assumes that you have successfully
configured, built, *and* installed FTorch using CMake.
```

pages/hpc.md
```diff
@@ -87,7 +90,31 @@ When compiling the final executable add the following _linker_ flag:
...
+you can easily query the compiler and linker flags of FTorch rather than
+manually specifying them as was shown above. FTorch provides a standard
+pkg-config file in both the directory in which FTorch was built (e.g.,
+`</path/to/FTorch>/build`) as well as the library directory in which it was
+installed (e.g., `</path/to/FTorch/install/location>/lib/pkgconfig`). For
+example, the following commands are equivalent to adding the manually specified
+flags:
+```sh
+FCFLAGS+=" $(pkg-config --cflags </path/to/FTorch/install/location>/lib/pkgconfig/ftorch.pc)"
+LDFLAGS+=" $(pkg-config --libs </path/to/FTorch/install/location>/lib/pkgconfig/ftorch.pc)"
```
jatkinson1000, line 105: Brilliant, thanks!

pages/hpc.md
```diff
@@ -87,7 +90,31 @@ When compiling the final executable add the following _linker_ flag:
...
+
+If you do not wish to specify the entire path to that particular FTorch
+installation pkgconfig directory during every invocation of `pkg-config`, you
+can update the `PKG_CONFIG_PATH` environment variable (e.g., in `~/.bashrc`)
+accordingly. Then, you only have to specify `ftorch` as an argument to
+`pkg-config`:
+```sh
+export PKG_CONFIG_PATH=${PKG_CONFIG_PATH}:</path/to/FTorch/install/location>/lib/pkgconfig/
+FCFLAGS+=" $(pkg-config --cflags ftorch)"
+LDFLAGS+=" $(pkg-config --libs ftorch)"
```
jatkinson1000: This is quite verbose and I am working in the [`docs-clean` branch](https://github.com/Cambridge-ICCS/FTorch/tree/docs-clean) (PR incoming) to try and reduce this and make things clearer.

Perhaps something like:

> You can add FTorch to the `PKG_CONFIG_PATH` environment variable to simplify these commands:
> ```sh
> export PKG_CONFIG_PATH=${PKG_CONFIG_PATH}:</path/to/FTorch/install/location>/lib/pkgconfig/
> FCFLAGS+=" $(pkg-config --cflags ftorch)"
> Comment view
> LDFLAGS+=" $(pkg-config --libs ftorch)"
> ```

CMakeLists.txt
```diff
@@ -122,17 +128,15 @@ target_compile_definitions(
 # Add an alias FTorch::ftorch for the library
 add_library(${PROJECT_NAME}::${LIB_NAME} ALIAS ${LIB_NAME})
 # cmake-format: off
-set_target_properties(
-  ${LIB_NAME} PROPERTIES PUBLIC_HEADER "src/ctorch.h"
-  Fortran_MODULE_DIRECTORY "${CMAKE_BINARY_DIR}/modules")
+set_target_properties(${LIB_NAME} PROPERTIES PUBLIC_HEADER "src/ctorch.h")
```
jatkinson1000, line 131: Why do we no longer need to provide `CMAKE_Fortran_MODULE_DIRECTORY` to `set_target_properties()`? Was it there erroneously previously or am I missing something more notable?

pages/hpc.md
jfdev001: Agreed :)

CMakeLists.txt
jfdev001: The default value for the property `Fortran_MODULE_DIRECTORY` is set to `CMAKE_Fortran_MODULE_DIRECTORY` (see [cmake docs: Fortran_MODULE_DIRECTORY](https://cmake.org/cmake/help/latest/prop_tgt/Fortran_MODULE_DIRECTORY.html#prop_tgt:Fortran_MODULE_DIRECTORY)). By explicitly setting the `CMAKE_Fortran_MODULE_DIRECTORY`, we (1) know that all subsequent targets will have the correct `Fortran_MODULE_DIRECTORY` property without needing to specify it, (2) We can use this new variable to not repeat ourselves here:

image

pages/hpc.md
jfdev001: I'll condense this, no problem!

CMakeLists.txt
jatkinson1000: Ah, super, thanks. I wondered if it might be a specially named variable that allowed automatic inclusion.

jatkinson1000 (approved):
Thanks @jfdev001 this is great!

PR #473, opened 2026-01-06
Pedantic Zizmor fixes

joewallwork:
Easing back into some FTorch dev and thought I might as well apply some zizmor fixes under the `--pedantic` option.

* Use commit hashes consistently for actions (and bump some version numbers while I'm at it).
* Give all jobs names.
* Use `timeout-minutes` consistently.
* Provide permissions justifications consistently.
* Fix indentation.

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -55,7 +55,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6
         with:
-          python-version: '3.13'
+          python-version: '3.13'  # Use 3.13 as no PyTorch wheels for 3.14 yet.
 
       - name: Install dependencies
         run: |
@@ -104,7 +104,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6
         with:
-          python-version: '3.13'
+          python-version: '3.13'  # Use 3.13 as no PyTorch wheels for 3.14 yet.
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/fypp_checks.yml
+++ b/.github/workflows/fypp_checks.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6
         with:
-          python-version: '3.13'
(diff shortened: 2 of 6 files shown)


.github/workflows/build_docs.yml
```diff
@@ -94,10 +96,10 @@ jobs:
           sudo apt-get install -y graphviz
 
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - name: Checkout code
+      - name: Checkout repository
```
jatkinson1000: nit: Specify which repository?
And similarly elsewhere.

.github/workflows/build_docs.yml
```diff
@@ -94,10 +96,10 @@ jobs:
...
         with:
           persist-credentials: true
-        uses: actions/checkout@v4
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
       - name: Install Python
         uses: actions/setup-python@v5
```
jatkinson1000: Perhaps this should also be pinned with a hash?
And similarly elsewhere.

.github/workflows/fypp_checks.yml
```diff
@@ -27,39 +27,46 @@ permissions: {}
...
+        with:
+          persist-credentials: true
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - name: Setup Python
+        uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6
+        with:
+          python-version: '3.13'
+
+      - name: Install Fortran pre-processor
```
jatkinson1000:
```suggestion
      - name: Install fypp, the Fortran pre-processor
```

.github/workflows/build_docs.yml
joewallwork: Mentioned FTorch repo specifically in 3ed55f8.

.github/workflows/build_docs.yml
joewallwork: Done in b8b3465.

jatkinson1000: I see you have removed the comment about pinning to 3.13
Is it worth trying 3.14 or leaving the comment in if it's still an issue?

joewallwork: Ah yeah I think that comment got removed during a copy-paste. I tried 3.14 but it didn't work (see failed jobs) so have reinstated the comment across the board.

Issue #122, opened 2024-04-19
spack recipe


jatkinson1000:
One thing I discussed with @ElliottKasoar at EGU is that it would be nice if we could make it simpler to install FTorch for those less familiar with git and more familiar with things like `pip`.

the fpm suggestion of #77 is similar, but I'd argue fpm users are even more sparse than git users at the moment.

However, spack might be a nice step and @TomMelt has decent knowledge of spack.

PR #477, opened 2026-01-09
Spack build for FTorch

### LINKED ISSUE #476: Adding a spack package build
It would be nice to be able to install ftorch using spack.  Variants for all of the builds will need to be made and tested.

thomas-robinson:
This adds a package.py for building in spack.  It also adds tests to amke sure the concretizer works.  

Resolves #476 also #122

### DIFF
--- a/.spack-package/package.py
+++ b/.spack-package/package.py
@@ -3,7 +3,7 @@ import os
 from spack.package import *
 
 
-class FTorch(CMakePackage):
+class Ftorch(CMakePackage):
     """FTorch: A library for coupling PyTorch models to Fortran.
 
     FTorch enables users to directly couple their PyTorch models to Fortran code,
--- a/.github/workflows/spack_build.yml
+++ b/.github/workflows/spack_build.yml
@@ -124,8 +124,8 @@ jobs:
       - name: Add FTorch package to Spack
         run: |
           # Create a custom Spack repository for FTorch
-          mkdir -p ftorch-repo/packages/FTorch
-          cp .spack-package/package.py ftorch-repo/packages/FTorch/
+          mkdir -p ftorch-repo/packages/ftorch
+          cp .spack-package/package.py ftorch-repo/packages/ftorch/
           
           # Create repo.yaml for the custom repo
           echo "repo:" > ftorch-repo/repo.yaml


.spack-package/package.py
joewallwork, line 1: Please could you move this into a subdirectory called `spack`?

package.py
```diff
@@ -0,0 +1,174 @@
+# Copyright Spack Project Developers. See LICENSE file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
```
joewallwork: I see that this copyright statement is in the package creation tutorial (https://spack-tutorial.readthedocs.io/en/latest/tutorial_packaging.html). I'm sceptical about including it as-is here, though. For a start, the license file is different in FTorch. I'm also a bit confused about the `Apache-2.0 OR MIT` part.

.spack-package/package.py
```diff
@@ -0,0 +1,174 @@
+# Copyright Spack Project Developers. See LICENSE file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import os
+
+from spack.package import *
+
+
+class Ftorch(CMakePackage):
```
joewallwork, line 6: Could this be
```suggestion
class FTorch(CMakePackage):
```
for consistency?

package.py
```diff
@@ -0,0 +1,174 @@
...
+    supporting both CPU and GPU execution on UNIX and Windows operating systems.
+    """
+
+    homepage = "https://github.com/Cambridge-ICCS/FTorch"
+    url = "https://github.com/Cambridge-ICCS/FTorch/archive/refs/tags/v1.0.0.tar.gz"
+    git = "https://github.com/Cambridge-ICCS/FTorch.git"
+
+    license("MIT")
+
+    version("1.0.0", sha256="e6e3bb6a27f4c5b42b9e6d5c5f5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5")
```
joewallwork: What's this sha? It doesn't seem to be consistent with https://github.com/Cambridge-ICCS/FTorch/releases/tag/v1.0.0.

.spack-package/package.py
```diff
@@ -0,0 +1,174 @@
...
+        values=("none", "gfx906", "gfx908", "gfx90a", "gfx942"),
+        multi=True,
+        description="AMD GPU target(s) to build for",
+        when="+hip",
+    )
+
+    # Core dependencies
+    depends_on("cmake@3.15:", type="build")
+    depends_on("fortran", type="build")
+    depends_on("py-torch", type=("build", "link"))
```
joewallwork, line 55: One issue I had when I was first trying to set up Spack for FTorch in #271 was that `py-torch` comes with the `intel-oneapi-mkl` dependency, which includes various options by default and bloats its size. I had been using `py-torch~cuda~mpi~openmp~mkldnn`. Do you think it'd be appropriate to do something similar here?

thomas-robinson: I wouldn't want to restrict a user from building py-torch however they want.  I'm not sure if we put `py-torch~cuda~mpi~openmp~mkldnn` that it would break someone's build who was building py-torch a different way OR who used an external py-torch.  When I attempted the build, I was having issues with building the py-torch dependencies, so I just installed it externally and used it as an external.

.spack-package/package.py
thomas-robinson: I tried to move it to `spack` but when I did that, it broke the testing because github clones spack into the `spack` directory.  So it is in a folder called `.spack-package`to avoid that conflict.  I can rename it something else if you prefer.

package.py
thomas-robinson: I removed this header entirely.

.spack-package/package.py
thomas-robinson: I tried to do this, but it caused a bunch of issues when I tried to do the concretization test.  I consulted an AI bot who said that spack is specifically looking for `Ftorch` and couldn't find `FTorch`.  So after running into errors, I switched it back.  I can pursue this further if you'd like.

package.py
thomas-robinson: I think I fixed this.  Can you confirm c4b6741e582623b7ecaecd59d02f779e8a6f6017f8068c85da8a034f468df375 is right?

.spack-package/package.py
thomas-robinson: On the spack site https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#package-class-names it says that CamelCase is for packages that have a `-` so I think `class FTorch` would mean the package is named `f-torch`

joewallwork: Ah okay, no worries. Thanks for looking into it.

.spack-package/package.py
joewallwork: Thanks. That should be fine for now but I may follow up your PR by creating a `packaging` directory and moving it to `packaging/spack`, as well as moving `conda` to `packaging/conda`. No need for you to do that in this PR.

package.py
joewallwork: Thanks

.spack-package/package.py
joewallwork: Good point, a user can always just put `py-torch~cuda~mpi~mkldnn` (or some other customisation) in their spack config. Perhaps we could mention this in the docs.

package.py
joewallwork: Yes, I can confirm that's correct. Please could you update it in this `package.py` file?

.spack-package/package.py
joewallwork: Leave it with me. I'll do this in a follow-up PR.

PR #484, opened 2026-01-16
Move unit tests to veggies

### LINKED ISSUE #481: Move to test-drive?
The unit tests are currently implemented using pFUnit. Unfortunately, this dependency has caused some issues for packaging:
* In the original Spack PR (#271), I was stuck on a linking error with pFUnit.
* When looking at fpm (#429), I found that pFUnit is incompatible.

One option is for us to move the unit testing to [test-drive](https://github.com/fortran-lang/test-drive), which can be easily built with fpm (which can, in turn, be built with spack if you so wish).

joewallwork:
Closes #481
Alternative to #482

Whilst I was able to move over to `test-drive` relatively easily in #482, I found that it's lacking some functionality, e.g., for parametrisation. I also noticed that it hasn't been developed much in recent years.

Another possible unit testing framework we could use is [veggies](https://gitlab.com/everythingfunctional/veggies). That unit testing framework appears to have more mature functionality, more active development, and is also fully compatible with fpm. I made a start on setting things up.

## Tasks

- [ ] Transfer build system
- [x] Transfer CI
- [ ] Transfer unit tests
  - [ ] tensor_constructors_destructors
  - [ ] tensor_interrogation
  - [ ] tensor_manipulation
  - [x] tensor_operators
  - [ ] tensor_operator_overloads
  - [ ] tensor_operators_autograd
  - [ ] tensor_operator_overloads_autograd
  - [ ] tensor_interrogation_cuda
  - [ ] tensor_manipulation_cuda
- [ ] Update docs

### DIFF
--- a/.github/workflows/test_suite_ubuntu_cpu_intel.yml
+++ b/.github/workflows/test_suite_ubuntu_cpu_intel.yml
@@ -90,7 +90,7 @@ jobs:
           . ftorch/bin/activate
           git clone https://gitlab.com/everythingfunctional/veggies.git
           cd veggies
-          fpm build
+          fpm build --compiler ${{ matrix.FC }}
 
       - name: Build and install FTorch
         env:


.github/workflows/test_suite_ubuntu_cpu_gnu.yml
```diff
@@ -161,12 +165,12 @@ jobs:
             -DCMAKE_CXX_COMPILER=g++  \
             -DCMAKE_INSTALL_PREFIX="${BUILD_DIR}" \
             -DCMAKE_BUILD_TESTS=TRUE \
-            -DCMAKE_PREFIX_PATH="${PFUNIT_DIR};${Torch_DIR}" \
-            -DCMAKE_Fortran_FLAGS="-std=${FORTRAN_STANDARD}"
-          cmake --build .
+            -DCMAKE_PREFIX_PATH="${Torch_DIR}" \
+            -DCMAKE_Fortran_FLAGS="-std=${FORTRAN_STANDARD} ${VEGGIES_FLAGS}"
```
joewallwork, line 169: I wrote a `Findveggies.cmake` file (based off the `test-drive` one in #482) but unfortunately it doesn't find my fpm `veggies` build and resorts to fetching from GitLab. When it does that, it tries to build veggies and its dependencies with CMake, which is not the preferred approach. When it does this, the build fails because the forgex (a dependency) build is incompatible with FTorch being a shared library.

A workaround is to pass the include and link files for the veggies build directly to CMake but unfortunately fpm uses a rather odd build directory structure including hashed subdirectories so I just add include and link flags for all of them even though some won't be needed.

Hopefully the workaround can be avoided by fixing the `Findveggies.cmake` file.

.github/workflows/test_suite_ubuntu_cpu_intel.yml
```diff
@@ -80,18 +80,17 @@ jobs:
...
-          make -j 4 install
+          . ftorch/bin/activate
+          pip install fpm
+
+      - name: Install veggies
+        run: |
+          . ftorch/bin/activate
+          git clone https://gitlab.com/everythingfunctional/veggies.git
+          cd veggies
+          fpm build --compiler ${{ matrix.FC }}
```
joewallwork, line 93: The veggies build appears to fail under both ifort and ifx.
https://github.com/Cambridge-ICCS/FTorch/actions/runs/21073137473

PR #486, opened 2026-01-21
Test out differentiating through a model

### LINKED ISSUE #213: Demonstrate differentiating through model output
Towards #111.

We should further test #152 by calling `tensor.backward` on a tensor outputted from a network. By doing so, we'll check that we're able to back-propagate through a network from code that depends on it.

### LINKED ISSUE #483: Could the manually defined function or similiary autograd operation be used with FTorch?
First, for example, I have defined a function to compute the automatic differentiation with python file:
```
    def compute_tangent_stiffness(self, last_sample_strain):
        pred_stress_norm = self.forward(last_sample_strain)
        last_step_stress_norm = pred_stress_norm[0, -1, :]
        last_tangent_stiffness = torch.zeros(self.output_dim, self.input_dim).to(last_sample_strain.device)
        with torch.enable_grad():
            for stress_idx in range(self.output_dim):
                grad = torch.autograd.grad(
                    outputs=last_step_stress_norm[stress_idx],
                    inputs=last_sample_strain,
                    retain_graph=True if stress_idx < self.output_dim-1 else False,
                    create_graph=False
                )[0]
                last_tangent_stiffness[stress_idx, :] = grad[0, -1, :]
        return last_step_stress_norm, last_tangent_stiffness
```
**In short words, the above function is aimed to calculate the automatic differentiation between the output stress (tensor form: N multiplies with 6, N is the loading step) and the input strain (same tensor form).  
I want to realize similar function with FTorch in Fortran.**
However, I have searched the user manual of FTorch, and only the torch_model_forward could be used with ts model, so the user defined function could only be realized by manually code?
I have seen the work example 8) Autograd, but I think the example could not fulfill my goal. 
Thanks for any advice.

Image

joewallwork:
Closes #483.
Closes #213.

This PR demonstrates that it's already possible to differentiate through calls to `torch_model_forward` in FTorch.

## Checklist

- [x] Python example
- [x] Fortran example
- [x] Update docs
- [x] Changelog


examples/9_Autograd/README.md
```diff
@@ -55,10 +62,36 @@ cmake .. -DCMAKE_PREFIX_PATH=<path/to/your/installation/of/library/> -DCMAKE_BUI
...
+```
+y = tensor([[ 2.,  4.,  6.,  8., 10.]], grad_fn=<MmBackward0>)
+dy/dx = tensor([[2., 2., 2., 2., 2.]])
+```
+
+To run the Fortran version, simply execute
+```
+./simplenet
 ```
+This should
```
jatkinson1000: More to write here?

examples/09_Autograd/simplenet_backward.f90
jatkinson1000, line 1: Could this be renamed `simplenet_backpropogation.f90` or similar to make intent clear?

examples/9_Autograd/README.md
joewallwork: Oops! Done in 595ab5b3.

examples/09_Autograd/simplenet_backward.f90
joewallwork: Done in c4f41cf2. Also renamed as `tensor_manipulation_backward` for consistency with the other example.

PR #488, opened 2026-01-21
Subroutine for printing model layer weights

### LINKED ISSUE #420: Subroutine for printing model layer weights
### Discussed in https://github.com/Cambridge-ICCS/FTorch/discussions/419

<div type='discussions-op-text'>

<sup>Originally posted by **rgknox** August  5, 2025</sup>
Hello.  I've generated a trained neural network model in pytorch, and loaded it with Ftorch with the intention of performing inference. But I'd like to verify that I am writing and loading what I think I am, by printing the weights on both sides of the process.  In my case it is just a couple of small linear layers and a relu.

In pytorch, I can loop through the items in the state_dict, like so:

```
    state_dict = model.state_dict()

    for name, tensor in state_dict.items():
        print(f"Parameter: {name}")
        print(tensor)
```

Are there wrapper functions on the FTorch side that can help with this? I see there is a wrapper for printing out the contents of a [tensor](https://github.com/Cambridge-ICCS/FTorch/blob/main/src/ftorch.F90#L2212), but I'm not sure how to loop through the contents of the model and access those tensors.  Thank you for your help!
</div>

joewallwork:
Closes #420.

This PR adds a new subroutine `torch_model_print_parameters`, which prints out all of the parameters (e.g., weights) associated with a Torch model.

The new subroutine is demonstrated in the SimpleNet example and the print statements in the other SimpleNet files are updated for clarity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ For specific details see the [FTorch online documentation](https://cambridge-icc
 
 - Add subroutine for printing parameters associated with a Torch Model
   [#488](https://github.com/Cambridge-ICCS/FTorch/pull/488)
+- Add unit testing for the Torch Model API [#496](https://github.com/Cambridge-ICCS/FTorch/pull/496)
 - Provide support for using FTorch with pkg-config. [#464](https://github.com/Cambridge-ICCS/FTorch/pull/464)
 - Support building FTorch as a static library. [#448](https://github.com/Cambridge-ICCS/FTorch/pull/448)
 - Intel-ifx and Intel-ifort CI and GCC v9-13 CI. Intel CI builds OpenMPI from source to accomodate MPI integration tests [#438](https://github.com/Cambridge-ICCS/FTorch/pull/438)


src/ctorch.cpp
```diff
@@ -616,6 +616,13 @@ void torch_jit_module_forward(const torch_jit_script_module_t module,
   }
 }
 
+void torch_module_print_parameters(const torch_jit_script_module_t module) {
+  auto m = reinterpret_cast<torch::jit::script::Module *>(module);
+  for (const auto &[key, value] : m->named_parameters()) {
+    std::cout << key << ":\n" << value << std::endl;
+  }
+}
+
```
jatkinson1000: Is there any error handling or exception catching that would be useful here?

examples/2_SimpleNet/simplenet_infer_python.py
jatkinson1000, line 1: Could this also print the model weights for consistency with the Fortran?

src/ftorch.fypp
jatkinson1000, line 1: I wonder if it's worth making this also a bound OO method of the model, similar to how we did with some tensors?

test/unit/unittest_tensor_operator_overloads_autograd.pf
```diff
@@ -8,8 +8,7 @@ module unittest_tensor_operator_overloads_autograd
   use funit
   use ftorch, only: assignment(=), ftorch_int, torch_kCPU, torch_kFloat32, &
                     torch_tensor, torch_tensor_backward, torch_tensor_delete, torch_tensor_empty, &
-                    torch_tensor_from_array, torch_tensor_ones, torch_tensor_get_gradient, &
-                    torch_tensor_print
+                    torch_tensor_from_array, torch_tensor_ones, torch_tensor_get_gradient
```
jatkinson1000, line 11: I'm guessing this was an oversight left in from previous?

joewallwork: Yeah I just spotted it when grepping for `_print`.

examples/2_SimpleNet/simplenet_infer_python.py
joewallwork: Done in dd39d06.

src/ftorch.fypp
joewallwork: Sure. Done in f7d2372, also for `torch_tensor_print`.

src/ctorch.cpp
joewallwork: Nothing I can think of beyond checking that the reinterpreted cast of the module is valid, but we don't do that anywhere else. I think we can rely on libtorch's testing for calling `named_parameters` doing the right thing.

src/ftorch.fypp
```diff
@@ -1239,9 +1243,47 @@ contains
                                    logical(requires_grad_value, c_bool))
   end subroutine torch_model_forward
 
+  !| Prints the parameters associated with a model
+  !  NOTE: While viewing parameters in this way can be helpful for small toy models, it will produce
+  !        large amounts of output for models with many, large, or high-dimensional parameters. In
+  !        particular, tensors of 3 or more dimensions will be represented in terms of 2D arrays.
+  subroutine torch_model_print_parameters(self)
```
jatkinson1000, line 1250: Consider `torch_model_print` instead to match tensor call and pola?

src/ctorch.cpp
```diff
@@ -616,6 +616,18 @@ void torch_jit_module_forward(const torch_jit_script_module_t module,
   }
 }
 
+void torch_module_print_parameters(const torch_jit_script_module_t module) {
```
jatkinson1000: Do we need to be specific that this is a jit module as we do in the other function calls?
I'm not sure, but we do explicitly cast to a `jit::script::module` as opposed to a torch module.

```suggestion
void torch_jit_module_print_parameters(const torch_jit_script_module_t module) {
```

jatkinson1000: Same for is_training function

CHANGELOG.md
jatkinson1000, line 1: Consider adding that unit tests for models are also added.

jatkinson1000 (commented):
Hi @joewallwork thanks for this.

A couple of queries for you to consider, after which I'd be happy to approve.

src/ftorch.fypp
joewallwork: But it's not really all the information associated with the model, right? It prints weights etc but doesn't print out what types of layers they are, for example.

src/ctorch.cpp
joewallwork: Done in 64a428d.

CHANGELOG.md
joewallwork: Done in 36d178f.

src/ftorch.fypp
jatkinson1000: Yep, fair enough.

PR #491, opened 2026-01-22
Bump minimum CMake version to 3.18

### LINKED ISSUE #234: Bump the minimum CMake version to 3.18, as they have in [PyTorch](https://github.com/pytorch/pytorch/blob/main/torch/CMakeLists.txt)?
Do you think it might be worth bumping the minimum CMake version to 3.18, as they have in [PyTorch](https://github.com/pytorch/pytorch/blob/main/torch/CMakeLists.txt)?

I think 3.15 still works in most cases, but [CMake 3.18 was the first version to include support for any compiler](https://cmake.org/cmake/help/latest/prop_tgt/CUDA_STANDARD.html), enabling `CMAKE_CUDA_STANDARD` 17 support (it also includes potentially useful things like [CMAKE_CUDA_ARCHITECTURES](https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_ARCHITECTURES.html)), so it may be safer, although C++17 is already in the FTorch requirements separately.

_Originally posted by @ElliottKasoar in https://github.com/Cambridge-ICCS/FTorch/issues/233#issuecomment-2587865382_

joewallwork:
Closes #234

This is the minimum used by PyTorch anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ For specific details see the [FTorch online documentation](https://cambridge-icc
 
 ### Changed
 
+- Bump the minimum CMake version from 3.15 to 3.18, for consistency with what's
+  used in PyTorch. [#491](https://github.com/Cambridge-ICCS/FTorch/pull/491)
 - Significant overhaul of the online FORD documentation and reduction of content in the README
   in [#459](https://github.com/Cambridge-ICCS/FTorch/pull/459)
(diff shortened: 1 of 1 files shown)


jatkinson1000 (commented):
Looks sensible, can you add to the changelog with a reasoning for bumping please?

jatkinson1000 (approved):
Pre-empted my comment as I typed. LGTM

PR #497, opened 2026-01-26
Add macOS testing workflow using gfortran

jatkinson1000:
Adds a workflow file for macOS

Uses gfortran compiler with system clang and clang++
- Testing f2008 and f2018 - is this really neccessary?

I don't think testing MPS over github CI is possible as the runners are virtualised, so no MPS backend is available...

Also updates installation directory to be ouitside build directory for standalone test following #515

### DIFF
--- a/.github/workflows/test_suite_macos_cpu_clang.yml
+++ b/.github/workflows/test_suite_macos_cpu_clang.yml
@@ -1,0 +1,237 @@
+# Workflow to run the FTorch test suite using jobs for GNU Fortran compiler
+# and system Clang compilers (macOS arm64)
+name: Test suite (macOS CPU)
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on pushes to the "main" branch, i.e., PR merges
+  push:
+    branches: [ "main" ]
+
+  # Triggers the workflow on pushes to open pull requests with code changes
+  pull_request:
+    paths:
+      # This workflow
+      - '.github/workflows/test_suite_macos_cpu_clang.yml'
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Cancel jobs running if new commits are pushed
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+# Write permissions are not required
+permissions: {}
+
+# Workflow run - one or more jobs that can run sequentially or in parallel
+jobs:
+  # Dynamically build matrix for job
+  setup-matrix:
+    name: setup matrix
+    runs-on: macos-latest
+
+    # Terminate the job if it runs for more than 5 minutes
+    timeout-minutes: 5
+
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    env:
+      EVENT_NAME: ${{ github.event_name }}
+    steps:
+      - id: set-matrix
+        run: |
+          MATRIX_JSON='
+          {
+            "std": ["f2008", "f2018"],
+            "compiler": ["clang"]
+          }'
+          # Convert json to compact line expected by github action
+          MATRIX_JSON=$(echo "$MATRIX_JSON" | jq -c .)
+          echo "matrix=${MATRIX_JSON}" >> $GITHUB_OUTPUT
+
+  test-suite:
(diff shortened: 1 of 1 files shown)


jatkinson1000:
Following this we will need to update the Mac OS install instructions.

We could explore using a complete gcc toolchain in future with the CMake flag to allow cross-compilation, as is the case on Windows, but this will be more cumbersome (need to install OpenMPI from source etc.) so perhaps not worth it.

.github/workflows/test_suite_macos_cpu_clang.yml
```diff
@@ -0,0 +1,237 @@
...
+on:
+  # Triggers the workflow on pushes to the "main" branch, i.e., PR merges
+  push:
+    branches: [ "main" ]
+
+  # Triggers the workflow on pushes to open pull requests with code changes
+  pull_request:
+    paths:
+      # This workflow
+      - '.github/workflows/test_suite_macos_cpu_gnu.yml'
```
joewallwork: Filename needs updating

.github/workflows/test_suite_macos_cpu_clang.yml
```diff
@@ -0,0 +1,237 @@
+# Workflow to run the FTorch test suite using jobs for GNU Fortran compiler
+# and system Clang compilers (macOS arm64)
+name: Test suite (macOS CPU GNU)
```
joewallwork: Confusion between gnu and clang

.github/workflows/test_suite_macos_cpu_clang.yml
```diff
@@ -0,0 +1,237 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+# Write permissions are not required
+permissions: {}
+
+# Workflow run - one or more jobs that can run sequentially or in parallel
+jobs:
+  # Dynamically build matrix for GNU job
```
joewallwork: Is GNU correct?

.github/workflows/test_suite_macos_cpu_clang.yml
```diff
@@ -0,0 +1,237 @@
...
+            -DCMAKE_C_COMPILER=$CC  \
+            -DCMAKE_CXX_COMPILER=$CXX  \
+            -DCMAKE_Fortran_COMPILER=$FC\
+            -DCMAKE_INSTALL_PREFIX="${FTORCH_INSTALL_DIR}" \
+            -DCMAKE_BUILD_TESTS=TRUE \
+            -DCMAKE_PREFIX_PATH="${PFUNIT_DIR};${Torch_DIR}" \
+            -DCMAKE_Fortran_FLAGS="-std=${FORTRAN_STANDARD}"\
+            -DCMAKE_C_FLAGS="-Xpreprocessor -fopenmp -I$(brew --prefix libomp)/include" \
+            -DCMAKE_CXX_FLAGS="-stdlib=libc++ -Xpreprocessor -fopenmp -I$(brew --prefix libomp)/include" \
+            -DCMAKE_EXE_LINKER_FLAGS="-L$(brew --prefix libomp)/lib -lomp"
```
joewallwork, line 153: Yeah, as mentioned in your comment, we should clarify these flags in the docs in a follow-up PR.

.github/workflows/test_suite_macos_cpu_clang.yml
jatkinson1000: Done in https://github.com/Cambridge-ICCS/FTorch/pull/497/commits/ba255dad94cb6c3782cfb502ed4e243f3712fd93

.github/workflows/test_suite_macos_cpu_clang.yml
jatkinson1000: Made it just `macOS CPU` in https://github.com/Cambridge-ICCS/FTorch/pull/497/commits/ba255dad94cb6c3782cfb502ed4e243f3712fd93

.github/workflows/test_suite_macos_cpu_clang.yml
jatkinson1000: Removed in https://github.com/Cambridge-ICCS/FTorch/pull/497/commits/ba255dad94cb6c3782cfb502ed4e243f3712fd93

PR #496, opened 2026-01-22
Unit tests for model API

joewallwork:
Closes #220.
Merges into #488.

In setting up testing for `torch_model_delete`, I realised there were a few improvements that could be made, consistently with what we did with `torch_tensor_delete`.

## Tasks

- [x] Torch model load
  - [x] Check model `is_training` set correctly
  - [-] Check model `requires_grad` set correctly
  - [-] Check model `device_type` set correctly
  - [-] Check model `device_index` set correctly
- [x] Torch model forward
  - [x] Propagation of `requires_grad` property 
- [x] Torch model print parameters
- [x] Torch model delete

Issue #343, opened 2025-03-28
How to properly make batch inference?


IsidreMas:
Hello FTorch developers, 

I'm a new user of the library trying to include it in an atmospheric chemistry emulator to interface with ML models. So far I managed to reproduce the inference results running the model 1 sample at a time, but when attempting to do batch inference the results are not numerically correct. Do you know what I could be doing wrong?

Here's the example code I'm using to load a dataset from a netcdf file and write the predicitons of the model to a file:

<details><summary>batchintest.F90</summary>

``` Fortran
program read_toy
    use, intrinsic :: iso_fortran_env, only: sp => real32, int64, dp => real64
    use netcdf
    ! IMPORT TORCH INTERFACE MODULES
    use ftorch, only: torch_model, torch_tensor, &
         torch_kCPU, torch_kCUDA, torch_kXPU, torch_kMPS, &
         torch_tensor_from_array, torch_model_load, torch_model_forward, &
         torch_delete
    implicit none

    ! SET PRECISION
    integer, parameter :: wp = sp

    integer :: ncid, status, varid, i, num_samples
    real(wp), allocatable :: gas_conc(:,:,:) ! 3D array to store gas concentrations

    ! COMMAND-LINE ARGUMENTS
    integer :: num_args, ix, device_index
    character(len=128), dimension(:), allocatable :: args

    ! USER CONFIGURATION: CHANGE INPUT/OUTPUT DIMENSIONS AS NEEDED.
    integer, parameter :: n_features = 3      ! (E.G., FOR CALIFORNIA HOUSING, THIS IS 8)
    integer, parameter :: n_outputs  = 3        ! (CHANGE IF YOUR MODEL RETURNS MULTIPLE VALUES)

    ! INPUT DATA ARRAY (SAMPLE): MODIFY THESE VALUES FOR YOUR USE-CASE.
    real(wp), dimension(:,:), target, allocatable :: in_data
    ! OUTPUT DATA ARRAY (INITIALLY ZEROED)
    real(wp), dimension(:,:), target, allocatable :: out_data

    ! DEFINE TENSOR LAYOUT: BATCH SIZE = 1 (SINGLE SAMPLE).
    integer, parameter :: tensor_layout(2) = [1,2]

    ! TORCH OBJECTS
    type(torch_model) :: model
    type(torch_tensor), dimension(:), allocatable :: in_tensors
    type(torch_tensor), dimension(:), allocatable :: out_tensors

[...]

PR #489, opened 2026-01-21
Docs on unit testing

joewallwork:
Closes #250.

## Checklist

- [x] Consistent naming of unit tests.
- [x] Tidying of existing tests, e.g. using pFUnit `@assert` features with tolerances etc.
- [x] Docs on the different unit test files, and how they are separated
- [x] Basic guide for contributing unit tests
- [x] pFUnit primer
- [x] Changelog entry

PR #500, opened 2026-01-28
Add Batching examples and associated documentation

jatkinson1000:
Adds examples for demonstrating batching and higher-dimensional inference in PyTorch and Fortran.

Introduces `BatchingNet` which multiplies each input feature by a distinct value, making the effect of batching more obvious than `SimpleNet`.

Currently labelled as example n, but we shoudl discuss where we want to slot this in in the "learning workflow".
Also worth deciding if we want to add this to integration testing (I think that makes sense) and if there are any unit tests, though perhaps this is for #496.

Closes: #157, #343

To do:

- [x] Add Python example of batching
- [x] Add equivalent FTorch example of batching
- [x] Add documentation pages around batching
- [x] Settle on example number and update accordingly
- [-] ~Add unit tests?~ To be handled in a separate issue/PR
- [x] Outfit for integration testing?
- [x] Update Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-All [notable](https://cambridge-iccs.github.io/FTorch/page/developer.html#versioning-and-changelog)
+All [notable](https://cambridge-iccs.github.io/FTorch/page/developer/developer.html#versioning-and-changelog)
 changes to the project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
@@ -11,6 +11,8 @@ For specific details see the [FTorch online documentation](https://cambridge-icc
 
 ### Added
 
+- Batching in FTorch documented, including a new worked example, in
+  [#500](https://github.com/Cambridge-ICCS/FTorch/pull/500).
 - Add subroutine for printing parameters associated with a Torch Model
   [#488](https://github.com/Cambridge-ICCS/FTorch/pull/488)
 - Add unit testing for the Torch Model API [#496](https://github.com/Cambridge-ICCS/FTorch/pull/496)
@@ -42,6 +44,9 @@ For specific details see the [FTorch online documentation](https://cambridge-icc
 
 ### Changed
 
+- Example numbers were bumped to account for new worked examples in
+  [#291](https://github.com/Cambridge-ICCS/FTorch/pull/291) and
+  [#500](https://github.com/Cambridge-ICCS/FTorch/pull/500).
 - Bump the minimum CMake version from 3.15 to 3.18, for consistency with what's
   used in PyTorch. [#491](https://github.com/Cambridge-ICCS/FTorch/pull/491)
 - Significant overhaul of the online FORD documentation and reduction of content in the README
@@ -54,8 +59,6 @@ For specific details see the [FTorch online documentation](https://cambridge-icc
 - scalar multiplication/division of tensors reworked to require the scalar to first be mapped to a `torch_tensor` in [#289](https://github.com/Cambridge-ICCS/FTorch/pull/289)
 - The unit tests for constructing and destroying tensors were separated out in
   [#319](https://github.com/Cambridge-ICCS/FTorch/pull/319)
-- Demo numbers were bumped to account for new demo in
-  [#291](https://github.com/Cambridge-ICCS/FTorch/pull/291).
 - Use interface for `torch_tensor_from_array` with default layout in tests and
   examples in [#348](https://github.com/Cambridge-ICCS/FTorch/pull/348).
 - Error handling in `ctorch.cpp` improved in [#347](https://github.com/Cambridge-ICCS/FTorch/pull/347).


examples/4_Batching/batching_demo.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    print("\n--- Error Case: Wrong Feature Dimension ---")
+    # Input tensor with wrong last dimension (should be 5)
+    input_tensor = torch.ones(5, 2)
+    print(f"Input shape: {input_tensor.shape}")
+    try:
+        output = model(input_tensor)
+        print(f"Output:\n{output}")
+    except RuntimeError as e:
+        print(f"Error occurred: {e}")
```
joewallwork, line 109: This looks a bit more like a unit test but I can see why you include it here.

We should keep it in but I think the explanatory docstring/comment could be a bit clearer. They should convey *why* the last dimension should be 5, otherwise it seems a bit like a magic number.

examples/n_batching/batching_demo.py
```diff
@@ -0,0 +1,121 @@
...
+    model = BatchingNet()
+    model.eval()
+    run_unbatched(model)
+    run_simple_batch(model)
+    run_higher_dim_batch(model)
+    run_shape_error_case(model)
+
+
+if __name__ == "__main__":
+    main()
```
joewallwork: Purely out of interest, how come you don't inline `main`? Is this considered better practice nowadays?

examples/4_Batching/README.md
```diff
@@ -0,0 +1,213 @@
...
+each input feature by a different value (0, 1, 2, 3, 4). The demo script
+`batching_demo.py` shows how to use this model for unbatched, batched, and
+higher-dimensional inference, illustrating the effect of batching and the location of
+the batch dimension. All outputs are as described in the "Why Batching?" section below.
+
+
+## Dependencies
+To run this example requires:
+
+- CMake
```
joewallwork, line 19: You do also provide a `Makefile` so not necessarily?

examples/n_batching/README.md
```diff
@@ -0,0 +1,213 @@
...
+
+
+## Key Considerations
+- **Shape:** The last dimension of the input must match the model's expected input
+  size (here, 5) otherwise it will result in an error.
+- **Location of Batching Dimension:** PyTorch modules such as `nn.Linear` always operate
+  on the last dimension of the input tensor. Any number of leading (batching) dimensions
+  are supported, and the operation is applied independently across all of them.
+  For example, `[batch, 5]`, `[batch, time, 5]`, or even `[batch1, batch2, ..., 5]` are
+  all valid as long as the final dimension is 5.
```
joewallwork:
```suggestion
  all valid as long as the final dimension is consistent with the input size.
```

examples/4_Batching/README.md
```diff
@@ -0,0 +1,213 @@
...
+## Key Considerations
+- **Shape:** The last dimension of the input must match the model's expected input
+  size (here, 5) otherwise it will result in an error.
+- **Location of Batching Dimension:** PyTorch modules such as `nn.Linear` always operate
+  on the last dimension of the input tensor. Any number of leading (batching) dimensions
+  are supported, and the operation is applied independently across all of them.
+  For example, `[batch, 5]`, `[batch, time, 5]`, or even `[batch1, batch2, ..., 5]` are
+  all valid as long as the final dimension is 5.
+- **Contiguity:** If you slice or permute tensors, ensure they are contiguous in memory
+  before passing to the model (use `.contiguous()` if needed).
```
joewallwork, line 128: This feels relevant to #417

examples/n_batching/README.md
```diff
@@ -0,0 +1,213 @@
...
+--- Multidimensional batched input/output: ---
+Input (1,1):    1.00000000       1.00000000       1.00000000       1.00000000       1.00000000
+Output (1,1):   0.00000000       1.00000000       2.00000000       3.00000000       4.00000000
+Input (2,3):    30.0000000       30.0000000       30.0000000       30.0000000       30.0000000
+Output (2,3):   0.00000000       30.0000000       60.0000000       90.0000000       120.000000
+
+BatchingNet Fortran example ran successfully
+```
+
+A commented-out line () in the Fortran code demonstrates how an incorrect feature dimension (e.g., `[2, 4]`) will cause a runtime error.
```
joewallwork: Is something meant to be in the brackets?

examples/n_batching/README.md
```diff
@@ -0,0 +1,213 @@
...
+Uncomment, rebuild and re-run to observe the error message for incorrect batching
+layout.
+
+You may also need to add the location of the dynamic library files (`.so` or `.dylib`) that FTorch links against at runtime to your `LD_LIBRARY_PATH`:
+```
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:</path/to/library/installation>/lib
+```
+or `DYLD_LIBRARY_PATH` on mac:
```
joewallwork:
```suggestion
or `DYLD_LIBRARY_PATH` on Mac:
```

examples/n_batching/README.md
```diff
@@ -0,0 +1,213 @@
...
+```
+export DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH:</path/to/library/installation>/lib
+```
+
+The Fortran code uses `assert_allclose` from `ftorch_test_utils` to check outputs match expectations.
```
joewallwork: Note this will get a deprecation warning once we merge #489

examples/n_batching/batchingnet_infer_fortran.f90
```diff
@@ -0,0 +1,144 @@
+program batchingnet_infer
+
+   use, intrinsic :: iso_fortran_env, only : sp => real32
+   use ftorch, only : torch_model, torch_tensor, torch_kCPU, torch_delete, &
+                      torch_tensor_from_array, torch_model_load, torch_model_forward
+   use ftorch_test_utils, only : assert_allclose
```
joewallwork: Again, note that this will become `allclose`.

examples/n_batching/batchingnet_infer_fortran.f90
```diff
@@ -0,0 +1,144 @@
...
+   ! Cleanup
+   call torch_delete(model)
+   call torch_delete(in_tensors_single)
+   call torch_delete(out_tensors_single)
+   call torch_delete(in_tensors_batch)
+   call torch_delete(out_tensors_batch)
+   call torch_delete(in_tensors_multi)
+   call torch_delete(out_tensors_multi)
+   call torch_delete(in_tensors_bad)
+   call torch_delete(out_tensors_bad)
```
joewallwork: Perhaps move these up so that the tensors are deleted at the end of each block when they're no longer needed?

examples/n_batching/README.md
jatkinson1000: Sigh. I think I'll leave this as working for now, and then do a rebase.
Thanks for flagging.

examples/4_Batching/README.md
jatkinson1000: Indeed, and also to some extent #423 and Peter's query

examples/4_Batching/batching_demo.py
jatkinson1000: See https://github.com/Cambridge-ICCS/FTorch/pull/500/commits/3bf76814085132b3c4a6bf0857bb5e1f78332da4 ?

examples/n_batching/batching_demo.py
jatkinson1000: Habit I think, but you are right that the other examples have it inlined.
See https://github.com/Cambridge-ICCS/FTorch/pull/500/commits/3bf76814085132b3c4a6bf0857bb5e1f78332da4

examples/4_Batching/README.md
jatkinson1000: I copied this format over from the other examples, so can update this more broadly across examples if you want?

examples/4_Batching/batching_demo.py
joewallwork: Yeah that works, thanks.

examples/n_batching/README.md
jatkinson1000: See https://github.com/Cambridge-ICCS/FTorch/pull/500/commits/a2a6d9d3af6783be4aee881701740ae64d468429

examples/4_Batching/README.md
joewallwork: Ah okay I hadn't realised it was the same in others. Feel free to either do that here or postpone for a follow-up PR.

examples/n_batching/README.md
jatkinson1000: Yes, the line number. Realised this is bad for maintenance so made it the call sign instead, and moved to be an "Extension" in https://github.com/Cambridge-ICCS/FTorch/pull/500/changes/a29b1387ad95bd8730e0c8e2dfe408e1a20311a2

examples/n_batching/README.md
```diff
@@ -0,0 +1,215 @@
...
+## Further Exploration
+
+- incorrect Dimension order
+    - A commented-out line in the Fortran code demonstrates how an incorrect feature
+      dimension (e.g., `[2, 4]`) will cause a runtime error.
+      ```fortran
+      call torch_model_forward(model, in_tensors_bad, out_tensors_bad)
+      ```
+    - Uncomment, rebuild and re-run to observe the error message for incorrect
+      batchinglayout.
```
joewallwork:
```suggestion
      batching layout.
```

examples/n_batching/README.md
jatkinson1000: Realised we don't add this advice in any exercises after Simplenet, so removed in https://github.com/Cambridge-ICCS/FTorch/pull/500/commits/9a573ac90041d680b5c3336ccde674197d22b442

examples/n_batching/batchingnet_infer_fortran.f90
jatkinson1000: Done in https://github.com/Cambridge-ICCS/FTorch/pull/500/changes/a29b1387ad95bd8730e0c8e2dfe408e1a20311a2

examples/4_Batching/README.md
jatkinson1000: See https://github.com/Cambridge-ICCS/FTorch/pull/505

examples/5_MultiIO/multiionet.py
joewallwork, line 1: This is a nicer implementation than my original one, thanks.

examples/5_MultiIO/README.md
```diff
@@ -97,3 +97,24 @@ This runs the model with the same inputs as above and should produce the output:
    0.00000000       2.00000000       4.00000000       6.00000000    
    0.00000000      -3.00000000      -6.00000000      -9.00000000
 ```
+
+---
+
+## Extension: Batching Support and Architectural Warnings
+
+As an extension exercise try adding implementing batching.
```
joewallwork: Refer/link back to the other example?

jatkinson1000:
@joewallwork just rebased on main to resolve conflicts.

I think this is now ready for a review with docs etc. in place.

Only thing left is to perhaps add a batching unit test to the model tests - I'm thinking something that just passes multidimensional data through SimpleNet to check it goes through OK, though I could implement a version of BatchingNet as SimpleNet doesn't actually explicitly demonstrate batching (everything gets multiplied by 2).

Alternatively I could add both BatchingNet and MultiIONet as model unit tests in a separate PR.

examples/5_MultiIO/README.md
jatkinson1000: See https://github.com/Cambridge-ICCS/FTorch/pull/500/commits/8cf7a22fc395b98679f4c881368a6e9a38b90fa7

PR #521, opened 2026-02-06
Expunge example numbers from CMake Testing

### LINKED ISSUE #506: Remove exercise numbering everywhere aside from the directories
In #500 it has become a real pain to insert a new example and increment the numbering of the rest.

I propose removing any references to numbers from the README, CMakeLists, etc and leaving them only in the directory names and web documentation.
This would allow future incrementation with a simple `git mv` and no need to edit isolated lines of source in many locations.

The test names for ctest should be inccremented to be instructive e.g. `simplenet_pt2ts` rather than `example2_pt2ts`.

jatkinson1000:
Closes #506 

Removes numbers from CMake test names to make insertion of exercises easier in future.

Adopts a new CTest naming convention:
`example_<test_name>_<<specifier>>`
example is to keep the regex matching for running integration tests
test_name is to identify which sub-example is being run
specifier is optional for examples with multiple stages.
e.g. `example_mpi_simplenet`

Running tests should work as before

We have chosen to leave the directories numbered to guide users through in a set order as we feel is appropriate.
We also leave the the titles in READMEs as numbered to match, and in any reference lists in the examples README or online docs. These are (relatively) easily bumped, the main pain point was the CMake.

TODO:

- [x] Check anywhere in docs that needs updating

### DIFF
--- a/pages/developer/testing.md
+++ b/pages/developer/testing.md
@@ -90,8 +90,9 @@ Integration tests may be executed in the following ways:
 1. To run just the integration tests, use `ctest -R example`.<br>
    (`ctest -R` accepts a regular expression to search for and this works because
    all integration tests start with `example`.)
-2. To run a specific integration test use `ctest -R example2`, for example.<br>
-   Alternatively navigate to the corresponding example in `${BUILD_DIR}/examples`
+2. To run a specific integration test, use its full name, e.g.,
+   `ctest -R example_simplenet` or `ctest -R example_resnet`.<br>
+   Alternatively, navigate to the corresponding example in `${BUILD_DIR}/examples`
    and call `ctest`.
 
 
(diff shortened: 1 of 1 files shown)


joewallwork (changes requested):
I agree that this is the right way to do this.

It looks like the CMake formatter made some slightly odd indentation changes in all the files? `add_test` gets moved back by one tab.

pages/developer/testing.md
```diff
@@ -90,8 +90,9 @@ Integration tests may be executed in the following ways:
 1. To run just the integration tests, use `ctest -R example`.<br>
    (`ctest -R` accepts a regular expression to search for and this works because
    all integration tests start with `example`.)
-2. To run a specific integration test use `ctest -R example2`, for example.<br>
-   Alternatively navigate to the corresponding example in `${BUILD_DIR}/examples`
+2. To run a specific integration test, use its full name, e.g.,
+   `ctest -R example_resnet_resnet18` or `ctest -R example_simplenet_pt2ts`.<br>
```
joewallwork: How about mentioning a specific example? e.g., `example_resnet`? Just because some example components won't work in isolation. e.g., `simplenet_infer_fortran` won't work unless the model was already created and written to TorchScript.

jatkinson1000: Good call. Given none are particularly time consuming I updated to just use the main example names.
Any CTest pros can run single examples if desired.

PR #522, opened 2026-02-06
Provide information about tensor finalizer

### LINKED ISSUE #474: Investigate warning about finalization
In some of the build logs I see
```
Warning: The structure constructor at (1) has been finalized. This feature was removed by f08/0011. Use -std=f2018 or -std=gnu to eliminate the finalization.
```
We should look into this.

joewallwork:
Closes #474.

After some experimentation (see the issue), I established that we do not need to worry about the finalizer warning.

In this PR, I created a "common warnings" section with some information, as well as adding more information in the Tensor API docs.

### DIFF
--- a/pages/usage/tensor.md
+++ b/pages/usage/tensor.md
@@ -14,72 +14,75 @@ date: Last Updated: October 2025
 
 ### Overview
 
-FTorch provides a [[ftorch(module):torch_tensor(type)]] derived type, which exposes the
-functionality of the `torch::Tensor` C++ class. The interface is designed to be familiar to
-Fortran programmers, whilst retaining strong similarity with `torch::Tensor` and
-the `torch.Tensor` Python class.
+FTorch provides a [[ftorch_tensor(module):torch_tensor(type)]] derived type,
+which exposes the functionality of the `torch::Tensor` C++ class. The interface
+is designed to be familiar to Fortran programmers, whilst retaining strong
+similarity with `torch::Tensor` and the `torch.Tensor` Python class.
 
-Under the hood, the [[ftorch(module):torch_tensor(type)]] type holds a pointer to a
-`torch::Tensor` object in C++ (implemented using `c_ptr` from the `iso_c_binding`
-intrinsic module). This allows us to avoid unnecessary data copies between C++ and
-Fortran.
+Under the hood, the [[ftorch_tensor(module):torch_tensor(type)]] type holds a
+pointer to a `torch::Tensor` object in C++ (implemented using `c_ptr` from the
+`iso_c_binding` intrinsic module). This allows us to avoid unnecessary data
+copies between C++ and Fortran.
 
 ### Procedures
 
 #### Constructors
 
-We provide several subroutines for constructing [[ftorch(module):torch_tensor(type)]]
-objects. These include:
+We provide several subroutines for constructing
+[[ftorch_tensor(module):torch_tensor(type)]] objects. These include:
 
-* [[ftorch(module):torch_tensor_empty(subroutine)]] which allocates memory for the
-  [[ftorch(module):torch_tensor(type)]] but does not set any values.
-* [[ftorch(module):torch_tensor_zeros(subroutine)]] which creates a
-  [[ftorch(module):torch_tensor(type)]] whose values are uniformly zero.
-* [[ftorch(module):torch_tensor_ones(subroutine)]] which creates a
-  [[ftorch(module):torch_tensor(type)]] whose values are
+* [[ftorch_tensor(module):torch_tensor_empty(subroutine)]] which allocates
+  memory for the [[ftorch_tensor(module):torch_tensor(type)]] but does not set
+  any values.
+* [[ftorch_tensor(module):torch_tensor_zeros(subroutine)]] which creates a
+  [[ftorch_tensor(module):torch_tensor(type)]] whose values are uniformly zero.
+* [[ftorch_tensor(module):torch_tensor_ones(subroutine)]] which creates a
+  [[ftorch_tensor(module):torch_tensor(type)]] whose values are
   uniformly one.
-* [[ftorch(module):torch_tensor_from_array(interface)]] which allows the user to create
-  a [[ftorch(module):torch_tensor(type)]] with the same rank, shape, and data type as a
-  given Fortran array. Note that the data is *not* copied - the tensor data points to the
-  Fortran array, meaning the array must have been declared with the `target` property.
-  The array will continue to be pointed to even when operations are applied to the
-  tensor, so this subroutine can be used 'in advance' to set up an array for
-  outputting data. [[ftorch(module):torch_tensor_from_array(interface)]] may be called
-  with or without the `layout` argument - an array which specifies the order in which
-  indices should be looped over. The default `layout` is `[1,2,...,n]` implies that data
-  will be read into the same indices by Torch. (See the
+* [[ftorch_tensor(module):torch_tensor_from_array(interface)]] which allows the
+  user to create a [[ftorch_tensor(module):torch_tensor(type)]] with the same
+  rank, shape, and data type as a given Fortran array. Note that the data is
+  *not* copied - the tensor data points to the Fortran array, meaning the array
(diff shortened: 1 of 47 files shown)


pages/usage/tensor.md
```diff
@@ -85,10 +85,18 @@ procedures. For example, `tensor%get_rank` can be used in place of
...
+[[ftorch(module):torch_delete(interface)]] is provided such that this can also
+be applied to arrays of tensors.
+
+Manually deallocating a tensor that you declared and constructed in your code is
+actually optional. If the tensor was declared in a subroutine then
+[[ftorch(module):torch_tensor_delete(subroutine)]] will get called as the
+finalizer of [[ftorch(module):torch_tensor(type)]] when it goes out of scope. If
+the tensor was declared in a program then the finalizer won't get called, but
+this is not considered to be an issue, in the same way that in Fortran it
+doesn't matter if allocated arrays aren't deallocated at the end of the program.
```
jatkinson1000, line 102: Could you add a note/link to the common-warnings section relevant to this ("you may see a warning referencing the finalizer when...")?

pages/usage/tensor.md
```diff
@@ -85,10 +85,18 @@ procedures. For example, `tensor%get_rank` can be used in place of
...
-[[ftorch(module):torch_tensor(type)]] object: [[ftorch(module):torch_delete(interface)]].
-An interface is provided such that this can also be applied to arrays of tensors.
-Calling this subroutine manually is optional as it is called as a destructor when a
-[[ftorch(module):torch_tensor(type)]] goes out of scope.
+[[ftorch(module):torch_tensor(type)]] object:
+[[ftorch(module):torch_tensor_delete(subroutine)]]. An interface
+[[ftorch(module):torch_delete(interface)]] is provided such that this can also
+be applied to arrays of tensors.
+
+Manually deallocating a tensor that you declared and constructed in your code is
```
jatkinson1000, line 96: Is there a link we can point people to if they want more information on finalizers in Fortran?

joewallwork: Good idea. Done in de1f1527ac1f7b7879501399557d849ddf5e7e00.

pages/usage/tensor.md
joewallwork: Done in a8704e2e0a495685d625a42aeba79a21c62d6252.

pages/usage/troubleshooting.md
```diff
@@ -95,3 +96,22 @@ of an equals sign, the overloaded assignment operator should be triggered.
...
+### Common warnings
+
+#### Finalizer with Fortran 2008
+
+If you are building FTorch with gfortran and are specifying the Fortran 2008
+standard (e.g., with the compiler flag `-std=f2008` or by default) then you may
+get compiler warnings of the form:
+```
+Warning: The structure constructor at (1) has been finalized. This feature was removed by f08/0011. Use -std=f2018 or -std=gnu to eliminate the finalization.
+```
```
Mikolaj-A-Kowalski, line 112: Should we perhaps include a code example that may trigger it? I am thinking that the warning  is OK in the particular case we observed, but it may not be in other cases. We should provide a context in which the warning should be safely ignored. 

Part of me wants to link the [`f08/0011`](https://wg5-fortran.org/N2001-N2050/N2006.txt) (I know I would appreciate it as a reader ;-) ), but on the other hand I accept that it may be too much for most readers of the doc. I leave it at your discretion.

pages/usage/troubleshooting.md
```diff
@@ -95,3 +96,22 @@ of an equals sign, the overloaded assignment operator should be triggered.
...
+These warn that the finalizer of the [[ftorch(module):torch_tensor(type)]]
+derived type is triggered when a tensor goes out of scope, despite the fact that
+this feature was removed from the 2008 standard. That is, the
+[[ftorch(module):torch_tensor_delete(subroutine)]] subroutine is called
+so that the associated memory is automatically freed. This is the behaviour that
+we want so nothing needs to be done. See the
+[tensor documentation](|page|/usage/tensor.html#deallocation) for more details.
```
Mikolaj-A-Kowalski: I may be misunderstanding something... but this paragraphs reads as if we rely on the finalizer getting triggered after structure constructor.  

In other words we rely on non-conforming behaviour 😬 and we will end up with leaks if we do request standard conformance (`--std=f2008` flag) or we use other complier that is more strict. 

Other option is that I am misunderstanding something about  what is going on and the `gfortran` is calling the finalizer twice, but we don't care because we check if the pointer in null before we try to deallocate it.

Mikolaj-A-Kowalski (approved):
Sorry for the delay. I forgot to do it yesterday and didn't have time on Friday.

Just couple of comments to consider.

pages/usage/troubleshooting.md
jatkinson1000: I would lean towards not here for the sake of not bloating the docs.

pages/usage/troubleshooting.md
jatkinson1000: A fair point - when we say "this feature was removed" it could imply that "finalizers are no longer called from 2008 spec onwards". Perhaps just some clarity about what was removed, and confirmation that the tensor finalizer is still called for later standards.

joewallwork: FYI I am currently investigating the warning to figure out what exactly is going on. I hadn't initially picked up on the fact that it mentions *structure* constructor.

 I discovered that it arises in the library build because of the automatic creation (and deletion) of the `external_gradient` tensor inside `torch_tensor_backward`. If we change to requiring an `external_gradient` be provided then the warning goes away. However, there are still many instances in the tests and examples that I'm investigating.

joewallwork: If I understand correctly, structure constructor refers to the following example usage
```fortran
  subroutine test()
    use, intrinsic :: iso_c_binding, only: c_null_ptr
    implicit none
    type(torch_tensor) :: t
    t = torch_tensor(c_null_ptr)
  end subroutine test
```
This compiles fine but raises the warning.

Mikolaj-A-Kowalski: > If I understand correctly, structure constructor refers to the following example usage
> 
> ```fortran
>   subroutine test()
>     use, intrinsic :: iso_c_binding, only: c_null_ptr
>     implicit none
>     type(torch_tensor) :: t
>     t = torch_tensor(c_null_ptr)
>   end subroutine test
> ```
> 
> This compiles fine but raises the warning.

This is exactly my understanding what the [structure constructor](https://www.ibm.com/docs/en/xl-fortran-linux/16.1.0?topic=types-structure-constructor) is. 

But the weird part is that when I was looking at compile logs the warning seemed to appear in places where no structure constructors were used (e.g. example 2). 

Could it be that the warning is a bit buggy? The location it reports is a bit weird (end of a subprogram).

I have some problem getting the warning to reproduce on [godbolt](https://godbolt.org/z/h7WW1aY1v). It seems a bit tricky to get it to appear. 

(Also scary is that the finalizers in the toy example do not get called at all with gfortran <= 12)  :grimacing:

joewallwork: > But the weird part is that when I was looking at compile logs the warning seemed to appear in places where no structure constructors were used (e.g. example 2).

Structure constructors aren't used anywhere in FTorch, which makes it confusing indeed.
 
> Could it be that the warning is a bit buggy? The location it reports is a bit weird (end of a subprogram).

Possibly.

> (Also scary is that the finalizers in the toy example do not get called at all with gfortran <= 12) 😬

Hm that is concerning indeed.

pages/usage/troubleshooting.md
joewallwork: Added a link in b0bf144f7f00069baf645ec9b661918fa1deba6a.

pages/usage/troubleshooting.md
joewallwork: Okay I made it more specific to the structure constructor in 474110177b725e4c93836e8b84652b38b8be0390. I think a docs update is probably sufficient for now but I will separately look into handling the gfortran <= 12 case.

joewallwork: See #529.

PR #508, opened 2026-02-02
Better modularisation

joewallwork:
Closes #507 

This PR overhauls FTorch to provide better modularisation _without breaking the current API_.

This is achieved by separating the contents of the existing `ftorch` module out into `ftorch_types`, `ftorch_tensor`, and `ftorch_model`, for better separation-of-concerns. These are `use`d in the main `ftorch` module and made public. The same interface for a generic `torch_delete` is maintained.

Notable changes:
* The pre-processing needed to be changed because it's now `ftorch_tensor` that uses `fypp`. This no longer needs to have the `F90` extension because it doesn't use the C preprocessor.
* The `ftorch_types` module does use the C preprocessor so needs the `F90` extension.
* I spotted that `torch_model` is missing a finalizer so added `torch_model_delete`, although noting to follow up in #474.

PR #526, opened 2026-02-10
Use specific modules in unit tests

joewallwork:
Follows on from #508.

Makes use of the new modules in the unit tests. As agreed, the examples are left as-is with `use ftorch`.

### DIFF
--- a/test/unit/model/unittest_model_constructors_destructors.pf
+++ b/test/unit/model/unittest_model_constructors_destructors.pf
@@ -6,7 +6,8 @@
 !    file for details.
 module unittest_model_constructors_destructors
   use funit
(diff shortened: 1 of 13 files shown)


jatkinson1000 (commented):
The code all seems correct, I just wonder what the argument is for using specific modules is beyond "it's generally accepted that this is what to do in software".

Part of me thinks we should try and be consistent across the codebase both how we advise usage externally but also internally.
`change_my_mind.jpg`

jatkinson1000 (approved):
> My view is that if we are going to have modular unit testing then the tests should use the module they are designed to test.

An excellent argument and I agree.
Approved!

PR #528, opened 2026-02-12
Allow for custom external gradients in `torch_tensor_backward`

### LINKED ISSUE #527: Allow non-unit external gradient
Unlike in (Py)Torch, `torch_tensor_backward` currently doesn't accept an external gradient and instead assumes it to be a tensor of appropriate dimension filled with ones. There is a `TODO` note to allow customisation.

During investigation of #474, I found that adding a compulsory `external_gradient` argument avoids a Fortran 2008 compiler warning when building the library. Combined with the current divergence from the (Py)Torch API, this makes a case for switching over.

joewallwork:
Closes #527.

This PR separates `torch_tensor_backward` out into two subroutines and provides a common interface:
* `torch_tensor_backward_with_external_gradient` allows the user to pass a custom external gradient.
* `torch_tensor_backward_without_external_gradient` uses the default external gradient consisting of a tensor of ones.

## Important notes

* This change allows us to avoid the `f08/0011` warning when building the FTorch library, as investigated in #522. This is because we no longer create and destroy an `external_gradient` tensor in Fortran.
* The default external gradient consisting of a tensor of **ones only works for scalar outputs**. This breaks the current API because the tests and autograd example have considered default external gradients of dimension $2\times3$. However, in practice most people will be calling `torch_tensor_backward` on the output of a loss function, so personally I'd be happy to accept this slight change in API.

## Checklist

* [x] Add unit tests for non-unit external gradients.
* [x] Make the situation clear in the docs.
* [x] Check if it's possible to pass the `retain_graph` argument but not an external gradient. (This didn't initially work for me on the C++ side.) See https://github.com/pytorch/pytorch/blob/main/torch/csrc/autograd/autograd.h.

### DIFF
--- a/src/ctorch.cpp
+++ b/src/ctorch.cpp
@@ -510,6 +510,9 @@ void torch_tensor_backward_with_external_gradient(
     validate_tensor(t, "Input tensor");
     validate_tensor(g, "External gradient");
 
+    // Check if the input tensor has the requires_grad property
+    validate_requires_grad(t, "Input tensor");
+
     // Perform backwards step
     t->backward(*g, retain_graph);
   } catch (const std::exception &e) {
@@ -522,8 +525,11 @@ void torch_tensor_backward_without_external_gradient(const torch_tensor_t tensor
   auto t = reinterpret_cast<torch::Tensor *>(tensor);
 
   try {
-    // Check if the tensors are valid and defined
+    // Check if the input tensor is valid and defined
     validate_tensor(t, "Input tensor");
+
+    // Check if the input tensor has the requires_grad property
+    validate_requires_grad(t, "Input tensor");
 
     // Perform backwards step
     t->backward({}, retain_graph);


joewallwork:
>     Check if it's possible to pass the `retain_graph` argument but not an external gradient. (This didn't initially work for me on the C++ side.) See https://github.com/pytorch/pytorch/blob/main/torch/csrc/autograd/autograd.h.

Done in https://github.com/Cambridge-ICCS/FTorch/pull/528/commits/5919560f1a90eedc9e3643383125c2744e22c973.

src/ctorch.cpp
jatkinson1000, line 1: Since the tensor needs to have `requires_grad` enabled is it possible to perform a check to catch this and raise a sensible error, either here or in the Fortran? Unless it already raises a sensible and traceable error from Torch.

examples/9_Autograd/autograd.f90
jatkinson1000, line 1: It it worth noting that external gradient is only needed when you are not using ones? Or just use the basic case in the example?

Otherwise users might leave this thinking they always need to provide an external gradient as the example here is ones.

CHANGELOG.md
```diff
@@ -65,6 +65,11 @@ For specific details see the [FTorch online documentation](https://cambridge-icc
 
 ### Changed
 
+- Modified `torch_tensor_backward` so that it can accept a custom external
+  gradient tensor as its second argument. The current API is broken slightly
+  because the original signature (with assumed unit external gradient) now only
+  works if the first argument is a tensor with a single entry. However, this
```
jatkinson1000: What do you mean as "single entry"?

test/unit/tensor/unittest_tensor_autograd.pf
```diff
@@ -185,46 +188,174 @@ contains
...
     call torch_tensor_from_array(dQda, out_data, device_type)
 
+    ! Create an external gradient filled with ones
+    call torch_tensor_ones(external_gradient, 1, scalar_shape, dtype, device_type)
+
     ! Apply back-propagation and retrieve the gradient and check it takes the expected value:
-    !   Q(a) = a => dQ/da = 1
-    call torch_tensor_backward(Q)
+    !   Q(a) = a => external_gradient * dQ/da = external_gradient * 1 = external_gradient
+    if (this%param%switch) then
```
jatkinson1000, line 225: I would have found this comment useful when reviewing. Intent became clear when viewing the whole file rather than the diff so I'll leave up to you.

```suggestion
    ! Test both provided and assumed external gradient functionality using parameter switch
    if (this%param%switch) then
```

test/unit/tensor/unittest_tensor_autograd.pf
```diff
@@ -185,46 +188,174 @@ contains
...
 
     implicit none
 
     class(TestCaseType), intent(inout) :: this
-    type(torch_tensor) :: Q, a, dQda
-    real(wp), dimension(2,3), target :: in_data, out_data
-    real(wp), dimension(2,3) :: expected
+    type(torch_tensor) :: Q, a, dQda, external_gradient
+    real(wp), dimension(1), target :: in_data, out_data
+    real(wp), dimension(1) :: expected
```
jatkinson1000, line 203: I presume the reduction in dimensions here is because it is not integral to testing `retain_graph` so was unnecessary compute?

test/unit/tensor/unittest_tensor_autograd.pf
```diff
@@ -185,46 +188,174 @@ contains
...
-    call torch_tensor_backward(Q, retain_graph=.true.)
+    call torch_tensor_backward(Q, external_gradient, retain_graph=.true.)
     call torch_tensor_get_gradient(dQda, a)
     @assertTrue(allclose(out_data, expected, test_name="test_retain_graph2"))
 
   end subroutine test_retain_graph
- 
+
+  @test(testParameters={get_parameters_switch()})
+  subroutine test_unit_external_grad_0d(this)
```
jatkinson1000, line 249: It isn't clear to me from the name what this is testing.
I saw external grad in the name, but then it uses the switch to check both external grad and assumed grad...?
Might need a renaming or a subroutine description to make it clear.

test/unit/tensor/unittest_tensor_autograd.pf
```diff
@@ -185,46 +188,174 @@ contains
...
+    !   Q(a) = a => external_gradient * dQ/da = external_gradient * 1 = external_gradient
+    call torch_tensor_backward(Q, external_gradient)
+    call torch_tensor_get_gradient(dQda, a)
+    expected(:) = grad_data
+    @assertTrue(allclose(out_data, expected, test_name="test_torch_tensor_nonunit_external_grad_0d"))
+
+  end subroutine test_nonunit_external_grad_0d
+
+  @test(testParameters={get_parameters_short()})
+  subroutine test_nonunit_external_grad_2d(this)
```
jatkinson1000, line 331: Is there a reason not to have `test_unit_external_grad_2d`?
Because in this case you'd just use the call without the provided gradient?
But why do we need the `0d` case?
I think this relates to my confusion in your changelog message, but exactly what it all means isn't clear to me from this review.

test/unit/tensor/unittest_tensor_operator_overloads_autograd.pf
jatkinson1000, line 1: Is there a reason to move these all to explicitly providing an external gradient?
Should the new implementation not still use ones?
Maybe this relates to my confusion about API changes in the Changelog and is now required?

test/unit/tensor/unittest_tensor_autograd.pf
joewallwork: No, it's because the default external gradient only works in that case.

test/unit/tensor/unittest_tensor_operator_overloads_autograd.pf
joewallwork: Yes. These are not scalar cases and so you now need to provide an external gradient. This is consistent with what's the case in (Py)Torch.

examples/9_Autograd/autograd.f90
joewallwork: Added a note in 31230495.

CHANGELOG.md
joewallwork: Clarified in 4e3853cd.

test/unit/tensor/unittest_tensor_autograd.pf
joewallwork: Applied in 16e87679.

test/unit/tensor/unittest_tensor_autograd.pf
joewallwork: Added in 19aaa78b.

test/unit/tensor/unittest_tensor_autograd.pf
joewallwork: > Is there a reason not to have `test_unit_external_grad_2d`?

There isn't a `test_unit_external_grad_2d`?

> Because in this case you'd just use the call without the provided gradient?

No, that's only possible for scalar-valued (a.k.a. 0-dimensional) tensors.

> But why do we need the `0d` case? I think this relates to my confusion in your changelog message, but exactly what it all means isn't clear to me from this review.


We could drop `test_nonunit_external_grad_0d` in favour of  just having `test_nonunit_external_grad_2d` but I thought it'd be worth having another case beyond the somewhat special 0d case.

Also added some descriptions in 42ce1d89.

test/unit/tensor/unittest_tensor_autograd.pf
jatkinson1000: Thanks, the new changes have clarified that now.

src/ctorch.cpp
joewallwork: Added in 8656e93f.

Unless you feel this is okay as an error message?
```
8: .[ERROR]: element 0 of tensors does not require grad and does not have a grad_fn
8: Exception raised from run_backward at /pytorch/torch/csrc/autograd/autograd.cpp:104 (most recent call first):
8: frame #0: c10::Error::Error(c10::SourceLocation, std::string) + 0x96 (0x75a18ab83056 in /home/joe/.virtualenvs/ftorch/lib/python3.12/site-packages/torch/lib/libc10.so)
8: frame #1: c10::detail::torchCheckFail(char const*, char const*, unsigned int, std::string const&) + 0x64 (0x75a18ab2c776 in /home/joe/.virtualenvs/ftorch/lib/python3.12/site-packages/torch/lib/libc10.so)
8: frame #2: <unknown function> + 0x5991ab2 (0x75a17b391ab2 in /home/joe/.virtualenvs/ftorch/lib/python3.12/site-packages/torch/lib/libtorch_cpu.so)
8: frame #3: torch::autograd::backward(std::vector<at::Tensor, std::allocator<at::Tensor> > const&, std::vector<at::Tensor, std::allocator<at::Tensor> > const&, std::optional<bool>, bool, std::vector<at::Tensor, std::allocator<at::Tensor> > const&) + 0x57 (0x75a17b3951a7 in /home/joe/.virtualenvs/ftorch/lib/python3.12/site-packages/torch/lib/libtorch_cpu.so)
8: frame #4: <unknown function> + 0x59fb80d (0x75a17b3fb80d in /home/joe/.virtualenvs/ftorch/lib/python3.12/site-packages/torch/lib/libtorch_cpu.so)
8: frame #5: at::Tensor::_backward(c10::ArrayRef<at::Tensor>, std::optional<at::Tensor> const&, std::optional<bool>, bool) const + 0x48 (0x75a1775a9598 in /home/joe/.virtualenvs/ftorch/lib/python3.12/site-packages/torch/lib/libtorch_cpu.so)
8: frame #6: torch_tensor_backward_without_external_gradient + 0x8b (0x75a18b15ec4b in /home/joe/software/ftorch/build/libftorch.so)
8: frame #7: <unknown function> + 0xf0ef (0x64d2b73e90ef in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #8: <unknown function> + 0xc6c8 (0x64d2b73e66c8 in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #9: <unknown function> + 0x22836 (0x64d2b73fc836 in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #10: <unknown function> + 0x227eb (0x64d2b73fc7eb in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #11: <unknown function> + 0x762cb (0x64d2b74502cb in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #12: <unknown function> + 0x229d4 (0x64d2b73fc9d4 in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #13: <unknown function> + 0x22aa4 (0x64d2b73fcaa4 in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #14: <unknown function> + 0x2598e (0x64d2b73ff98e in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #15: <unknown function> + 0x2598e (0x64d2b73ff98e in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #16: <unknown function> + 0x793d0 (0x64d2b74533d0 in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #17: <unknown function> + 0x7dcd0 (0x64d2b7457cd0 in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #18: <unknown function> + 0x330bd (0x64d2b740d0bd in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #19: <unknown function> + 0x3a229 (0x64d2b7414229 in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #20: <unknown function> + 0x3a28a (0x64d2b741428a in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #21: <unknown function> + 0xc59c (0x64d2b73e659c in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8: frame #22: <unknown function> + 0x2a1ca (0x75a18a82a1ca in /lib/x86_64-linux-gnu/libc.so.6)
8: frame #23: __libc_start_main + 0x8b (0x75a18a82a28b in /lib/x86_64-linux-gnu/libc.so.6)
8: frame #24: <unknown function> + 0xc5d5 (0x64d2b73e65d5 in /home/joe/software/ftorch/build/test/unit/tensor/unittest_tensor_autograd)
8:  in torch_tensor_backward
1/1 Test #8: unittest_tensor_autograd .........***Failed    0.28 sec
```

PR #530, opened 2026-02-12
Use pointers in array tensor constructor functions

### LINKED ISSUE #417: Protect `torch_tensor_from_array` from abuse

joewallwork:
Closes #417
Supersedes #418

This PR picks up where @Mikolaj-A-Kowalski left off, rebases on top of `main`, and provides the additional documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,13 +49,6 @@ For specific details see the [FTorch online documentation](https://cambridge-icc
 - Example numbers were bumped to account for new worked examples in
   [#291](https://github.com/Cambridge-ICCS/FTorch/pull/291) and
   [#500](https://github.com/Cambridge-ICCS/FTorch/pull/500).
-- Make input array for `torch_tensor_from_array` have the `pointer, contiguous`
-  properties rather than `target`
-  [#530](https://github.com/Cambridge-ICCS/FTorch/pull/530). This change
-  technically breaks the API because it becomes no longer possible to pass
-  temporary Fortran arrays to the second argument of `torch_tensor_from_array`.
-  However, that was a bug rather than a feature, so any workflow crashes due to
-  this change will provide the user with information on how to remove the error.
 - Bump the minimum CMake version from 3.15 to 3.18, for consistency with what's
   used in PyTorch. [#491](https://github.com/Cambridge-ICCS/FTorch/pull/491)
 - Significant overhaul of the online FORD documentation and reduction of content in the README
@@ -75,6 +68,16 @@ For specific details see the [FTorch online documentation](https://cambridge-icc
 ### Removed
 
 - Windows CI disabled until GitHub runner issues resolved in [50ea6d7](https://github.com/Cambridge-ICCS/FTorch/commit/50ea6d78d79ebe638ebe597e745c015549f12a61)
+
+### Fixed
+
+- Make input array for `torch_tensor_from_array` have the `pointer, contiguous`
+  properties rather than `target`
+  [#530](https://github.com/Cambridge-ICCS/FTorch/pull/530). This change
+  technically breaks the API because it becomes no longer possible to pass
+  temporary Fortran arrays to the second argument of `torch_tensor_from_array`.
+  However, that was a bug rather than a feature, so any workflow crashes due to
+  this change will provide the user with information on how to remove the error.
 
 ### Patch Releases
 


examples/9_Autograd/autograd.f90
```diff
@@ -28,16 +30,24 @@ program autograd
...
 
   ! Scalar multiplication and division are not currently implemented in FTorch. However, you can
-  ! achieve the same thing by defining a rank-1 tensor with a single entry, as follows:
-  call torch_tensor_from_array(multiplier, [3.0_wp], torch_kCPU)
-  call torch_tensor_from_array(divisor, [3.0_wp], torch_kCPU)
+  ! achieve the same thing by defining a rank-1 tensor with a single entry, as follows.
+  ! NOTE: You must pay close attention to the lifetime and aliasing of the underlying memory. You
+  !  need to specify an array for each scalar value and make sure that the tensor does not outlive
+  !  it. Remember that if you provide the same array to two tensors and one of them modifies it, the
+  !  other will also see the change.
```
jatkinson1000: So in layperson's terms, one Fortran array to one tensor, unless you really know what you're doing.

examples/9_Autograd/autograd.f90
```diff
@@ -28,16 +30,24 @@ program autograd
...
+  multiplier_value(1) = 3.0_wp
```
jatkinson1000, line 48: I think the `(1)` is optional here?
Might make intent clearer, though one could also argue the converse.

pages/usage/troubleshooting.md
```diff
@@ -62,7 +62,30 @@ Error: There is no specific subroutine for the generic ‘torch_tensor_from_arra
 The first thing to do in this instance is to inspect the interface you are trying to
 call, and instead attempt to call the specific procedure you expect to use.
 This can often provide more instructive error messages about what you are doing
-incorrectly
+incorrectly.
+
+Such errors can occur if you pass a temporary array where the procedure expects
```
jatkinson1000:
```suggestion
Such errors can also occur if you pass a temporary array where the procedure expects
```

pages/usage/troubleshooting.md
```diff
@@ -62,7 +62,30 @@ Error: There is no specific subroutine for the generic ‘torch_tensor_from_arra
...
+to receive a Fortran array with the `target` property. For example:
+```
+   34 |   call torch_tensor_from_array(a, [1.0_wp], torch_kCPU, requires_grad=.true.)
+      |                                                                             1
+Error: There is no specific subroutine for the generic ‘torch_tensor_from_array’ at (1)
+```
+That is, the second argument should be a Fortran array with the `target`
+property, not the temporary array `[1.0_wp]`. This kind of thing was possible in
+FTorch at v1.0 but has since been removed because it is errorneous. Similarly
```
jatkinson1000:
```suggestion
FTorch at v1.0 but has since been removed because it is erroneous. Similarly
```

src/ftorch_tensor.fypp
```diff
@@ -308,6 +308,11 @@ contains
 
   !| Exposes the given data as a tensor without taking ownership of the original data.
   !  This routine will take an (i, j, k) array and return an (k, j, i) tensor.
+  !
+  ! Note that `data` needs to be a pointer to a **contiguous** block of memory!
+  ! It is generally not a case when calling `c_loc` on a Fortran array as these may not be
+  ! contiguous as a result of array slicing. Please consider asserting that the data is contiguous
+  ! with `is_contiguous` implicit procedure before calling this routine.
```
jatkinson1000: This did not read gramatically to me. Suggested change, provided it remains correct:
```suggestion
  ! This is not generally not the case when calling `c_loc` on a Fortran array as
  ! array slicing can lead to non-contiguous memory.
  ! Please consider asserting that the data is contiguous with the `is_contiguous`
  ! implicit procedure before calling this routine.
```

jatkinson1000 (commented):
Thanks @Mikolaj-A-Kowalski and @joewallwork 

I'm happy with the overall proposals here, just some suggestions to make the docs/communication of this clearer.

I also wonder if we need a "How to update your code for V2.0" page, similarly to how we did with v1.0?

Mikolaj-A-Kowalski (approved):
On my side I don't really have anything substantial to add. 

Except for thank you for picking this up and cleaning it up :-)

examples/9_Autograd/autograd.f90
joewallwork: It is optional but I felt that it was a nice reminder that it's a tensor with a single entry.

examples/9_Autograd/autograd.f90
joewallwork: Added this in 06700cd05984585b95650f7f25e0bfc0cdf64a71 because I found it a handy summary.

Issue #428, opened 2025-09-05
GPU workflow should run tests on the GPU


jatkinson1000:
As raised by #427 whilst there is now a GPU workflow to run examples (integration tests), many of them are still being run on CPU by default.

To run on GPU would require passing `--device='cuda'` from the commandline call to the python files and pt2ts script.

Really we should look to update the script to run everything possible on the GPU.

cc'ing @niccolozanotti as the original author of the workflow in case he has thoughts/comments.

zhenkunl:
That's a good idea. For integration tests, `CMakeLists.txt` in all examples need to be adjusted as well.

PR #565, opened 2026-03-18
Allow pre-trained models with `pt2ts`

joewallwork:
Merges into #563.

This PR modifies the `pt2ts` script so that it accepts pre-trained models. This is achieved by making the model definition file and input model file arguments optional. If neither are provided then an attempt is made to load the model name (now the only positional argument) from `torchvision.models`.

## Checklist

- [x] Allow pre-trained models in `pt2ts` script
- [x] Add precision argument to `pt2ts` script (needed for ResNet)
- [x] Add model weights argument to `pt2ts` script (needed for ResNet)
- [x] Move ResNet example to new approach
- [x] Update other examples to account for changes to `pt2ts` arguments
- [x] Update Python unit tests to account for changes
- [x] Add Python unit tests to cover the new functionality
- [x] Update documentation
- [x] Add changelog entry


CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ For specific details see the [FTorch online documentation](https://cambridge-icc
   [#555](https://github.com/Cambridge-ICCS/FTorch/pull/555)
 - Changed process for saving TorchScript models
   [#555](https://github.com/Cambridge-ICCS/FTorch/pull/555)
+- Extended `pt2ts` to accout for pre-trained torchvision models, weights, and
```
jatkinson1000:
```suggestion
- Extended `pt2ts` to account for pre-trained torchvision models, weights, and
```

examples/07_MultiGPU/CMakeLists.txt
```diff
@@ -65,8 +65,10 @@ if(CMAKE_BUILD_TESTS)
     #     and run basic checks
     add_test(
       NAME example_multigpu_pt2ts
-      COMMAND pt2ts ${PROJECT_SOURCE_DIR}/simplenet.py
-                    SimpleNet
+      COMMAND pt2ts SimpleNet
+                    --model_definition_file
+                    ${PROJECT_SOURCE_DIR}/simplenet.py
+                    --input_model_file
```
joewallwork, line 71: Looks like I missed this in the previous PRs and it didn't get caught because that example isn't run. More weight behind addressing #428. I'll get back to it.

examples/03_ResNet/resnet18.py
```diff
@@ -1,16 +1,21 @@
...
 
 
-# Initialize everything
 def initialize(precision: torch.dtype) -> torch.nn.Module:
     """
     Download pre-trained ResNet-18 model and prepare for inference.
 
+    These steps duplicate the process for loading pre-trained models in the pt2ts
+    script.
+
```
jatkinson1000: Maybe make this a `NOTE:`? My understanding is that it's worth keeping this in to allow running standalone from Python, but the statement here initially made it sound like it should perhaps be removed?

examples/03_ResNet/resnet18.py
jatkinson1000, line 1: I presume there was a good reason to bring model setup, processing of input image, and running out of functions and into `"__main__"`, but it's not immediately clear to me?

examples/03_ResNet/resnet_infer_python.py
```diff
@@ -88,15 +71,22 @@ def check_results(output: torch.Tensor) -> None:
         type=str,
         default=os.path.dirname(__file__),
     )
+    parser.add_argument(
+        "--data_dir",
```
jatkinson1000, line 75: We are now allowing users to set the data directory, but always assume its contents are hard coded as `categories.txt` and `dog.jpg`. It makes sense to hard code categories matching the weights, but I wonder if we also need to allow user-tailoring of the image file (dog remaining default for testing)? Otherwise I'm not sure what the benefit of not hard-coding `data-dir` as well is?

Unless it's to do with the CMake implementation?

examples/04_Batching/README.md
jatkinson1000, line 1: Is this a change that slipped through the cracks of #555 ?

ftorch_utils/pt2ts.py
```diff
@@ -42,15 +42,20 @@ def parse_user_input():
     )
     parser.add_argument(
         "--model_definition_file",
-        help="Filename for the definition of the PyTorch model, including path",
+        help=(
+            "Filename for the definition of the PyTorch model, including path. If not"
+            " provided, an attempt will be made to load the model from TorchVision's"
```
jatkinson1000: "the model" -> "a model matching the `model_name`"?

ftorch_utils/pt2ts.py
```diff
@@ -83,6 +97,12 @@ def parse_user_input():
         ),
         action="store_true",
     )
+    parser.add_argument(
+        "--precision",
+        help="Set the working precision of the model.",
```
jatkinson1000: Would this allow conversion of a model saved from one dytpe to another in TorchScript?
If so I can see some uses of that...?

jatkinson1000: It's used to set precision for all torch operations, not just the model?

ftorch_utils/README.md
```diff
@@ -29,16 +29,24 @@ The Torchscript model will be saved locally in the specified location.
...
-  path to the directory in which the TorchScript model should be saved.
+  the local file in which a PyTorch model is defined. (Not required for
+  pre-trained models.)
+* `--input_model_file </path/to/saved/model>`, which specifies the path to the
+  directory in which a local PyTorch model is saved. (Not required for
+  pre-trained models.)
+* `--output_model_file </path/to/save/model>`, which specifies the path to the
+  directory in which the TorchScript model should be saved. (Defaults to match
+  input model file.)
 * `--trace`, which allows you to switch from scripting to tracing.
```
jatkinson1000, line 42: I feel these comments/names are ambiguous as to whether they should be full filepaths, or paths to a directory.

ftorch_utils/README.md
```diff
@@ -29,16 +29,24 @@ The Torchscript model will be saved locally in the specified location.
...
 * `--test`, which allows you to run basic tests to check things are working as
   expected.
+* `--precision`, which specifies the working precision.
```
jatkinson1000: Perhaps an example? would I be expected to provide `32`, `float32`, `torch.float32`?

jatkinson1000 (commented):
Generally looking good @joewallwork thanks.

A few suggestions for clarification.
I'll also admit I didn't look too carefully at the exact test implementations beyond what they are testing for as I'll trust you and the CI.

examples/03_ResNet/resnet18.py
joewallwork: It didn't feel like it served any other purpose so I inlined it. I can revert to having functions if preferred.

examples/04_Batching/README.md
joewallwork: Yes

examples/03_ResNet/resnet18.py
joewallwork: Clarified in 7be93d54.

examples/03_ResNet/resnet_infer_python.py
joewallwork: Hard-coded datadir in bd1c23ed. We always want a subdir of the source dir so this is fine.

ftorch_utils/pt2ts.py
joewallwork: Fixed in 9d368a59.

ftorch_utils/pt2ts.py
joewallwork: Yes I believe so. Fixed in b50b7bf9530f97f408d470b6e9d8935def3970ca.

ftorch_utils/README.md
joewallwork: Ah yes. Fixed in 2a3e7382905e6a0e47facfb24f30e421a02cfd24.

ftorch_utils/README.md
joewallwork: Fixed in b50b7bf9530f97f408d470b6e9d8935def3970ca.

examples/03_ResNet/resnet18.py
jatkinson1000: I'm not overly wedded either way, just wondered why the change.
Fine to leave as is as I agree there is not particular merit either way given it's single-use and not egregiously long.

examples/03_ResNet/resnet_infer_python.py
jatkinson1000: Thanks, I agree that the most simple for now.

jatkinson1000 (approved):
Thanks for making those small changes @joewallwork I think this is good to go!

PR #575, opened 2026-04-14
Fixes for Zizmor security alerts

joewallwork:
Closes #518

There are currently several security alerts on the [Code scanning](https://github.com/Cambridge-ICCS/FTorch/security/code-scanning) panel. There are two issues being reported in several places:
* Commit hash for a version tag not matching the tag referenced in the inline comment following it. This has happened in places where we weren't specific enough about the tag and it changed. For example, some packages update the `vX` tag whenever a minor `vX.Y.Z` tag is released, so using `vX` mean things get out of date.
* Use of Windows CMD, which has not been the recommended shell for several years. I updated these to use Powershell. (See https://docs.zizmor.sh/audits/#misfeature.)

PR #576, opened 2026-04-14
Fix Powershell syntax

joewallwork:
Merges into #575


joewallwork:
@copilot Finish fixing the Powershell syntax here.

joewallwork:
@copilot I think there is an issue with how Torch_DIR is set because it isn't being picked up correctly.

PR #579, opened 2026-04-14
Test MultiGPU example in CI

### LINKED ISSUE #428: GPU workflow should run tests on the GPU

### LINKED ISSUE #544: Make the MultiGPU example more concise
See #541 for several ways that we can cut down and improve how the MultiGPU example is set up. Notably there is duplication in the `CMakeLists.txt`.

joewallwork:
Closes #428.
Closes #544.

This PR enables the MultiGPU example in the CI, albeit only on the one CUDA GPU that's available. This is done using conditional logic on the `MULTI_GPU` variable.

In addition, I made the CTest setup much more concise by handling the `GPU_DEVICE` variable.

The CUDA CI didn't initially trigger for my changes so I made some tweaks to get it to trigger when changes are made to the MultiGPU example specifically. While I was at it, I noticed the triggers for dependencies were out of date - they should've been `pyproject.toml` rather than `requirements.txt`.


joewallwork:
@copilot Fix the error in the CUDA CI with the ResNet example

joewallwork:
@copilot Fix the error in the CUDA CI with the ResNet example

.github/workflows/test_suite_ubuntu_cuda_gnu.yml
```diff
@@ -124,14 +130,14 @@ jobs:
...
 
-      - name: Run unit tests
+      - name: Run GPU-enabled Fortran unit tests
         run: |
           . ftorch/bin/activate
           cd build
           ctest --verbose --tests-regex cuda
 
-      - name: Run integration tests
+      - name: Run GPU-enabled integration tests
```
jatkinson1000: We just run the one test for now?

```suggestion
      - name: Run Multi-GPU integration test
```

.github/workflows/test_suite_ubuntu_cuda_gnu.yml
```diff
@@ -76,15 +82,15 @@ jobs:
           # Check this is <= what GPU-runner has installed (backwards compatible)
           pip install . --extra-index-url https://download.pytorch.org/whl/cu129 --group test
 
-      - name: Install cmake and NVIDIA dev toolkit
+      - name: Run GPU-enabled Python unit tests
```
jatkinson1000:
```suggestion
      - name: Run GPU-specific FTorch-utils unit tests
```

.github/workflows/test_suite_ubuntu_cuda_gnu.yml
```diff
@@ -76,15 +82,15 @@ jobs:
...
+          . ftorch/bin/activate
+          pytest --verbose test/ftorch_utils -k cuda
 
-        # Currently used by example_mpi
-      - name: Install an MPI distribution
+      - name: Install cmake, NVIDIA dev toolkit, and an MPI distribution
         run: |
-          sudo apt install -y openmpi-bin openmpi-common libopenmpi-dev
+          sudo apt update
+          sudo apt install -y cmake nvidia-cuda-toolkit openmpi-bin openmpi-common libopenmpi-dev
```
jatkinson1000: I'd maybe suggest these were moved with installation of other dependencies above, so that installation and running of tests are collected together.

examples/07_MultiGPU/multigpu_infer_fortran.f90
```diff
@@ -37,26 +37,29 @@ program inference
    ! Flag for testing
    logical :: test_pass
 
-   ! Get device type as first command line argument and TorchScript model file as second command
-   ! line argument
+   ! Get TorchScript model file, device type, and number of devices as second command line arguments
```
jatkinson1000: They can't all be the second argument...?

examples/07_MultiGPU/multigpu_infer_fortran.f90
```diff
@@ -37,26 +37,29 @@ program inference
...
-    else if (trim(args(1)) == "hip") then
+    else if (trim(args(2)) == "hip") then
       device_type = torch_kHIP
-   else if (trim(args(1)) == "xpu") then
+   else if (trim(args(2)) == "xpu") then
       device_type = torch_kXPU
-   else if (trim(args(1)) == "mps") then
+   else if (trim(args(2)) == "mps") then
       device_type = torch_kMPS
-      num_devices = 1
```
jatkinson1000, line 55: I think this still needs to be enforced, unless newer M chips have introduced multiple devices?

Might be easiest to reinstate this line and move line 62 (setting num_devices) before the device type loop?

examples/07_MultiGPU/multigpu_infer_python.py
```diff
@@ -79,18 +79,22 @@ def deploy(saved_model: str, device: str, batch_size: int = 1) -> torch.Tensor:
...
+    num_devices = parsed_args.num_devices
     saved_model_file = os.path.join(
         filepath, f"torchscript_multigpu_model_{device_type}.pt"
     )
 
     batch_size_to_run = 1
 
-    # Use 2 devices unless MPS for which there is only one
-    num_devices = 1 if device_type == "mps" else 2
-
```
jatkinson1000, line 93: As with the Fortran, I think this is still required?

examples/07_MultiGPU/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ cmake_policy(SET CMP0076 NEW)
 
 set(PROJECT_NAME MultiGPUExample)
 
-project(${PROJECT_NAME} LANGUAGES Fortran)
+project(${PROJECT_NAME} LANGUAGES C CXX Fortran)
```
jatkinson1000: Why do we need to add C and C++?

examples/07_MultiGPU/simplenet.py
```diff
@@ -20,7 +20,7 @@ def __init__(
         self._fwd_seq = nn.Sequential(
             nn.Linear(5, 5, bias=False),
         )
-        with torch.inference_mode():
+        with torch.no_grad():
```
jatkinson1000, line 23: The other examples use `inference_mode`.
Is there a reason to prefer `no_grad`?
If so we should have a follow-up to change elsewhere.

See also ll. 77

examples/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ if(CMAKE_BUILD_TESTS)
   # NOTE: Example 06_Looping is built here but not run as part of the
   #       integration test suite. This is because it demonstrates
   #       'good' versus 'bad' practice, as opposed to functionality.
-  if(NOT "${GPU_DEVICE}" STREQUAL "NONE" AND MULTI_GPU)
+  if(NOT "${GPU_DEVICE}" STREQUAL "NONE")
```
jatkinson1000: The multiGPU CMakeLists also checks for `"${GPU_DEVICE}" STREQUAL ""` is that also needed here?

ftorch_utils/pt2ts.py
jatkinson1000, line 1: [Our docs](https://cambridge-iccs.github.io/FTorch/page/installation/gpu.html) suggest that to save a torchscript model to be run on a GPU device you have to move it to the device before calling `jit.save`.

And indeed, pt2ts used to include device in the filename etc.

But looking at this we save from CPU or GPU and then load to GPU in Fortran.
If that is working correctly we should update the docs.

I note resnet_infer_python loads to model and then explicitly calls `.to(device)`.

ftorch_utils/validation.py
```diff
@@ -124,3 +126,42 @@ def validate_output_tensors(expected_tensor, result_tensor):
...
+        The TorchScript model to be tested.
+    """
+    device_type = list(model.parameters())[0].device.type
+    for parameter in model.parameters():
+        if parameter.device.type != device_type:
+            raise RuntimeError(
+                f"The model has parameters on different devices ('{device_type}' vs."
+                f" '{parameter.device.type}'). Ensure all model parameters are on the"
+                " same device and try again."
+            )
```
jatkinson1000:
```suggestion
                "The model has parameters on different devices (model on "
                f"'{device_type}' vs. parameters on '{parameter.device.type}')."
                "Ensure all model parameters are on the same device and try again."
            )
```

ftorch_utils/validation.py
```diff
@@ -124,3 +126,42 @@ def validate_output_tensors(expected_tensor, result_tensor):
...
+    input_tensors : tuple of torch.Tensor
+        The input tensors used for testing the model.
+    """
+    device_type = list(model.parameters())[0].device.type
+    for i, input_tensor in enumerate(input_tensors):
+        if device_type != input_tensor.device.type:
+            raise RuntimeError(
+                f"The model is on a different device from input tensor {i}"
+                f" ('{device_type}' vs. '{input_tensor.device.type}'). Ensure they are"
+                " on the same device and try again."
```
jatkinson1000, line 191:
```suggestion
                f"The model is on a different device from input tensor {i}"
                f" (model on '{device_type}' vs. input tensor on"
                f" '{input_tensor.device.type}'). Ensure they are"
                " on the same device and try again."
```

ftorch_utils/validation.py
jatkinson1000, line 1: Should these also be checking device number as well as type...?

test/ftorch_utils/test_validation.py
```diff
@@ -150,3 +153,61 @@ def test_tuple_mismatching(self):
...
+        if torch.cuda.device_count() == 0:
+            pytest.skip("No CUDA devices available, skipping test.")
+        try:
+            validate_device_types(
+                torch.nn.Linear(2, 2).to("cuda"),
+                torch.tensor([1.0, 2.0, 3.0]).to("cuda"),
+            )
+        except RuntimeError:
+            pytest.fail("validate_device_types raised RuntimeError unexpectedly!")
+
```
jatkinson1000, line 225: Not sure I've seen try, except, pytest.fail as a way to check expected behaviour works before. I'll have to remember this as a tactic in future!

examples/07_MultiGPU/README.md
```diff
@@ -84,9 +84,15 @@ compiler:
...
-cmake .. -DCMAKE_PREFIX_PATH=<path/to/your/installation/of/library/> -DCMAKE_BUILD_TYPE=Release
+cmake .. \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DCMAKE_PREFIX_PATH=<path/to/your/installation/of/library/> \
+  -DGPU_DEVICE=<my_device_type>
 cmake --build .
 ```
+Note that the `GPU_DEVICE` flag needs to be used again, despite the fact that it
+was provided during the FTorch build. This is because it is used in the CTest
+setup logic to automatically support testing on different device types.
```
jatkinson1000, line 95:
```suggestion
setup logic to automatically support testing on different device types.
In general, when building FTorch code for GPU on your project, this is not needed.
```

jatkinson1000 (commented):
Thanks @joewallwork have read through and a few suggestions/queries, but all generally pretty small.

.github/workflows/test_suite_ubuntu_cuda_gnu.yml
joewallwork: Done in cbbdaa3.

examples/07_MultiGPU/multigpu_infer_fortran.f90
joewallwork: Fixed in 5cf1d2f.

examples/07_MultiGPU/multigpu_infer_fortran.f90
joewallwork: Done in 9cc4deb.

examples/07_MultiGPU/multigpu_infer_python.py
joewallwork: Done in 9cc4deb.

examples/CMakeLists.txt
joewallwork: Done in 05d50dd.

examples/07_MultiGPU/CMakeLists.txt
joewallwork: Tried dropping in 94d363f, let's see what happens.

ftorch_utils/validation.py
joewallwork: Added in 3004f54 although the tests will never run on the CI since it only has one CUDA device.

examples/07_MultiGPU/simplenet.py
joewallwork: Reverted in a5deb4a.

ftorch_utils/pt2ts.py
joewallwork: I don't fully follow what you're suggesting here but added something in 5bffb7f. Also generalised the ResNet device handling in cb32b4cadc27311659f2cd84b62d1ffa4c2fe2e1.

examples/07_MultiGPU/simplenet.py
joewallwork: Ah okay, this is why:
https://github.com/Cambridge-ICCS/FTorch/actions/runs/28456446785/job/84332440079
Re-reverted in a3e881e.

examples/07_MultiGPU/CMakeLists.txt
jatkinson1000: Nothing, it seems ;)

joewallwork: I think it had slipped in due to #542. But yes, it's not needed here.

joewallwork:
@jatkinson1000 Please could you re-review this? I think it's pretty much good to go.

ftorch_utils/pt2ts.py
jatkinson1000: [The docs say that a TorchScript model to be loaded to GPU has to be saved from GPU](https://cambridge-iccs.github.io/FTorch/page/installation/gpu.html).

Looking at the [definition of `load_pytorch`](https://github.com/Cambridge-ICCS/FTorch/blob/2010dd4ca4ee98d3f9b82db31f62774637c69312/ftorch_utils/torchscript.py#L56) from pt2ts, however, I think it [will load the `torch.save`d model to CPU](https://docs.pytorch.org/tutorials/beginner/saving_loading_models.html#save-on-gpu-load-on-gpu), regardless of the device it was saved from when using `torch.save`. Because the new model we are loading the saved weights to has been instantiated on the CPU?

So this model is then saved to TorchScript from CPU.

And the model then gets cast to the correct device for inference, either explicitly in Python using `.to()` or at load time in FTorch via `device_type=torch_kCUDA` etc?

.github/workflows/test_suite_ubuntu_cuda_gnu.yml
```diff
@@ -78,16 +84,16 @@ jobs:
           # NOTE: CUDA version should be consistent with the download below, i.e., 13.2
           pip install . --extra-index-url https://download.pytorch.org/whl/cu132 --group test
 
-      - name: Install cmake and CUDA toolkit
+      - name: Install cmake, GNU compilers, m4, and CUDA toolkit
         run: |
           sudo apt update
           # NOTE: CUDA version should be consistent with the download above, i.e., 13.2
```
jatkinson1000, line 90: I don't think this (CUDA version) is being explicitly forced anywhere now.
Is it left as a breadcrumb incase things go wrong in future, or should it be removed?

examples/03_ResNet/resnet_infer_python.py
```diff
@@ -18,7 +18,7 @@ def deploy(
     saved_model : str
         location of ResNet-18 saved to Torchscript
     device : str
-        Torch device to run model on, 'cpu' or 'cuda'
+        Torch device to run model on, e.g., 'cpu' or 'cuda'
```
jatkinson1000, line 21: This argument appears to be unused in the `deploy` function.
To my reading the model is always deployed/run on CPU here?

ftorch_utils/pt2ts.py
```diff
@@ -162,13 +166,21 @@ def main_cli():
...
+        validate_device_indices(model, input_tensors)
+
         # Propagate the input tensor through the PyTorch model
         # If something isn't working then this will generate an error
         pt_model_outputs = model(*input_tensors)
 
         # Load the TorchScript model, propagate the same input tensor, and check the
         # results match
         ts_model = load_torchscript(output_model_file)
+        ts_model = ts_model.to(input_tensors[0].device)
```
jatkinson1000, line 183: Is this required?
I think it is, based on the behavior of creating a model and loading from the state dict, but part of the above discussion.

ftorch_utils/pt2ts.py
```diff
@@ -162,13 +166,21 @@ def main_cli():
     if test:
         validate_file_exists(output_model_file, "Saved TorchScript output model")
 
+        # Transfer model to the same device as the input tensor
+        model = model.to(input_tensors[0].device)
```
jatkinson1000, line 170: Is this required?
I think it is, based on the behavior of creating a model and loading from the state dict, but part of the above discussion.

ftorch_utils/pt2ts.py
jatkinson1000: OK, wrote myself some toy code and followed this through, I think I was partly lagging in my understanding of the deeper parts of the new pt2ts, apologies.

My understanding now is that when users `torch.save` from a device, this is passed to the saved file, and persists* in subsequent loading and saving to TorchScript, and then loading from TorchScript (provided [`map_location`](https://docs.pytorch.org/docs/main/generated/torch.load.html) is not specified).
[*well... it always gets read to CPU first, but is then mapped to the device it was saved from]

What threw me off is that the [testing part of pt2ts](https://github.com/Cambridge-ICCS/FTorch/blob/a3e881e4a059d5ad1877d9b479c0d3cc9b5e986d/ftorch_utils/pt2ts.py#L166-L187) now explicitly maps models and tensors `.to()` devices, which, based on the above statement, I thought would be redundant? (I have commented a query on these lines explicitly below.)


However, I then took a closer look at the definition of [`load pytorch`](https://github.com/Cambridge-ICCS/FTorch/blob/2010dd4ca4ee98d3f9b82db31f62774637c69312/ftorch_utils/torchscript.py#L56) which I think always constructs the model on the CPU as it instantiates the class by importing from the definition file and then calling [`load_state_dict`](https://docs.pytorch.org/docs/2.12/generated/torch.nn.Module.html#torch.nn.Module.load_state_dict) which copies to the model?

This would make sense for why the explicit mapping `.to()` had to be added in pt2ts below (where I have queried it).

However, in this case it seems redundant/confusing to go through the process of saving from GPU, reading to a CPU-based model, and casting back to GPU - we could have just saved from CPU. The process of setting the device from `input_tensors[0]` also feels a little fragile. I wonder if instead there should be a `device` argument to pt2ts that then makes use of [`map_location`](https://docs.pytorch.org/docs/main/generated/torch.load.html) to ensure things are saved and run on appropriate devices? Though this would go to a follow-up PR.

It may be that I have all this wrapped around my head though, so do let me know if this makes any sense.

jatkinson1000: In summary, I think this can probably pass for the purposes of this PR, but would appreciate your thoughts and think it should be dealt with (if required) before a release bundling the new pt2ts.

pages/installation/gpu.md
```diff
@@ -102,10 +102,10 @@ TorchScript model. For further details on the `pt2ts` script, call
 
 **4) Specify the target device from FTorch**  
 When calling [[ftorch_tensor(module):torch_tensor_from_array(interface)]] and
-[[ftorch_model(module):torch_model_load(subroutine)]]  in Fortran,
-the device type for the input tensor(s) and model should be set to the appropriate
-device type (`torch_kCUDA`, `torch_kHIP`, `torch_kXPU`, or `torch_kMPS`) rather
-than `torch_kCPU`.
+[[ftorch_model(module):torch_model_load(subroutine)]] in Fortran, the device
+for the input tensor(s) and model should be set to the same device type that
+they were written out using (`torch_kCUDA`, `torch_kHIP`, `torch_kXPU`, or
```
jatkinson1000, line 107: Input tensors are probably not being loaded from a saved file?
I think "appropriate" was probably appropriate here, apologies, and have tried to clarify what I was getting at in the pt2ts discussion.

jatkinson1000 (commented):
Thanks @joewallwork I have re-reviewed and done some investigation to check my understanding.

I think the bulk of this is fine, but would still like to check we have the same understanding of devices models go to and from when saving, and whether there should be any subsequent changes to pt2ts.

.github/workflows/test_suite_ubuntu_cuda_gnu.yml
joewallwork: Done in bab5710.

ftorch_utils/pt2ts.py
joewallwork: Okay hopefully I addressed things correctly in https://github.com/Cambridge-ICCS/FTorch/pull/579/commits/b4e0e702850fd03584122f8b76319806aa44517e.

ftorch_utils/pt2ts.py
joewallwork: Okay hopefully I addressed things correctly in https://github.com/Cambridge-ICCS/FTorch/pull/579/commits/b4e0e702850fd03584122f8b76319806aa44517e.

examples/03_ResNet/resnet_infer_python.py
joewallwork: Good spot. Done in 612bc6813a796ffc2c5e34b3b4d91b5210dd8b7c.

pages/installation/gpu.md
joewallwork: Updated docs in 1b32195a64c88adf53906a0409a21b6aa9d69d5e.

test/ftorch_utils/test_torchscript.py
```diff
@@ -118,6 +118,35 @@ def test_load_pytorch_missing_saved_model_file(filename):
...
+        model_definition_file=os.path.abspath(__file__),
+        saved_model_file=filename,
+    )
+    # NOTE: models don't have a device attribute themselves so you have to query params
+    assert next(model.parameters()).device.type == "cpu"
+
+
+def test_load_pytorch_cuda_device(filename):
+    """Check that `load_pytorch` is able to load a model on a CUDA device on request."""
+    if torch.cuda.device_count() == 0:
```
jatkinson1000: I think:

```suggestion
    if not torch.cuda.is_available():
```

is the canonical way way to check this.

examples/03_ResNet/resnet_infer_python.py
```diff
@@ -32,30 +32,17 @@ def deploy(
     transposed_shape = [224, 224, 3, batch_size]
     precision = np.float32
 
+    # Setup input tensor
     np_data = np.fromfile(os.path.join(data_dir, "image_tensor.dat"), dtype=precision)
     np_data = np_data.reshape(transposed_shape)
     np_data = np_data.transpose()
     input_tensor = torch.from_numpy(np_data)
```
jatkinson1000: Should this be:
```suggestion
    input_tensor = torch.from_numpy(np_data).to(device)
```
to match the model below?

examples/03_ResNet/resnet_infer_python.py
```diff
@@ -32,30 +32,17 @@ def deploy(
...
-        model = torch.jit.load(saved_model)
-        input_tensor_gpu = input_tensor.to(torch.device("cuda"))
-        output_gpu = model.forward(input_tensor_gpu)
-        output = output_gpu.to(torch.device("cpu"))
-
-    else:
-        device_error = f"Device '{device}' not recognised."
-        raise ValueError(device_error)
+    # Load saved TorchScript model
+    model = torch.jit.load(saved_model).to(device)
```
jatkinson1000: Based on the long text, I would have thought this was no longer strictly required (new `pt2ts` should save TorchScript with the device encoded, and then load bac on to this device).

Is this that something that was added because it didn't work otherwise (and the above is wrong) or just to be extra safe?

As a side note I think the input tensor also needs transferring to `device` a couple of lines above.

jatkinson1000 (commented):
Thank you so much for revisiting this @joewallwork 

This now looks sensible to my understanding (as documented in the thread) but disclosure it has been a while so I had to remind myself of this.

A couple of quick final queries to check before I approve this but I think it's pretty much there.

If I understood wrong or something doesn't work we'll probably find out pretty soon 😅

examples/03_ResNet/resnet_infer_python.py
joewallwork: Ah yes. Should be addressed in a72b4cb.

jatkinson1000 (approved):
Thanks @joewallwork I think this is now good to go!

joewallwork:
Great! Thanks for the re-review.

PR #581, opened 2026-04-22
Fortran training example

joewallwork:
Closes #446.

This PR demonstrates training a model in Fortran using an optimizer. Good old SimpleNet is retrained to be able to permute an input array, shifting by one entry. This is demonstrated to work, given some fairly arbitrary training parameters. Here's a loss curve I was able to plot:

losses

To enable this, we need to be able to access the model parameters to pass to the model so some source changes were required. See https://docs.pytorch.org/cppdocs/frontend.html#end-to-end-example for an end-to-end example in C++.

## Checklist

- [x] Add README
- [x] Docs
- [x] Drop Python version as not needed
- [x] Better explain what we are trying to retrain the model to do.
- [x] Tests for the source code changes
- [x] Changelog entry.

PR #589, opened 2026-05-20
CI fixes after merging online training

joewallwork:
This PR fixes the following CI failures:
* https://github.com/Cambridge-ICCS/FTorch/actions/runs/26089177778/job/76710058860 (fypp)
* https://github.com/Cambridge-ICCS/FTorch/actions/runs/26089177838/job/76710058731 (CUDA)
* https://github.com/Cambridge-ICCS/FTorch/actions/runs/26089177768/job/76710058940 (Windows)
* https://github.com/Cambridge-ICCS/FTorch/actions/runs/26089177791/job/76710058759 (Ubuntu / Intel)

## Summary

* Apply `fypp` to get `ftorch_tensor.fypp` and `ftorch_tensor.f90` realigned.
* Fix a rather silly error in `ftorch_model.f90` that was causing the Intel compiler to crash.
* Align CUDA version between toolkit and PyTorch.

## Other changes added in

* Tweaked flags passed to Intel compiler to reduce warnings.

### DIFF
--- a/.github/workflows/test_suite_ubuntu_cuda_gnu.yml
+++ b/.github/workflows/test_suite_ubuntu_cuda_gnu.yml
@@ -100,6 +100,9 @@ jobs:
         env:
           FORTRAN_STANDARD: f2008
         run: |
+          # GitHub Actions doesn't source /etc/profile.d/ scripts, so the CUDA bin directory
+          # (added by the apt package post-install script) won't be in PATH automatically.
+          export PATH=/usr/local/cuda/bin:$PATH
           . ftorch/bin/activate
           VN=$(python -c "import sys; print('.'.join(sys.version.split('.')[:2]))")
           export Torch_DIR=${VIRTUAL_ENV}/lib/python${VN}/site-packages
--- a/.github/workflows/test_suite_ubuntu_cuda_gnu.yml
+++ b/.github/workflows/test_suite_ubuntu_cuda_gnu.yml
@@ -72,13 +72,12 @@ jobs:
           python -m pip install --upgrade pip
           python -m venv ftorch
           . ftorch/bin/activate
-          # CUDA v12.9 chosen as latest supported by PyTorch
-          # Check this is <= what GPU-runner has installed (backwards compatible)
-          pip install . --extra-index-url https://download.pytorch.org/whl/cu129 --group test
+          pip install . --extra-index-url https://download.pytorch.org/whl/cu132 --group test
 
       - name: Install cmake and CUDA toolkit
         run: |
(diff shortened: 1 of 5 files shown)


joewallwork:
I managed to reproduce the Intel errors locally and am working on a fix.

joewallwork:
@copilot Fix the issue with the CUDA CI workflow. There seems to be a CUDA version mismatch between the toolkit and the PyTorch installation that comes as a dependency of `ftorch_utils`. Previously, we were using `sudo apt install nvidia-cuda-toolkit` but this comes with CUDA 12.0 which doesn't have a corresponding PyTorch wheel at https://download.pytorch.org/whl/.

.github/workflows/test_suite_ubuntu_cuda_gnu.yml
```diff
@@ -72,14 +72,13 @@ jobs:
           python -m pip install --upgrade pip
           python -m venv ftorch
           . ftorch/bin/activate
-          # CUDA v12.9 chosen as latest supported by PyTorch
-          # Check this is <= what GPU-runner has installed (backwards compatible)
-          pip install . --extra-index-url https://download.pytorch.org/whl/cu129 --group test
+          pip install . --extra-index-url https://download.pytorch.org/whl/cu132 --group test
```
jatkinson1000, line 76: Is it worth preserving a comment in here for developers as it seems a common failure point?

jatkinson1000 (approved):
Thanks @joewallwork if the tests pass that seems OK.

Interested as to why the changes are required - we specify the compiler version so I'm surprised `-fpscomp logicals` behaviour changes, unless it's something specifically related to the training code from #581?

And similarly for needing standard in Windows.

Happy for you to merge as a fix after checking whether we need to update Windows/intel installation documentation to add these additional flags.

joewallwork:
> Interested as to why the changes are required - we specify the compiler version so I'm surprised `-fpscomp logicals` behaviour changes, unless it's something specifically related to the training code from #581?

It's not needed, it reduces the number of warnings produced by the compiler. Sorry I should've been clearer in my mention of this at the bottom of the PR body.
 
> And similarly for needing standard in Windows.

That was just for consistency (and to trigger the Windows CI...)

> Happy for you to merge as a fix after checking whether we need to update Windows/intel installation documentation to add these additional flags.

Could add a common warnings entry I suppose?

jatkinson1000:
Ah, in that case I don't think there is anything critical to be added to the docs, so will leave up to you if its worth mentioning to reduce warnings or whether to let the user choose additional flags.

PR #591, opened 2026-05-22
Implement MSE loss function

### LINKED ISSUE #592: Implement MSELoss
I decided to keep this separate from CrossEntropyLoss, for simplicity.

joewallwork:
Closes #592

[Ignore the branch name - I implemented MSELoss first and then decided to stick with that rather than implementing CrossEntropy and others in the same PR.]

## Checklist

- [x] Implement `MSELoss`
- [x] Implement enumerator to specify reduction type
- [x] Allow kNone and kSum in reduction type
- [x] Use `MSELoss` in optimizer example
- [x] Unit tests to check consistency of `MSELoss` with `torch_tensor_mean` and `torch_tensor_sum` when appropriate reduction parameters used
- [x] Add changelog entry
- [x] Add docs

## Questions for reviewer

* ~~I realise that `torch_tensor_mean` has its output as the first argument whereas my proposed `torch_loss_mse` has its output as the last argument. Is there a standard pattern we want to follow?~~ Edit: I decided it's best to put the output first and implemented in 4a2d8cd. The loss functions return the output and often have optional arguments so it makes sense for subroutine implementations to have the output argument first.


joewallwork:
@copilot+claude-sonnet-4.6 Can you fix the FIXME notes in `src/ctorch.cpp` for the implementation of `get_libtorch_reduction_type`? There is some inconsistency with the different options in the Torch C++ API. You may need to look at the PyTorch source code. This is also causing a CI failure.

jatkinson1000 (commented):
Thanks @joewallwork this generally looks sensible to me.

The only thing that isn't obvious to me from the code is the reduction types?
We can call the mseloss with a reduction other than mean? or is this the reduction we take on the vector after applying mse?

If you could add some explanation of what this argument is somewhere I think this is good to go.

joewallwork:
> The only thing that isn't obvious to me from the code is the reduction types? We can call the mseloss with a reduction other than mean? or is this the reduction we take on the vector after applying mse?

I was just going with the naming used by PyTorch.

> If you could add some explanation of what this argument is somewhere I think this is good to go.

Added something in e8e2fda.

joewallwork:
Better explanation in 6737658.

src/ftorch_loss.f90
```diff
@@ -0,0 +1,61 @@
...
+
+contains
+
+  ! ============================================================================
+  ! --- Procedures for evaluating specific loss functions
+  ! ============================================================================
+
+  !| Evaluate MSELoss
+  !
+  !  Note that if the reduction type relates to the operation to perform across batches. Within each
```
jatkinson1000:
```suggestion
  !  Note that the reduction type relates to the operation to perform across batches. Within each
```

jatkinson1000 (commented):
Thanks @joewallwork one small typo but I still have a question about the definition.

Your docs now say (to my reading) that the reduction relates to reducing across multiple batches, but looking at the way the unit tests are written it would seem that the reduction applies within a single batch...? Unless I have misunderstood something.

joewallwork:
> Thanks @joewallwork one small typo but I still have a question about the definition.
> 
> Your docs now say (to my reading) that the reduction relates to reducing across multiple batches, but looking at the way the unit tests are written it would seem that the reduction applies within a single batch...? Unless I have misunderstood something.

Playing around with PyTorch it seems that my initial understanding was correct. The reduction operator is indeed applied to the output rather than over batches. See https://docs.pytorch.org/docs/main/generated/torch.nn.modules.loss.MSELoss.html

Unfortunately it seems to be inconsistent with CrossEntropyLoss
```python
>>> import torch
>>> mse = torch.nn.MSELoss(reduction="none")
>>> input = torch.randn(3, 5, requires_grad=True)
>>> target = input
>>> mse(target, input)
tensor([[0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0.]], grad_fn=<MseLossBackward0>)
>>> mse = torch.nn.MSELoss(reduction="mean")
>>> mse(target, input)
tensor(0., grad_fn=<MseLossBackward0>)
>>> ce = torch.nn.CrossEntropyLoss(reduction="none")
>>> ce(target, input)
tensor([ -3.7469, -10.7889,  -2.1221], grad_fn=<NegBackward0>)
>>> ce = torch.nn.CrossEntropyLoss(reduction="mean")
>>> ce(target, input)
tensor(-5.5526, grad_fn=<DivBackward1>)
```

So I guess revert to previous explanation and add extra clarification?

jatkinson1000:
I'm in no way an ML expert, but I think this is maybe correct...?
Based on [docs](https://docs.pytorch.org/docs/2.12/generated/torch.nn.CrossEntropyLoss.html) CEL collapses across classes (5 in your above example) to give a vector of batch length?
Because we are looking at classification for a measure of "right" or "wrong".

I think clarify that reduction operates within a batch (PyTorch docs use "minibatch"?) and then point to PyTorch docs for longer definition of a specific loss type?
I guess some clarification of expected sizes here is useful as people may be defining a Fortran array to recieve the loss tensor.

joewallwork:
> I'm in no way an ML expert, but I think this is maybe correct...? Based on [docs](https://docs.pytorch.org/docs/2.12/generated/torch.nn.CrossEntropyLoss.html) CEL collapses across classes (5 in your above example) to give a vector of batch length? Because we are looking at classification for a measure of "right" or "wrong".
> 
> I think clarify that reduction operates within a batch (PyTorch docs use "minibatch"?) and then point to PyTorch docs for longer definition of a specific loss type? I guess some clarification of expected sizes here is useful as people may be defining a Fortran array to recieve the loss tensor.

Okay how about 5efbebf?

jatkinson1000 (approved):
LGTM, Thanks @joewallwork

PR #599, opened 2026-06-09
Setup fpm

joewallwork:
Closes #77 
Supersedes #429 

## Checklist

- [x] Add fpm build option to GNU test workflow
- [x] Build and run standalone example using fpm
- [ ] Move to pip installing fpm once https://github.com/fortran-lang/fpm/pull/1250 merged 
- [ ] See if we can run Python scripts using fpm, too
- [-] ~~Enable `auto-examples` to run integration tests under fpm in CI~~
- [ ] Add docs on building with fpm
- [ ] CHANGELOG entry


joewallwork:
@rouson Here is the PR I mentioned where we're trying to set up fpm for FTorch

joewallwork:
@copilot Fix linking error with libtorch in the `Standalone Tensor Example` CI job.

joewallwork:
@copilot Fix the error in the Standalone Tensor Example CI job under the fpm build system. It worked fine with the tensor manipulation example so why does it fail when applied to the SimpleNet example?

joewallwork:
Opened https://github.com/fortran-lang/fpm/issues/1304 to query about whether it's possible to run Python scripts etc using fpm.

PR #608, opened 2026-06-23
Ensure model output file is not `None`

### LINKED ISSUE #607: Disallow `None` being the output model file name in `pt2ts`
Going through the FTorch summer school run-through today I noticed that if you forget to include an output model filename then it defaults to `None` and this gets accepted. We should disallow filenames not ending in `.pt`.

joewallwork:
Closes #607

Does what it says on the tin.

Note that we were skipping the model output name validation if it was `None`, allowing this to be used as the output filename.


jatkinson1000 (approved):
LGTM

PR #614, opened 2026-06-30
Check for consistent device types when applying overloaded operators

### LINKED ISSUE #301: Raise error if overloaded operator applied to tensors on different device indices
In #272 it was pointed out that as well as having a guard against mixing device types we should also perhaps add one for mixing device indices.

This would be added in a similar way to #272.

joewallwork:
Closes #301.


jatkinson1000 (approved):
Seems legit.

PR #617, opened 2026-07-23
Add Fortitude check for `ftorch_loss.f90`

joewallwork:
I just spotted that we missed `ftorch_loss.f90` here, too.

How come we are spelling out the files in `src/` rather than just doing `fortitude check src/`?


jatkinson1000:
I suspect for the same reason that we are installing all mod files by name instead of a glob pattern - we started with just ftorch.f90 named explicitly, then added optimizers, then just kept following the pattern.

jatkinson1000 (approved):
LGTM

PR #618, opened 2026-07-29
Update MacOS workflow to use OPENMP_ROOT instead of explicit flags.

jatkinson1000:
Rebuilding on my machine and tacking a few issues it seems that the problem seen previously was a conflict between bundled OMP in Torch and the system OMP installed via brew.

This is a cleaner fix on my machine so seeing if it can be applied here.

Since it works I have also updated the MacOS documentation.


jatkinson1000:
Additional context - setting the compile flags globally is what was causing my resnet example to segfault on my local machine.

joewallwork (approved):
If you and the Mac CI are happy then I am happy.

joewallwork:
Certainly much neater!

PR #625, opened 2026-09-16
Bump CXX standard from 17 to 20 to meet latest PyTorch.

### LINKED ISSUE #624: Bump CXX standard to 20
We currently specify CXX 17 is required, but Torch now uses CXX 20 features.

We need to bump this in the CMakeLists and docs.

Also check if Applying 20 is backwards compatible to previous versions of Torch.

jatkinson1000:
Closes #624

Bumps CXX standard expected by FTroch in CMakeLists to 20 to match that in the latest Torch, which uses CXX20 features.

I checked against PyTorch 3.11 and tests still build and run OK.


joewallwork (approved):
Makes sense. I think you got everywhere we need to mention this.

PR #604, opened 2026-06-16
Refactor torch_tensor_from_array

jatkinson1000:
Closes #596 

Creates a new routine as described in #596.

May also close #97 and #56.
May also address queries in #412.
Will replace #423 and #415 

## TODO:

- [x] Update docs that are now out of date:
  - [x] usage/tensor
  - [x] usage/transposing
  - [x] usage/troubleshooting
- [x] Update CHANGELOG

PR #628, opened 2026-09-16
Api types refactor

### LINKED ISSUE #619: Removal of `ftorch_int`?
My original title and reason for opening this issue was that`ftorch_int` leaked into the public interface via the `layout` argument.
However, digging into it I thing there are other issues stemming from `ftorch_int`.

Effects have so-far gone unnoticed, likely because `ftorch_int` (and `c_int` used for other subroutine inputs) is `int32`, matching the typical `integer` on most systems.

It is currently used only as the layout input to `ttfa` However, for proper compatibility with the rest of the API we should make this `c_int`, consistent with other input arguments.

I think the documented usage as being changable for "large size tensors" is also false, as `layout` holds only dimension numbers, not dimension sizes. Indeed, tensor shapes and strides are already hard-coded as `c_int64_t`.
Also, changing `ftorch_int` to `int64` as instructed in the docs breaks the build!

---

One option would be to just remove the `ftorch_int` from the codebase and change to `int32`/`c_int` as it doesn't fulfill the stated purpose in the docs.

---

However, I notice when looking at this that a lot of the public API uses `integer(c_int)` (or `integer(c_int64_t)` for shape).
For a public Fortran API it feels like these should really be `integer` or precisions from `iso_fortran_env`.

So there is perhaps an argument for a much larger change to make all subroutine inputs `integer` or `int32` and then rely on an internal casting to `c_int` for the `torch_tensor_from_blob` call (and `zeros`, `ones`, `empty`).
This would then prevent any compatibility issues where users with a different default integer kind need to import `iso_c_binding` and cast to `c_int` just to use FTorch. Or indeed see the examples where shapes are declared as `c_int64_t`.

~We could also consider keeping `ftorch_int` and using it where we currently have `c_int64_t` to match what is described in the docs.
However, are we ever going to see a tensor with more than 2,147,483,647 elements in a single dimension...??? That'd be 6mm global resolution!~ Edit: Torch defaults to `int64` everywhere, so perhaps we should match this.

These changes could go in now as part of V2 - realistically no-one will notice, but sets us up better for future?

---

To summarise I think my preferred approach is:

- Remove `ftorch_int` from codebase, replacing by explicit `int32`
- Make all interfaces use Fortran types (`int32`, `int64` etc.), not C-interoperability types (`c_int`, `c_int64_t` etc.)
  - This will require adding an explicit cast to C-types at `blob`, `zeros`, `ones`, and `empty` C-boundaries but makes the API Fortran native.
  - examples will be updated to use `iso_fortran_env: int64` instead of `iso_c_binding: c_int64_t`
  - get_shape and get_stride will have to return `int_64`
- ~Add back `ftorch_int` replacing instances of `int64` (currently `c_int64_t`) for shape, satisfying its original stated purpose and defaulting to `int32`.~ Edit, maybe don't bother, see note above.

---

Side note to also fix: `reduction_type` in loss functions should be `c_int` as passed a C enum.

jatkinson1000:
Closes #619 

This PR ensures we have a purely Fortran type interface to FTorch, as well as removing the ambiguously used `ftorch_int`.

- single values are copied to a C type from the Fortran type, which for most compilers becomes a no-op
- arrays now have to be allocated before copying, but are never more than 1D with `Rank` elements
- I made get_strides and get_sizes return arrays rather than C pointers
    - Argument being the Torch could be deleted and leave them dangling, and users may want to use them downstream in Fortran. I would appreciate your thoughts though @joewallwork 
- Small internal bugfixes in passing enums as `c_int` and argument order in `AdamW` optimizer

This will be followed by stacked PRs updating the examples and test suites.

TODO:

- [x] Merge #630 and #629 
- [x] Update docs as required
- [x] Update CHANGELOG


src/ftorch_tensor.fypp
```diff
@@ -489,10 +520,11 @@ contains
   function torch_tensor_get_shape(self) result(sizes)
     use, intrinsic :: iso_c_binding, only : c_f_pointer, c_int, c_int64_t, c_ptr
     class(torch_tensor), intent(in) :: self         !! Tensor to get the shape of
-    integer(kind=c_int64_t), pointer :: sizes(:)       !! Pointer to tensor data
+    integer(kind=int64), allocatable :: sizes(:)    !! Array holding the shape of the tensor
 
     ! Local data
     integer(kind=int32) :: ndims(1)
+    integer(kind=c_int64_t), pointer :: sizes_ptr(:)  !! Temporary pointer to Torch-owned memory
```
joewallwork, line 527: Do we want to use `_c_ptr` here for clarity?

src/ftorch_tensor.fypp
```diff
@@ -511,17 +543,23 @@ contains
...
   !> Return the strides of the tensor
   function torch_tensor_get_stride(self) result(strides)
     use, intrinsic :: iso_c_binding, only : c_f_pointer, c_int, c_int64_t, c_ptr
     class(torch_tensor), intent(in) :: self         !! Tensor to get the strides of
-    integer(kind=c_int64_t), pointer :: strides(:)      !! Pointer to tensor data
+    integer(kind=int64), allocatable :: strides(:)  !! Array holding the strides of the tensor
 
     ! Local data
     integer(kind=int32) :: ndims(1)
+    integer(kind=c_int64_t), pointer :: strides_ptr(:)  !! Temporary pointer to Torch-owned memory
```
joewallwork, line 562: Here too?

joewallwork (approved):
Looks good, just a couple of small queries and the outstanding tasks.

jatkinson1000:
Completed a gnarly rebase on main after #604 #632 and others.

src/ftorch_tensor.fypp
jatkinson1000: See https://github.com/Cambridge-ICCS/FTorch/pull/628/commits/90580d5ce49a449f3123b0d3f6c20f1936a0206a went for very explicit `c_int64_ptr` as it is a Fortran pointer that gets mapped to a `cptr` lower in the function, but needs the `c_int64` type as this is what Torch returns (before we map out to an `iso_env` `int64` to return to the user).

src/ftorch_tensor.fypp
jatkinson1000: See above reply.

joewallwork (approved):
All looks good

PR #631, opened 2026-09-16
Drop GCC 9 and 10 from test suite and add 14 and 15

### LINKED ISSUE #626: Drop support for GCC 9 and 10
With the requirement of CXX20 in latest Torch (see #625) the GCC 9 and 10 test suites fail.

The gfortran is fine, but g++ is not fully CXX20 compatible.

I suggest we drop support for these (and add any newer versions if they exist).

I also think it allows us to remove the 4.12 workaround on pFUnit

jatkinson1000:
Closes #626 

GCC 9 and 10 have no C++ 20 support in g++ so fail after #625

Also adds GCC 14 and 15, cascading options appropriately.

I also think we can use the latest pFUnit for everything now so update that too.


jatkinson1000:
Note, I initially hardcoded the pFUnit version into the workflow, but see we still set it as a matrix for other compilers, so have reverted to match that.

Happy to clean up all workflows to have it hardcoded if we feel reduced verbosity would be beneficial.

joewallwork (approved):
Nice to avoid those various `exclude` conditions.

PR #632, opened 2026-09-16
Bump fortitude to latest (0.9.2) and apply autofixes.

### LINKED ISSUE #627: Bump fortitude
Fortitude has undergone several releases since we pinned it.

We should do a bump to the latest, resolving any issues.

jatkinson1000:
Closes #627 

Updates to v 0.9.2 (latest).

A few autofixes that could be applied easily:
- Consistent use of double quotes
- space in `end if`

282 errors due to comments now being included in line-length checks.


jatkinson1000:
Long variable descriptors are moved 1 line down in FORD docs to satisfy fortitude now applying line-length checks to comments.
Do we want to move all to be consistent? Could get long.

Made use of the fact that line continuations in Fortran can take place mid-string with a pair of `&`, as explained by @zedthree in the pub at RSECon. Not sure I wouldn't prefer writing as 2 separate strings, however, but perhaps that's just Python stockholm syndrome.
I suppose it does make it clear that the string has not ended that line and will be continued.

joewallwork (approved):
All makes sense

PR #633, opened 2026-09-22
Add `ftorch_test_utils` as private module members

### LINKED ISSUE #311: Consider making some elements of ftorch module private
In a recent PR (#310) we explicitly made all elements of the `ftorch` module `public`.

This preserves the current behaviour under better style practice, but we should consider if some elements ought to be made `private`.

joewallwork:
Closes #311 

I went through the current members of the main `ftorch` module and couldn't see any obvious candidates for things to make private.

We could add the interfaces from `ftorch_test_utils` to the main module as private members, as proposed here.

I'd also be happy to just leave things as they are and close #311 as having been considered, though.


jatkinson1000 (approved):
Thanks @joewallwork makes sense.
Users can still import explicitly, but given common names makes sense to keep out of default scope.

